PR #9, opened 2026-06-05
fix(init): provide writable /run tmpfs; relocate guest tooling to /.lens

### LINKED ISSUE #2: WordPress (and other standard images) fail: /run is not a writable tmpfs (mkdir /run/lock: Permission denied)
## Summary

Booting the WordPress OCI image fails early with:

```
mkdir: cannot create directory '/run/lock': Permission denied
```

Root cause: the guest never mounts a writable `/run` (and `/run/lock`) tmpfs. Images that follow the standard FHS / systemd / SysV layout assume `/run` is a writable tmpfs and `/run/lock` is world-writable (sticky, mode `1777`). Because we rely on whatever `/run` ships in the image layer (often read-only or absent), the container's init/entrypoint can't `mkdir /run/lock`.

## Evidence

- `crates/lns-init/src/mount.rs` (`mount_composefs_and_exec_broker_inner`, ~lines 428–527): mounts `/dev` (devtmpfs), `/sys` (sysfs), `/dev/pts` (devpts), and the composefs overlay at `/newroot` — but **never mounts a tmpfs at `/run`**.
- `crates/lns-service/src/initramfs.rs` (`encode_cpio`, ~lines 62–84): creates `proc, sys, dev, mnt, …` but no `/run`.
- `crates/lns-session-broker/src/network.rs` does `mkdir -p /run/lns/...`, which only works incidentally when `/run` happens to be writable.

## Desired behavior (general, not WordPress-specific)

The guest should provide a standard, writable `/run` so any systemd/SysV-style image boots:

- `tmpfs` at `/run`, e.g. `nosuid,nodev,mode=0755`
- `tmpfs` at `/run/lock`, e.g. `nosuid,nodev,noexec,mode=1777`
- (consider `/run/user` as well if/when needed)

WordPress is just the first image to surface this; nginx, mysql, and most distro base images hit the same path.

## Open questions

- Mount in `lns-init` after the overlay is set up, vs. have the broker do it pre-workload? `lns-init` is the natural owner since it already does the early mounts.
- Should `/run` size be capped (`size=` option) to avoid a runaway tmpfs eating guest RAM?
- Does the busybox-driven broker glue need `/run/lns` to survive the new tmpfs mount ordering? (see broker network setup)

## Acceptance criteria

- [ ] `lns run wordpress` (or equivalent) boots past `/run/lock` creation.
- [ ] `/run` is a writable tmpfs (0755) and `/run/lock` is `1777` inside the guest.
- [ ] Covered by a Layer 3 unit test of the mount setup (real microVM boot is out of scope for CI per CLAUDE.md).

jansav:
## Summary

Make the guest's `/run` a proper writable tmpfs, and relocate lns guest tooling off `/run` so the two don't collide.

### 1. Writable `/run` tmpfs
`/run` is mounted as a hardened tmpfs (`nosuid,nodev`, mode 0755, size-capped) and `/run/lock` as a sticky `noexec` tmpfs (1777). Previously `/run` wasn't a real writable runtime dir, so standard images couldn't create runtime dirs (e.g. apache's `/var/run/apache2`).

### 2. Relocate guest tooling `/run/lns` → `/.lens`
Because `/run` is now a fresh tmpfs, anything lns places under `/run` would be shadowed. All lns-owned guest paths move to `/.lens` (supervisor, busybox/guest-tools, dhcp-dns, cmdline mask, PATH).

### 3. Chown `/run` to the sandbox user
The workload runs unprivileged, so the `/run` mountpoint is chowned to the run-as uid (mirroring the home-dir chown) so it's actually writable; `/run/lock` stays 1777, and a rootful workload keeps the standard root-owned `/run`.

Closes #2.

## Note
This relocation surfaced a separate guest-tools cache-invalidation gap (a stale cache serves busybox at the old `/run/lns` path) — tracked in #18 and fixed separately.

## Stacked follow-ups
- low-ports PR #19 (for #15) — based on this branch
- /dev/fd symlinks PR #20 (for #17) — based on the low-ports branch

### DIFF
--- a/crates/lns-init/src/mount.rs
+++ b/crates/lns-init/src/mount.rs
@@ -322,6 +322,14 @@ fn do_mkdir_p(sys: &dyn Syscalls, path: &str, mode: u32) -> Result<(), MountErro
     Ok(())
 }
 
+fn do_lchown(sys: &dyn Syscalls, path: &str, uid: u32) -> Result<(), MountError> {
+    let c = cstring(path, "lchown-path")?;
+    sys.lchown(&c, uid, uid).map_err(|err| MountError::Syscall {
+        op: format!("lchown({path}, {uid})"),
+        err,
+    })
+}
+
 fn do_umount(sys: &dyn Syscalls, target: &str) -> Result<(), MountError> {
     let c = cstring(target, "umount-target")?;
     sys.umount(&c).map_err(|err| MountError::Syscall {
@@ -353,7 +361,11 @@ fn mount_volumes(
     Ok(())
 }
 
-fn mount_run_tmpfs(sys: &dyn Syscalls, newroot: &str) -> Result<(), MountError> {
+fn mount_run_tmpfs(
+    sys: &dyn Syscalls,
+    newroot: &str,
+    sandbox_user: Option<&SandboxUser>,
+) -> Result<(), MountError> {
     let run = format!("{newroot}{RUN}");
     do_mkdir(sys, &run, 0o755)?;
     do_mount(
@@ -364,6 +376,9 @@ fn mount_run_tmpfs(sys: &dyn Syscalls, newroot: &str) -> Result<(), MountError> 
         MountFlags::Tmpfs,
         Some(RUN_TMPFS_OPTS),
     )?;
+    if let Some(user) = sandbox_user {
+        do_lchown(sys, &run, user.uid)?;
+    }
     let lock = format!("{newroot}{RUN_LOCK}");
     do_mkdir(sys, &lock, 0o755)?;
     do_mount(
@@ -528,7 +543,7 @@ fn mount_composefs_and_exec_broker_inner(
 
     mount_volumes(sys, &params.volumes, newroot)?;
 
-    mount_run_tmpfs(sys, newroot)?;
+    mount_run_tmpfs(sys, newroot, sandbox_user)?;
 
     let broker_path_c = cstring(INIT_BROKER_PATH, "broker-path")?;
     let broker_fd = open_broker_fd(sys, &broker_path_c)?;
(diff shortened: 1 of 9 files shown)


crates/lns-init/src/mount.rs
```diff
@@ -24,13 +24,19 @@ const NEWROOT: &str = "/newroot";
 const INIT_BROKER_PATH: &str = "/init-broker";
 const VOLUME_SEED_MOUNT: &str = "/mnt/vol-seed";
 const PROC_CMDLINE: &str = "/proc/cmdline";
-const CMDLINE_MASK_FILE: &str = "/run/lns/.cmdline";
+const CMDLINE_MASK_FILE: &str = "/.lens/.cmdline";
+const RUN: &str = "/run";
+const RUN_LOCK: &str = "/run/lock";
+const RUN_TMPFS_OPTS: &str = "mode=0755,size=64m";
```
heikki-laitala, line 30: Nit: the 64m `/run` cap is hardcoded. Standard distros size the `/run` tmpfs as a fraction of RAM (typically 10%); a fixed 64m could be tight for images that stage non-trivial state under `/run` (caches, per-service runtime dirs). Probably fine for the sandbox's intended workloads — just flagging that it's a fixed ceiling, not RAM-relative, in case some image hits it.

crates/lns-init/src/mount/real.rs
```diff
@@ -30,6 +30,8 @@ impl Syscalls for RealSyscalls {
             MountFlags::None => 0,
             MountFlags::ReadOnly => libc::MS_RDONLY,
             MountFlags::Bind => libc::MS_BIND,
+            MountFlags::Tmpfs => libc::MS_NOSUID | libc::MS_NODEV,
```
heikki-laitala, line 33: Nit/confirm: `/run` is mounted `nosuid,nodev` but *exec-allowed*, and it's chowned to (and writable by) the unprivileged workload user. Leaving exec on `/run` matches mainstream distro behavior and the apache rationale in the PR, so this is defensible — just confirming it's a deliberate posture choice for a hardened sandbox rather than an oversight (`/run/lock` correctly gets `noexec`). It adds little confinement against the workload itself, which already controls its own binaries.

heikki-laitala (approved):

crates/lns-init/src/mount.rs
jansav: Deliberate — the cap is there to *bound* `/run`, not to size it. An uncapped tmpfs defaults to 50% of guest RAM, so a runaway workload could starve the microVM; the fixed 64m is a conservative ceiling for the sandbox's workloads (agents/commands/OCI images), none of which stage heavy state under `/run`. Easy to bump the constant or make it RAM-relative if a real image ever hits it.

crates/lns-init/src/mount/real.rs
jansav: Confirmed, deliberate. `/run` keeps exec to match mainstream distro posture (and the Apache rationale in the PR description); `noexec` there would block legitimate runtime use while adding ~no confinement, since the workload owns its own binaries anyway. `/run/lock` gets `noexec` as the one spot where nothing should ever execute.

PR #21, opened 2026-06-08
feat(run): default OCI image runs to the image's USER

### LINKED ISSUE #16: OCI image USER directive ignored; workloads always run as sandbox/65534
## Symptom

Workloads run as uid 65534 regardless of the image's `USER`:
- `wordpress` (no `USER` -> should be root) ran as 65534 -> `could not open error log file /var/log/apache2/error.log`.
- `bitnami/wordpress` (`USER 1001`) ran as 65534 -> couldn't write its 1001-owned `/opt/bitnami` tree.

## Root cause

The run-as identity is hardcoded to `--sandbox-user`/`--sandbox-uid` (default `sandbox`/`65534`). The image config's `User` field is never read (we honor `Cmd`, `Entrypoint`, `Env`, `WorkingDir`). `lns-init` also injects the synthetic `sandbox` user into the image's passwd, so the supervisor resolves it and drops every workload to 65534.

## Proposed fix

Default the run-as identity to the image's `USER`: empty/unset -> root (OCI/Docker default), numeric -> that uid, named -> the name (supervisor drops to it by passwd lookup). Explicit `--sandbox-user`/`--sandbox-uid` override; imageless/agent runs stay on `sandbox`/65534.

Status: fix prepared locally, pending PR.

jansav:
## Summary

The run-as identity was hardcoded to `--sandbox-user`/`--sandbox-uid` (default `sandbox`/`65534`), and the image config's `User` field was never read (we honor `Cmd`/`Entrypoint`/`Env`/`WorkingDir`). So `wordpress` (no `USER` → should be root) and `bitnami/wordpress` (`USER 1001`) both ran as 65534 and failed to write image-owned dirs.

Default the run-as identity to the image's `USER`: empty/unset → root (OCI/Docker default), numeric → that uid, named → the name (the supervisor drops to it by passwd lookup). Explicit `--sandbox-user`/`--sandbox-uid` still override; imageless/agent runs stay on `sandbox`/65534. Resolution is a pure, host-tested function (`vm::resolve_run_as`).

A named `USER` that the image's `/etc/passwd` doesn't define (and that carries no uid to seed) is a hard boot refusal in lns-init (`MountError::UnresolvableUser`), not a silent fall-through to root — mirroring runc, which rejects such images with `unable to find user`.

Closes #16.


crates/lns-service/src/vm/mod.rs
```diff
@@ -184,6 +187,34 @@ fn image_config_workdir(image_config: Option<&oci_client::config::ConfigFile>) -
...
+            Some(explicit_uid.unwrap_or(DEFAULT_SANDBOX_UID)),
+        );
+    }
+    if imageless {
+        return (DEFAULT_SANDBOX_USER.to_string(), Some(DEFAULT_SANDBOX_UID));
+    }
+    match image_user.map(str::trim).filter(|u| !u.is_empty()) {
+        None => (String::new(), Some(0)),
+        Some(spec) => {
+            let name = spec.split(':').next().unwrap_or(spec);
```
jakolehm: This discards the group component from OCI `USER`, so `USER 1001:0` runs as `1001:1001`, and `USER app:root` runs with `app`'s primary passwd gid instead of the explicitly requested group. Images often rely on group ownership for writable dirs, so this can preserve the same permission failures this PR is trying to fix.

jakolehm (approved):

crates/lns-service/src/vm/mod.rs
jansav: Fixed — `resolve_run_as` now keeps the group component and the guest honors it. A numeric group seeds the user at the requested gid (so `USER 1001:0` runs as `1001:0`, not `1001:1001`), and a named group is resolved against the image's `/etc/group`, overriding the user's primary passwd gid (so `USER app:root` runs with root's gid). An explicitly requested group that is neither numeric nor present in `/etc/group` is now an unresolvable boot error rather than a silent wrong-group run. Added Layer 3 tests for both image-`USER` cases on the host and for the gid resolution + `LENS_RUN_GID` export on the guest. (commit 2e0aa5b)

PR #49, opened 2026-06-10
fix(tray): stack concurrent notifications and show on active Space

jansav:
## Summary

Two independent bugs made the tray approval window unreliable when more than one service was active at once, and on multi-Space macOS setups.

### 1. Concurrent notifications hid each other

The old renderer showed only the **first** pending entry of each kind (first network prompt, first credential prompt, first sign-in card). Every additional concurrent notification from a second service was silently queued behind it — invisible until the user acted on the first.

The renderer is replaced with a `stack_items`-driven scrollable list. Every pending entry across all four kinds (informs, network prompts, sign-ins, credential prompts) gets its own rendered card, stacked top-to-bottom in the window:

```
before: Network card for service A (service B hidden)
after:  Network card for service A
        ────────────────────────────
        Network card for service B
```

The window height is computed by `target_height` — sum of per-item heights plus chrome, clamped to the usable monitor height so a long stack scrolls inside the window rather than running off-screen.

### 2. Window stayed pinned to the Space it was created on

On macOS, an always-on-top window is by default bound to the Space (virtual desktop) where it first appeared. If the user switched to a full-screen app on another Space, the approval window was invisible even though a prompt was pending.

`join_all_spaces()` sets `NSWindowCollectionBehavior.CanJoinAllSpaces | .FullScreenAuxiliary` on the tray window at reveal time via `objc2-app-kit`. The window now follows the active Space, including full-screen app Spaces.

### 3. Positioning race on first reveal

The window position and size are now set **before** `Visible(true)`, eliminating a one-frame flash where the un-positioned window appeared mid-screen. If the monitor size isn't reported yet on the first frame, reveal is deferred by one repaint tick.

## Test instructions

1. Start two workloads in separate directories that each hit a distinct host. With both running, the tray window should show two stacked network cards simultaneously — neither hidden behind the other.
2. Switch to a full-screen app on a different macOS Space while a network prompt is pending. The approval window should appear on that Space, not remain invisible on the original desktop.
3. Dismiss one of two stacked inform banners. The clicked banner should disappear; the other should remain.
4. Reveal the "Use a token instead" fallback on one card in a multi-card stack. The window should grow to accommodate the token field without clipping the other cards.
5. With a tall enough stack that it exceeds the usable monitor height, scroll inside the window — all cards should be reachable without the window running off-screen.

### DIFF
--- a/crates/lns-service/src/tray.rs
+++ b/crates/lns-service/src/tray.rs
@@ -77,7 +77,6 @@ struct TrayApp {
     window_state: Arc<WindowState>,
     _tray: tray_icon::TrayIcon,
     last_visible: bool,
-    positioned: bool,
     /// Kept on `TrayApp` (not [`WindowState`]) so the snapshot passed to [`render_stack`] stays immutable.
     credential_inputs: HashMap<String, String>,
     /// Per-card progressive-disclosure state for the "use a token instead" fallback, keyed by the shown card's id.
@@ -139,7 +138,6 @@ impl TrayApp {
             window_state,
             _tray: tray,
             last_visible: false,
-            positioned: false,
             credential_inputs: HashMap::new(),
             token_drafts: HashMap::new(),
             current_height: WINDOW_HEIGHT,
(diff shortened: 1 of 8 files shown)


crates/lns-service/src/tray.rs
```diff
@@ -293,72 +322,132 @@ enum TokenFallbackEvent {
...
+            } else {
+                render_network_card(ui, prompt)
             }
+        }
+        StackItem::SignIn(i) => {
+            let card = &snapshot.sign_ins[i];
+            let draft = token_drafts.entry(card.credential_id.clone()).or_default();
+            render_sign_in_card(ui, card, draft)
+        }
+        StackItem::Credential(i) => {
```
jakolehm, line 418: The stacked renderer shows credential consent cards unconditionally. If a network offer starts an integration connect while a credential prompt for that same integration is pending, the credential card remains visible and clickable until `connect_integration_now` succeeds and releases it, so the user can deny it or start a second connect/sign-in while the first sign-in is in flight. The old single-card priority hid that duplicate action behind the active network/sign-in flow; stacking needs to keep the same-integration credential card suppressed or otherwise disabled while that connect is active.

jakolehm (approved):
LGTM. The comment can be fixed in a follow-up (not a blocker).

crates/lns-service/src/tray.rs
jansav: Fixing this in separate PR. Will redesign the Approval/Credential flow a bit to prevent this happening again.

PR #52, opened 2026-06-10
feat: lns volume — list, create, inspect, rm, prune

jansav:
Implements the **Volume management** lifecycle for the named volumes mounted with `lns run -v`, the way image & cache management covers images.

## What

- **`lns volume ls`** — table of named volumes with on-disk size, creation time, and the run holding them (idle volumes show `-`).
- **`lns volume create <name>`** — provision a volume ahead of its first attach; idempotent like `docker volume create`.
- **`lns volume inspect <name>`** — JSON details: logical capacity (`size_bytes`), allocated bytes (`disk_bytes`), `created`, `in_use_by`.
- **`lns volume rm <name>`** — delete one volume; refused while a run holds it, naming the holder.
- **`lns volume prune`** — delete every volume not attached to a running sandbox, reporting the space actually reclaimed; prompts unless `-f/--force`.

## How

- The store stays service-owned: five new `Request`/`Response` pairs in `lns-ipc`, dispatched in `lns-service` to new `volume_store` ops (`list/create/inspect/remove/prune` behind the existing `Fs` port, extended with `read_dir`/`metadata`/`remove_file`). The CLI is a thin verb-to-request mapper behind a `VolumeService` seam, mirroring the integration sign-in pattern.
- `rm`/`prune` take a short-lived maintenance lease in the existing `LeaseRegistry`, so deletion can never race a live attach.
- Backing images are 10 GiB **sparse** ext4 files, so sizes are reported as allocated blocks (`st_blocks * 512`) — `ls` shows `ON DISK` and prune's "Total reclaimed space" is honest; logical capacity remains visible in `inspect`.

## Tests

- Layer 2 Gherkin: `volume_management.feature` (service store semantics, 17 scenarios) and `volume_cli.feature` (CLI parsing, rendering, prompt, error surfaces, 14 scenarios), both on fakes.
- Layer 3 units for every store error path (read_dir/metadata/remove failures, lease release on failure, sort/filter rules) and CLI renderers (`format_size`, table, confirm).
- Layer 1 e2e: `volume_management.feature` drives the real binaries through create→ls→inspect→rm and prune against an isolated cache home.

## Drive-by fixes

- The e2e headless filter only consulted scenario-level tags, so the `@microvm` volumes feature ran (and failed as unmatched steps) on every `make e2e`, locally and in CI — it now honours feature-level tags. Note: the e2e harness uses `filter_run`, which never propagates scenario failures to the exit code, so `make e2e` can't actually fail CI today (ubuntu CI currently has 4 silently-failing service_lifecycle scenarios). Left as is to keep this PR green and scoped; worth a follow-up.
- HOME-mutating tests in `lns-service` were split across two serial-test groups (`env` / `cache_root`) that didn't serialize against each other; merged into `env`.
- `rfc3339` formatting moved from `ipc` to a shared `time_fmt` module so the store can stamp `created` without depending on the IPC layer.


panuhorsmalahti (approved):

crates/lns-service/src/volume_store/mod.rs
```diff
@@ -133,23 +141,163 @@ pub async fn acquire_with<F: Fs>(
...
+    let mut names: Vec<String> = entries.iter().filter_map(|p| volume_name_of(p)).collect();
+    names.sort();
+    let mut out = Vec::with_capacity(names.len());
+    for name in &names {
+        out.push(info_for(fs, registry, &image_path_in(store_root, name), name).await?);
+    }
+    Ok(out)
+}
+
+pub async fn create_with<F: Fs>(
```
jakolehm, line 224: `create_with` calls `ensure_image` without taking the per-volume lease that `acquire_with` uses. Since `write_ext4` writes through the fixed sibling temp path `<name>.img.tmp`, concurrent `lns volume create <name>` and first `lns run -v <name>:/path` can unlink/rename each other's temp file and return success against a partially written or failed image. The lifecycle command needs to join the same lease before mutating this backing image path.

jansav: Good catch — fixed in 56da5cf.

`create_with` now takes the same per-volume maintenance lease that `remove`/`prune` use before it writes a new backing image, so a concurrent `lns volume create <name>` and the first `lns run -v <name>:/path` can no longer race over the fixed `<name>.img.tmp` path. The lease is only taken on the creation branch — an already-present image stays a read-only status report, so `create` on a running volume still just reports `in_use_by` without contending for the lease.

Added a regression test (`create_of_a_new_image_yields_to_a_concurrent_holder_instead_of_racing_the_temp_file`): a held volume whose image does not yet exist now refuses `create` ("in use by run #N") and writes no image, instead of racing it.

PR #54, opened 2026-06-10
feat: lns image — pull, ls, rm, prune with digest pinning

jansav:
## Summary

The service caches everything a pull produces — layer blobs, pinned manifests — but nothing records *which images* are cached, nothing reports what the cache holds, and nothing can reclaim it: caches grow forever and the only "fix" is deleting the cache directory by hand. There is also no way to pre-warm an image before a run, and no UX for pinning a run to a digest beyond knowing the `@sha256:…` syntax exists.

This PR adds `lns image` — `docker image`-style lifecycle management for the pulled-image cache:

```bash
lns image pull <IMAGE>   # pre-warm; prints the resolved digest + a pinnable repo@sha256:… reference
lns image ls             # cached images: digest, size, pull time, and the run using them
lns image rm <IMAGE>     # drop one cached image; refused while a run uses it
lns image prune [-f]     # drop every cached image no running sandbox uses
```

### 1. Image index in the service

Every successful pull — `lns run` and `lns image pull` share `image::pull` as the choke point — records the image under the cache root: normalized reference, resolved manifest digest, per-layer digests and sizes, pull time. The new `image_store` module owns the index behind `Fs`/`Caches` ports with in-memory fakes, mirroring `volume_store`'s shape.

### 2. In-use protection and honest reclamation

`rm` is refused and `prune` skips an image while a registered running workload uses it (the run registry's raw reference is normalized against the index, so `some-image:1.0` matches `docker.io/library/some-image:1.0`). Removal drops the index record and the image's manifest-cache entry, then sweeps only layer blobs that no surviving cached image references — shared layers survive, and the reported reclaimed bytes are what was actually deleted. The content store backing live composefs mounts is deliberately untouched.

### 3. Digest pinning UX

`pull` prints a ready-to-paste pin line and `ls` carries a digest column:

```
$ lns image pull alpine:3.20
Pulled docker.io/library/alpine:3.20
Digest: sha256:8a2c…
Size: 3.5 MiB (1 layer)
Pin: docker.io/library/alpine@sha256:8a2c…
```

Pinned manifests were already cached and digest-verified on pull; the docs now spell out that a pinned rerun doesn't re-consult the registry.

### 4. Protocol and CLI plumbing

`PullImage`/`ListImages`/`RemoveImage`/`PruneImages` requests with `ImageInfo`-carrying responses in `lns-ipc`; every CLI verb is a thin IPC call behind an `ImageService` trait, host-tested via canned/fake services. `prune` prompts for confirmation unless `-f`.

### 5. Shared plumbing with #52

The `time_fmt` extraction, the `serial(env)` test-key rename, and the e2e service-in-home harness step are replicated byte-identically from the volume-management branch, so whichever PR merges second rebases with a single copy and no conflict.

## Docs

`docs/cli-reference.md` gains the `lns image` section; `docs/running-workloads.md` gains "Pinning an image by digest" and "Managing the image cache".

## Test instructions

1. `lns service start`, then `lns image ls` in a fresh cache home: empty table with `REFERENCE  DIGEST  SIZE  PULLED  IN USE` headers, exit 0.
2. `lns image pull alpine:3.20`: prints `Pulled docker.io/library/alpine:3.20`, the resolved `Digest:`, the layer count/size, and a `Pin: docker.io/library/alpine@sha256:…` line.
3. `lns run` the printed pin reference: boots without re-resolving the tag; afterwards `lns image ls` shows both entries.
4. Start a long-running sandbox from an image, then `lns image rm` that image: refused with `image "…" in use by run #N`; `lns image ls` shows `run #N` in the IN USE column.
5. After the run exits, `lns image rm` the same reference: echoes the normalized reference and `Total reclaimed space: …`; a re-pull downloads the layers again.
6. `lns image rm registry.example.test/absent:1`: exits non-zero with `no such image: …`.
7. `lns image prune` with cached idle images: prompts `Continue? [y/N]`; `n` aborts without removing anything; `lns image prune --force` removes them and reports the total reclaimed space; on an empty cache it prints `No unused images.`


panuhorsmalahti (approved):

crates/lns-service/src/image_store/mod.rs
```diff
@@ -0,0 +1,891 @@
...
+pub async fn record(pulled: &PulledImage) -> Result<()> {
+    record_with(
+        &real::RealFs,
+        &images_root()?,
+        &record_for(pulled, now_unix_secs()),
+    )
+    .await
+}
+
+pub async fn pull(image: &str) -> Result<lns_ipc::ImageInfo> {
```
jakolehm, line 228: `image_store::pull` delegates to `crate::image::pull`, but that path records the image index as a warning-only side effect in `image/real.rs`. If the index write fails, `lns image pull` still returns `ImagePulled`, so the user sees `Pulled ...` even though `lns image ls` will not show the image and `lns image rm <IMAGE>` can immediately say `no such image`. For `lns run`, best-effort indexing is reasonable because a cache-management write should not break a workload; for explicit `lns image pull`, creating the managed cache entry is the command's contract, so index persistence needs to be part of the returned success path.

jansav: Good catch — fixed and pushed. `lns image pull` now persists the index on the success path via a new `image_store::pull_with` seam (`record_with(...).await?` then `info_from`), so a failed index write surfaces as an error instead of a bogus `Pulled`. The `lns run` path keeps the best-effort write inside `crate::image::pull`, as you suggested.

Pinned by two Layer 2 scenarios: a pulled image is recorded so it lists/removes afterward, and an index-write failure is reported as a failure rather than success.

PR #71, opened 2026-06-15
fix: cap HTTP download bodies before buffering them into memory

jansav:
## What

Three download paths read the entire HTTP response body into RAM with no size ceiling:

- kernel download (`kernel::RealFetcher::fetch`)
- update-check manifest fetch (`update_check::RealFetcher::get_manifest`)
- CLI self-update manifest + release tarball (`update::http_get_bytes`)

A malicious or compromised CDN — or a MITM that gets past TLS — could stream an arbitrarily large body and exhaust host memory before the post-download SHA-256 verification ever runs.

## Change

Each download now streams the body through a small `CappedBuffer`:

- a declared `Content-Length` over the cap is rejected before any body bytes are read;
- the running total is checked on every chunk, so a lying or absent `Content-Length` still can't grow the buffer past the cap.

Caps: kernel 64 MiB, update-check manifest 1 MiB, CLI download 256 MiB (sized for the release tarball). Post-download checksum verification is unchanged — the cap is an independent pre-OOM guard.

The cap decision (`CappedBuffer::new` / `push`) is pure and unit-tested for every branch (oversized declared length, mid-stream overflow, transport-error chunk, in-order accumulation); the async `bytes_stream()` wiring lives in the e2e-covered fetch leaves. `lns-service` shares one `http_cap` module across its two leaves.

## Tests

Layer 3 unit tests for `CappedBuffer` in both crates, plus an httpmock test that an over-cap declared length is rejected. `make lint`, complexity, and the lns-cli + lns-service per-file coverage gates (100%) pass locally.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3200,6 +3200,7 @@ dependencies = [
  "cucumber",
  "dirs 5.0.1",
  "flate2",
+ "futures-util",
  "httpmock",
  "libc",
  "lns-ipc",
--- a/crates/lns-cli/Cargo.toml
+++ b/crates/lns-cli/Cargo.toml
@@ -39,6 +39,7 @@ anstyle = "1"
 # shape (reqwest with rustls + stream features, tar+flate2 for the
 # release tarball, sha2 for the manifest-pinned checksum).
 reqwest = { version = "0.12", default-features = false, features = ["rustls-tls", "stream"] }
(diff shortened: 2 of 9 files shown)


panuhorsmalahti:
Sounds like a bit far fetched attack vector..

panuhorsmalahti (approved):

PR #77, opened 2026-06-16
fix(lns-service): run headless when no display is present

jansav:
## Summary

`make e2e` could not fail CI: the harness used cucumber's `filter_run`, which never sets a non-zero exit code. Every e2e run reported success regardless of outcome — masking **20 of 31 scenarios failing on Linux CI** while the job stayed green.

Making the gate enforce surfaced the failures it had been hiding; this PR fixes them and flips the gate:

### 1. `lns-service` aborted at startup on any headless host (the big one)

`main` unconditionally started the eframe/winit tray, so with no `WAYLAND_DISPLAY`/`DISPLAY` (Linux servers, CI, plain SSH) the service died with a winit `EventLoopError` *before becoming ready* — taking the IPC server (already on its own thread) down with it. This broke the core "background service you don't turn off" promise on headless Linux, not just CI. Fix: `display_present()` gates the tray; with no display the service runs headless (`run_headless` waits for shutdown, joins the IPC thread). Headless hosts pre-authorize access in `lns-policy.yaml`.

### 2. `kill` of an unknown run was inconsistent across platforms

`kill_request` forwarded to the session layer without first checking the run exists, unlike `stop`/`inspect`/`logs`. For an unknown run that surfaced a host-specific message ("no active session" on macOS, the macOS-only capability stub on Linux) instead of the run-not-found error its siblings give. Now it checks existence first and reports `no active run with id <n>` on every host.

### 3. Stale audit fixture

The "intact chain verifies cleanly" scenario wrote only `audit.jsonl` and asserted exit 0, but `5f02c2b` made a missing `audit.anchor` exit non-zero by default. The fixture now builds a real chain **and** anchor via `lns_ipc::AuditChain`.

### 4. The gate now enforces

`filter_run` → `filter_run_and_exit`, so a failed scenario fails the job. The harness also now dumps the failing service's `service.log` (written beside the socket) into the failure output — the diagnostic that root-caused #1, since that log is ephemeral in CI.

## Result

| | macOS | ubuntu CI |
|---|---|---|
| Before | 30/31 (exit 0) | 11/31 (exit 0 — green) |
| After | 31/31 | expected 31/31, **and a real failure now turns CI red** |

Locally verified both directions: green path exits 0; an injected assertion failure exits 2.

## Test instructions

1. `make e2e` → 31/31, exit 0.
2. Break any scenario assertion → `make e2e` exits non-zero (previously it stayed 0).
3. On a headless Linux host, `lns service start` now comes up (no tray) instead of aborting.
4. `lns <kill|stop|inspect> 999999` against a running service → all report `no active run with id 999999`.

### DIFF
--- a/crates/lns-service/src/ipc/mod.rs
+++ b/crates/lns-service/src/ipc/mod.rs
@@ -223,10 +223,13 @@ async fn kill_request(run: &str, signal: lns_ipc::SignalKind) -> Response {
         Ok(id) => id,
         Err(message) => return Response::Error { message },
     };
-    if crate::run_registry::status(id).is_none() {
+    let Some(status) = crate::run_registry::status(id) else {
         return Response::Error {
             message: format!("no active run with id {id}"),
         };
+    };
+    if matches!(status, lns_ipc::RunStatus::Exited { .. }) {
+        return Response::Acknowledged;
     }
     forward_session_input(id, session_input_from_signal(signal), "Kill").await
 }
@@ -1027,6 +1030,28 @@ mod tests {
                 "expected a forwarding outcome for a sessionless registered run, got {other:?}"
             ),
         }
+    }
+
+    #[tokio::test]
+    #[serial_test::serial(global_runs)]
+    async fn handle_request_kill_of_an_already_exited_run_succeeds_without_forwarding() {
+        let id = crate::run_registry::allocate_run_id();
+        register_running(id);
+        crate::run_registry::set_exit_code(id, 0);
+        let response = handle_request(
+            &Request::Kill {
+                run: id.to_string(),
+                signal: lns_ipc::SignalKind::Kill,
+            },
+            Instant::now(),
+        )
+        .await;
+        crate::run_registry::deregister(id);
+        assert_eq!(
+            response,
+            Response::Acknowledged,
+            "killing an exited run is already satisfied and must not forward to a dead session"
+        );
     }
 
     #[tokio::test]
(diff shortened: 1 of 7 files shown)


crates/lns-service/src/ipc/mod.rs
```diff
@@ -219,10 +219,16 @@ pub async fn handle_request(request: &Request, started_at: Instant) -> Response
...
 
 async fn kill_request(run: &str, signal: lns_ipc::SignalKind) -> Response {
-    match crate::run_registry::resolve(run) {
-        Ok(id) => forward_session_input(id, session_input_from_signal(signal), "Kill").await,
-        Err(message) => Response::Error { message },
+    let id = match crate::run_registry::resolve(run) {
+        Ok(id) => id,
+        Err(message) => return Response::Error { message },
+    };
+    if crate::run_registry::status(id).is_none() {
```
heikki-laitala: Nit (optional): this existence check uses `status(id).is_none()`, so an *already-exited* run (status `Some(Exited{..})`) passes the check and forwards a kill to a dead session — which the session layer will reject with its own error. Sibling `stop_run_with` (line 319) instead treats an exited run as success (`RunStopped { forced: false }`). The PR's goal — consistent behavior for *unknown* runs — is fully met; this is just a residual divergence for the exited-run case. Erroring on kill-of-exited is arguably fine, so feel free to leave it; flagging only so the difference is intentional rather than incidental.

heikki-laitala (approved):

crates/lns-service/src/ipc/mod.rs
jansav: Fixed — `kill_request` now mirrors `stop_run_with`: it binds the status and short-circuits an already-exited run to `Response::Acknowledged` rather than forwarding a signal to a dead session and surfacing the incidental session-layer error. Unknown runs still report `no active run with id {id}`. Added `handle_request_kill_of_an_already_exited_run_succeeds_without_forwarding`, which registers a run, marks it exited, and asserts the kill is acknowledged without forwarding. (commit b9e537c)

PR #78, opened 2026-06-17
feat: Linux (KVM/Cloud Hypervisor) host support

jansav:
Implements the host-side Linux microVM backend with macOS feature parity and the identical security model. The portable subsystems (IPC, caches, OCI ingest, composefs, policy, credential/oauth flows, audit) already ran on Linux; this fills the gap concentrated in the VMM backend and run orchestrator — and **boots a real guest end-to-end on x86_64 KVM**.

## Design

Drive `cloud-hypervisor` + `virtiofsd` as child processes over unix sockets (bundled like the kernel pin, per the no-system-dependencies principle), rather than embedding a VMM library. Because the backend is processes + unix sockets + pure arg-builders, almost all of it is unit-tested on macOS via a Process-port fake and real unix sockets; the real `/dev/kvm` boot is exercised by a nightly KVM CI job (below).

## What's here

- **CH backend**: pure launch-spec builder (device order, vsock/console/fs, `shared=on`, `console=hvc0`); hybrid-vsock transport (host→guest `CONNECT`, guest→host listener feeding the relay); process orchestration (virtiofsd → wait → relay listener → cloud-hypervisor → wait → connector) behind a Spawner/Child port.
- **`GuestTransport` seam**: one trait both the Vz connector and the CH connector satisfy; `VmStopGuard` is platform-neutral; CH stops gracefully via the api-socket.
- **Unified run orchestrator**: Linux does boot → connect-to-broker → session → shutdown with stdout/stderr streaming, exit codes, supervisor relay, and interactive stdin/resize/signals.
- **Port publishing (`-p`)** on Linux; **x86_64 guest cross-build** (build.rs derives the guest musl triple from host arch); CI x86_64-musl env so the release matrix ships `lns` + `lns-service` for linux-aarch64 and linux-x86_64 with matching guests.
- **Host bind mounts (`-v /host:/guest[:ro]`) on Linux**: each bind is served as its own virtio-fs share by a dedicated virtiofsd (`--fs tag=lns-bind-{i}`), mirroring the Vz per-bind shares. Because the unprivileged workload runs as `nobody`, the bind's virtiofsd runs under `--sandbox=namespace` with `--uid-map`/`--gid-map` so the workload's guest uid/gid act as the host service user — it reads host files and writes land back on the host, matching Vz's act-as-host-user semantics. A missing bind source is rejected before boot; every virtiofsd is torn down on launch failure / CH exit; read-only is enforced guest-side via the `bind.{i}.ro` `MS_RDONLY` mount. Verified on a real Cloud Hypervisor guest (`@microvm` suite, 31/31). Closes the last Vz/CH feature gap.
- **Boots a working guest**: declares `image_type=raw` on the CH disks. cloud-hypervisor ≥ v51 disables sector-0 writes on *auto-detected* raw images, which silently broke the guest writing its upper/overlay filesystem (the run hung mid-boot); declaring the type restores it. **Requires cloud-hypervisor ≥ v51.**
- **Real-guest e2e on Linux**: `make e2e-microvm` runs the `@microvm` suite against a real Cloud Hypervisor guest (a `preflight-microvm.sh` step checks `/dev/kvm` + cloud-hypervisor + virtiofsd first). A **non-required nightly CI workflow** (`e2e-microvm.yml`) runs it on an x86_64 KVM runner — the first real-boot coverage of the backend. A new `@microvm` host-bind scenario (writable + read-only) backs the bind shares with real-boot coverage (runs on the next nightly KVM job and exercises the Vz path under `make e2e-microvm` on macOS).
- **Zero-env startup on Linux**:
  - virtiofsd auto-resolves from `/usr/libexec` (and `/usr/lib/virtiofsd`, `/usr/lib/qemu`) after PATH, so a stock apt install needs no `LNS_VIRTIOFSD_BIN`.
  - the system tray runs on a dedicated GTK main loop (`gtk::init` → build → `gtk::main`), the only place GTK lives, while the winit/eframe approval window keeps the main thread — fixing the `gtk::init` panic that previously crashed `lns-service` at startup in a Linux desktop session. Degrades to no-tray if GTK can't init. macOS is unchanged.
  - a failed `vm::boot` now surfaces the real cause (e.g. "virtiofsd not found") instead of a generic "VM backend never produced a VsockConnector".
- `cloud-hypervisor`/`virtiofsd` resolved from PATH / well-known dirs / `LNS_*_BIN`; installer probes for them and `/dev/kvm`.

## Verification

- **macOS**: lib tests, clippy, fmt green; the Vz path is unchanged (the `@microvm` suite still passes 29/29 against a real Vz guest).
- **Linux build**: verified by this PR's CI (Ubuntu lint/test/coverage) — `lns-service` can't be cross-built from macOS (glib/gtk).
- **Real guest boot on x86_64 KVM**: the `e2e-microvm` job boots a real Cloud Hypervisor guest and the full `@microvm` suite passes **29 scenarios / 145 steps** — imageless runs, OCI image pull, volumes (persist / read-only / name-keyed / audit / in-use), detached stop/kill lifecycle, port publishing, the egress network cage, env injection, credentials, `--cpus`, workdir, and unprivileged execution. This matches the macOS Vz suite (also 29/29) and is the first real-boot coverage of the CH backend.
- **Tray + service startup**: verified on a Linux desktop (Ubuntu/Wayland) — service starts with no panic, tray icon shows, Quit tears down cleanly.

## Platform notes

- Requires **cloud-hypervisor ≥ v51** + virtiofsd, on x86_64 or aarch64.
- aarch64 guest boot needs a host with working (non-nested-broken) KVM. It does **not** run under Apple Virtualization.framework's nested virtualization — the guest kernel stalls at virtio device probe (a nested-GICv3 interrupt-virtualization limitation of the host hypervisor, not lns) — so the nightly real-boot verification is x86_64. Bare-metal/non-nested aarch64 Linux is expected to work and is unverified here.

## Known follow-ups (need external resources)

- **CDN-bundle the VMMs** (sha-pinned per arch like the kernel) for zero-install UX; PATH / `LNS_*_BIN` / well-known-dirs work today. Progress:
  - ✅ `publish-virtiofsd.yml` — builds a static-musl `virtiofsd` for **both** guest arches from pinned source (upstream ships x86_64 only) and uploads a sha-pinned binary to the CDN. Both builds verified fully static (`statically linked` / `static-pie`, no `libseccomp`/`libcap-ng`/`ld-musl` refs).
  - ✅ `publish-cloud-hypervisor.yml` — re-hosts upstream's fully-static `cloud-hypervisor` (both arches; verified static) to the CDN, sha-pinned; no custom build.
  - ✅ `crate::download` — shared sha-pinned fetch+install resolver (`ensure_pinned`: override→cache(sha)→CDN→verify→atomic-install, with an optional exec bit); `kernel.rs` refactored to delegate. The VMM resolver is a thin call over it with `mode=0o755`.
  - Remaining (consumer side): a `vmm.toml` pin + bump tooling, a `vmm::ensure()` wiring over `crate::download` for CH+virtiofsd, then flip `vmm_bin::resolve`'s default from PATH to the pinned artifact.
- Promote the nightly `e2e-microvm` to a required gate once it has proven stable.


chenhunghan:
## Self-review

**Problem:** Lens Sandbox only ran on macOS (Vz); this branch adds a real Linux backend (KVM via Cloud Hypervisor + virtiofsd), so `lns run` boots a guest, shares host binds, publishes ports, and streams sessions on Linux too.

**Approach:** Introduces a `GuestTransport` trait abstracting the host VMM behind `connect`/`request_stop`, collapsing the run orchestrator, registry, port-forwarder, and stats path from dual `#[cfg]` signatures into one cross-platform flow. The CH backend is built on injected `Spawner`/`Child` ports and a shared `ConnectOnce` backoff, so the full launch lifecycle (virtiofsd → socket-wait → relay → CH → connector, kill-on-failure) is host-tested with fakes; only the two syscall leaves land in the coverage IGNORES table with reasons.

**Product alignment:** Holds the "no system dependencies" line — resolves `cloud-hypervisor`/`virtiofsd` from PATH (+ apt's `/usr/libexec`) with actionable errors plus installer/preflight probes — and keeps the security model identical across platforms (same policy, same bind/volume semantics).

Verified specifically:
- Drop-ordering of `_vm_stop_guard` vs `vm_task` on both success and early-`?` paths — the guard's `request_stop` fires before the detached boot task, so CH is always torn down.
- Relay accept-loop lifetime: spawned on CH's nested current-thread runtime, reaped when that runtime drops (no per-run task leak).
- `forward/real.rs` now bridges over `run_registry::connector` (dyn `GuestTransport`), so it compiles and works on both OSes.
- `host_known_workload_gid` → bind id-map falls back to 1:1 passthrough for the non-default-uid case (the explicitly-deferred follow-up).
- No dangling refs to the removed `VsockConnector::new_for_testing`/`VmStopGuard`, no debug artifacts, no `unwrap`/`expect` outside tests, clean conventional-commit history.

**Verdict: Ship it.** ✅

chenhunghan (approved):

PR #79, opened 2026-06-17
feat(lns-run): host bind mounts with KEEP/DROP secret guard

jansav:
## Summary

The day-one motivating use case — "run an agent against *this* repo" — requires live host files in the sandbox. Named volumes only give you scratch space. This PR adds **host bind mounts** to `lns run -v`, while enforcing the core principle that real secrets stay outside the workload.

### 1. `-v` disambiguation: absolute path → host bind, bare name → named volume

`MountSpec` replaces `VolumeMount` as the `-v` value type. Parsing is shape-based: a leading `/` in the source segment routes to `BindSpec`; anything else stays a `VolumeMount`. No flags, no ambiguity.

```
lns run -v "$(pwd)":/work alpine          # host bind (new)
lns run -v build-cache:/cache alpine      # named volume (unchanged)
lns run -v /etc/myapp:/config:ro alpine   # read-only host bind
```

The `MountSpec` type flows through `RunArgs`, `RunDefaults`, and `apply_run_defaults`. A missing host path is refused before the run starts — not silently created (deliberate Docker divergence, documented).

### 2. Secret guard: KEEP/DROP prompt with per-machine decisions

Before the run starts, `lns` scans the **top level** of each bind root for secret-shaped files (`.env*`, `*.pem`, `*.key`, `id_rsa`, `.npmrc`, `.netrc`, `.ssh/`, `.aws/`, `.gnupg`, …) and prompts once per undecided file:

```
Host bind: /Users/you/proj/.env looks like a secret. Expose it to the workload? [k]eep / [D]rop (default):
```

- Decisions are stored in `~/.lns-host-bind-decisions.json` — per-machine, never committed.
- **Default on bare Enter is DROP** — the safe choice.
- A **detached run** (`-d`) drops all undecided secrets regardless of TTY, so `lns run -d` never blocks.
- A **non-interactive run** (no terminal) drops and reports on stderr without prompting.

For team-wide "never expose these" rules, commit a `.lensignore` in the bind root — one path per line. Entries may be nested (`packages/api/.env`), must stay inside the bind (no leading `/`, no `..`), and rules for absent files are silent no-ops.

### 3. Guest-side masking (lns-init)

Dropped paths ride the kernel cmdline (`bind.N.drop.M=<name>`) to PID-1 in the guest. `lns-init` mounts each bind as a tagged virtio-fs share, then masks each dropped path **before** the workload sees the filesystem:

- A regular file → `/dev/null` bind-mount over it.
- A directory → empty `tmpfs` (mode `0000`, 4 KiB) over it.

A truncated drop-list (cmdline cut short) makes the guest **refuse to boot** with `IncompleteBind` rather than silently expose a file the operator chose to drop.

### 4. Service: virtio-fs share per bind (Vz)

Each bind becomes its own `VZVirtioFileSystemDeviceConfiguration` on the macOS Vz backend, tagged `lns-bind-{N}`. The content share tag is unaffected. The Cloud Hypervisor (Linux) backend continues to `bail!` entirely — no half-wired Linux path.

### 5. Audit chain records exposed vs masked

`record_bind_attached` appends a `bind_attached` event with `origin: host`, `source`, `target`, `exposed_secrets`, and `dropped_secrets` to the run's audit log. A KEEP is a deliberate risk acceptance, and the audit record names it explicitly.

## Docs

`docs/running-workloads.md` gains a "Host bind mounts" section covering the grammar, the secret flow, `.lensignore`, the top-level-only scan limitation, and mitigations. `docs/getting-started.md` and `docs/cli-reference.md` updated to surface the feature at first-touch.

## Test instructions

1. **No binds — unchanged behavior**: `lns run build-cache:/cache alpine -- sh -c "ls /cache"` — named volume attach works as before, no prompt, no new output.

2. **Clean bind**: `lns run -v "$(pwd)":/work alpine -- ls /work` — if the bind root has no secret-shaped files, no prompt appears and the run proceeds.

3. **First-time secret prompt — DROP**: Create `/tmp/test-bind/.env` with `SECRET=1`. Run `lns run -v /tmp/test-bind:/work alpine -- ls /work`. The prompt appears for `.env`. Press Enter (default DROP). Verify `/work/.env` is absent in the workload (or a zero-byte file). Run again — no prompt.

4. **First-time secret prompt — KEEP**: Same setup. Answer `k` at the prompt. Verify `/work/.env` is present and readable in the workload. Run again — no prompt.

5. **Read-only bind**: `lns run -v /tmp/test-bind:/work:ro alpine -- touch /work/x` — the mount is read-only, the write fails.

6. **`.lensignore`**: Add `notes.txt` to `/tmp/test-bind` and a `.lensignore` listing `notes.txt`. Run — no prompt, `notes.txt` is absent in `/work`.

7. **Detached run drops without prompting**: `lns run -d -v /tmp/test-bind:/work alpine -- sleep 30` — no prompt; a message on stderr reports the dropped `.env`.

8. **Missing host path**: `lns run -v /tmp/does-not-exist:/work alpine` — fails immediately with "host path does not exist" before connecting to the daemon.

9. **Run summary**: Any bind run — the `lns run` summary header lists `Bind: /tmp/test-bind → /work (read-write)` and, after resolution, `.env: kept (exposed)` or `.env: dropped`.

### DIFF
--- a/crates/lns-cli/src/run/host_bind.rs
+++ b/crates/lns-cli/src/run/host_bind.rs
@@ -53,15 +53,42 @@ pub fn looks_like_secret(name: &str) -> bool {
         || EXACT_SECRET_NAMES.contains(&name)
 }
 
-/// Top-level names that need a KEEP/DROP decision: secret-shaped files, plus anything the bind's `.lensignore` names so an explicit "never expose this" works for any file, not only secret-shaped ones.
-fn names_needing_a_decision(scan: &dyn DirScan, root: &Path, ignore: &[String]) -> Vec<String> {
+fn secret_shaped_top_level(scan: &dyn DirScan, root: &Path) -> Vec<String> {
     let mut found: Vec<String> = scan
         .entries(root)
         .into_iter()
-        .filter(|name| looks_like_secret(name) || is_ignored(name, ignore))
+        .filter(|name| looks_like_secret(name))
         .collect();
     found.sort();
     found
+}
+
+/// `.lensignore`-listed paths that exist in the bind, dropped with no prompt; a nested entry is honored so an explicit "never expose this" reaches a secret the top-level scan can't see.
+fn ignored_paths_present(
+    scan: &dyn DirScan,
+    root: &Path,
+    ignore: &[String],
+) -> Result<Vec<String>> {
+    let mut found = Vec::new();
+    for entry in ignore {
+        validate_lensignore_entry(entry)?;
+        if scan.exists(&root.join(entry)) {
+            found.push(entry.clone());
+        }
+    }
+    found.sort();
+    Ok(found)
+}
+
+/// A `.lensignore` entry is masked at `target/<entry>` in the guest, so an absolute or `..`-bearing entry would escape the bind — refuse it loudly rather than silently expose the file the operator listed.
+fn validate_lensignore_entry(entry: &str) -> Result<()> {
+    if entry.starts_with('/') {
+        bail!(".lensignore entry {entry:?} must be relative to the bind, not an absolute path");
+    }
+    if entry.split('/').any(|seg| seg == "." || seg == "..") {
+        bail!(".lensignore entry {entry:?} must not contain `.` or `..` path segments");
+    }
+    Ok(())
 }
 
 pub fn lensignore_patterns(scan: &dyn DirScan, root: &Path) -> Vec<String> {
@@ -97,6 +124,7 @@ impl ResolvedBind {
             target: self.target.clone(),
             read_only: self.read_only,
             dropped_paths: self.dropped.clone(),
+            kept_paths: self.kept.clone(),
         }
     }
 }
@@ -121,9 +149,11 @@ pub fn resolve_binds(
         let ignore = lensignore_patterns(scan, root);
         let mut kept = Vec::new();
         let mut dropped = Vec::new();
-        for name in names_needing_a_decision(scan, root, &ignore) {
+        for path in ignored_paths_present(scan, root, &ignore)? {
+            push_drop(&mut dropped, path)?;
+        }
+        for name in secret_shaped_top_level(scan, root) {
             if is_ignored(&name, &ignore) {
-                push_drop(&mut dropped, name)?;
                 continue;
             }
             let key = root.join(&name).to_string_lossy().into_owned();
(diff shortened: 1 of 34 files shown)


crates/lns-cli/src/run/host_bind.rs
```diff
@@ -0,0 +1,689 @@
...
+    "credentials",
+    ".ssh",
+    ".aws",
+    ".gnupg",
+];
+
+pub fn looks_like_secret(name: &str) -> bool {
+    name.starts_with(".env")
+        || name.ends_with(".pem")
+        || name.ends_with(".key")
```
panuhorsmalahti, line 60: We could add more: `*.ppk, *.keystore, credentials.*`

crates/lns-cli/src/run/host_bind.rs
```diff
@@ -0,0 +1,689 @@
...
+    ".pgpass",
+    "id_rsa",
+    "id_dsa",
+    "id_ecdsa",
+    "id_ed25519",
+    "credentials",
+    ".ssh",
+    ".aws",
+    ".gnupg",
+];
```
panuhorsmalahti, line 54: We could add more: .kube, .azure, .oci, .docker, .yarnrc.yml, .pypirc, credentials, auth.json, .netrc

crates/lns-cli/src/run/host_bind.rs
jansav: Done — added `*.ppk` (PuTTY keys) and `*.keystore` (Java keystores) as suffix matches, plus a `credentials.*` prefix match (bare `credentials` was already an exact entry). Extended the `looks_like_secret` unit test to pin `server.ppk`, `release.keystore`, and `credentials.json`. (commit e96bd2b)

crates/lns-cli/src/run/host_bind.rs
jansav: Done — added `.kube`, `.azure`, `.oci`, `.docker`, `.yarnrc.yml`, `.pypirc`, and `auth.json` to `EXACT_SECRET_NAMES` (`credentials` and `.netrc` were already in the list). Extended the unit test to pin the new entries. (commit e96bd2b)

panuhorsmalahti (approved):

PR #106, opened 2026-06-25
feat(sandbox): lns sandbox commit — round-trip a modified mount back to the registry

msa0311:
**Stacked on #61** (base = `feat/registry-login-policy-artifacts`; retarget to `main` once #61 merges). Depends on #61's registry/artifact-mount code.

Adds a `docker commit`-style round-trip so config a user (or a compliant agent) edits *inside* a running sandbox can be pushed back to the registry — today those edits land only in the ephemeral overlay upper and are lost on teardown.

## `lns sandbox commit <run> [--mount <ref|path>] [--to <ref>]`

Captures the current content of an **application-layer** mount from a **running** sandbox and re-pushes it as an updated artifact. Two capture paths, by family:

- **Layer-based (`fileset` / `tool` / `knowledge`)** — pulls the source layer to get its *original* entry list, then re-tars exactly those paths' current content (`tar -C <path> -czf - <entries>`). A mount like hermes' `/opt/data` that the agent filled with runtime junk (cron/, logs/…) still yields just `config.yaml`. Pushes the source config blob unchanged + the new layer.
- **`model`** (the compliant-agent config at `/etc/agent/model`) — re-reads the mounted spec and folds it back into the source envelope's `spec` (inverts the boot-time spec unwrap), pushing a new config blob (no layer).

Default target is the mount's own source ref (overwrites that tag); `--to <ref>` pushes elsewhere (e.g. a new tag); `--mount` disambiguates when a run has several application-layer mounts.

## How it works

- The merged content is only readable inside the guest, so commit runs against the **live** run's `GuestTransport` (errors "must be running to commit" otherwise). It captures via the existing one-shot session channel (`capture_session_output`); since that returns a `String`, the guest command base64-encodes (`… | base64 -w0`) and the host decodes. "Original files only" keeps configs well under the 1 MB capture cap.
- Reuses `artifact::{pull, pull_artifact_layers, push_artifact}`, the `expand_layer_to_specs` tar-walk (entry listing), and `run_registry::{resolve, inspect, connector}` (mount→source-ref mapping via `RunConfig.artifact_mounts`).
- New IPC `Request::CommitRun` / `Response::Committed`; service `handle_commit` mirrors `handle_stats` (streaming dispatch to a live run). The orchestration (`commit_mount_with`) is host-tested at 100% with fake capture + registry ports; only the connector/capture leaf in `ipc/adapter.rs` is IGNORE'd.

## Tests / gate

Full local gate green (lint + complexity + 100% coverage on the touched crates). Layer-3 covers both capture paths (fileset layer re-pack + model spec fold), `--to` override, `--mount` selection, and the error paths (non-running run, no app-layer mount, ambiguous selection, empty source layer, bad base64, family-less ref); CLI handler + IPC round-trip tested.

## Follow-ups / notes

- **Live-run only.** Post-exit capture (host-side ext4 + overlay reconstruction) is a separate, larger effort.
- **Guest deps:** capture needs `sh`+`tar`+`base64` in the image; the dependency-free path (a broker-side Rust tar streaming raw bytes over vsock, no String/1 MB limit) is the robust follow-up for large/binary content.
- **Demo + live round-trip** (boot hermes, edit `/opt/data/config.yaml`, `lns sandbox commit`, pull back) not yet run — will exercise once this is reviewed.


chenhunghan:

PR #107, opened 2026-06-29
feat: make the `lns run` detach chord a docker-style detach

### LINKED ISSUE #58: lns run: make the detach chord a docker-style detach (leave the run running)
## Problem

`lns run -it`'s detach chord (`ctrl-p,ctrl-q`) is **terminate-and-exit**, not a docker-style detach: it sends `SIGHUP` to the workload's foreground process group and the CLI returns the workload's exit code (129 for an `sh` killed by SIGHUP). The run does not survive. This diverges from docker, where the detach keys leave the container running.

`lns sandbox attach` was fixed to do a true docker-style detach on the `lifecycle-consolidation` branch (chord leaves the run running, no signal). The same fix can't be applied to `lns run` without service-side work, so `run` and `attach` now intentionally differ.

## Why `lns run` is harder

The service ties an attached (non-`-d`) run to its CLI. `post_pump_action` in `crates/lns-service/src/ipc/mod.rs` maps a CLI stream-close on a non-detached run to `CancelAndDeregister` — it cancels (kills) the run. Only `-d` runs get `BackgroundDrain` and survive. So even if the CLI stopped sending SIGHUP and just disconnected, the service would cancel the run anyway.

## Proposed approach

1. Add a deliberate **Detach** intent to the host↔CLI IPC (`crates/lns-ipc`) — the CLI signals "I'm detaching on purpose" before closing the stream.
2. Service: on the Detach intent, keep the run alive (treat like `BackgroundDrain`) instead of `CancelAndDeregister`; distinguish deliberate detach from an accidental CLI crash.
3. CLI: on the chord for `lns run`/`exec`, send the Detach intent, stop relaying, and return `0` without SIGHUP and without draining-to-exit — the `DetachBehaviour::LeaveRunning` path already used by `attach` in `crates/lns-cli/src/service/orchestrator.rs`.
4. Rewrite smoke-test contract #3 in `crates/lns-cli/tests/smoke/interactive-shell.exp` — it currently asserts exit `129` (128 + SIGHUP). New contract: after the chord the CLI returns `0`, the run is still listed by `lns sandbox ls`, and `lns sandbox attach` re-joins it.
5. Update `--detach-keys` help in `crates/lns-cli/src/cli.rs` and the detach sections of `docs/running-workloads.md` + `docs/cli-reference.md`.

## Acceptance criteria

- [ ] `lns run -it  -- sh`, then `ctrl-p ctrl-q`: CLI returns `0`, the run keeps running, `lns sandbox ls` still lists it, and `lns sandbox attach <id>` re-joins it.
- [ ] No SIGHUP is sent to the workload on detach.
- [ ] An accidental CLI disconnect (killing the `lns` process) follows an explicit, documented policy (cancel or survive — decide and pin it).
- [ ] Smoke test `interactive-shell.exp` updated to the new contract.
- [ ] Help text + docs describe `lns run` detach as leaving the run running.

## Context

- `DetachBehaviour` seam + `attach` fix: `crates/lns-cli/src/service/orchestrator.rs`, `crates/lns-cli/src/sandbox/mod.rs`
- Cancel-on-disconnect policy: `post_pump_action` in `crates/lns-service/src/ipc/mod.rs`
- Current chord contract: `crates/lns-cli/tests/smoke/interactive-shell.exp`

chenhunghan:
## What

`lns run -it`'s detach chord (`ctrl-p,ctrl-q`) was **terminate-and-exit**: it sent `SIGHUP` to the workload's foreground process group and the CLI returned the workload's exit code (129 for an `sh` killed by SIGHUP). The run did not survive.

It is now a **docker-style detach**, matching `lns sandbox attach`: the CLI returns `0` and the run keeps executing in the background. Re-join it with `lns sandbox attach`.

Closes #58.

## How

The service ties an attached (non-`-d`) run to its CLI — `post_pump_action` mapped a CLI stream-close on a non-detached run to `CancelAndDeregister`. The fix threads a deliberate **detach intent** through the IPC so the service can tell a deliberate detach apart from an accidental CLI crash:

- **`lns-ipc`** — new `Request::RunDetach { run_id }` / `Response::DetachAccepted`.
- **`lns-service`** — `RunDetach` fires a per-run oneshot (`detach_tx` on `RunHandle`); the response pump (`pump_responses`) returns a new `PumpOutcome::Detached`, which `post_pump_action` maps to `BackgroundDrain` (keep running) instead of `CancelAndDeregister`. The run's output keeps flowing into the log buffer, so a later `attach` re-joins it.
- **`lns-cli`** — `lns run`/`exec` no longer hardcode `SignalAndDrain`. `lns run` drives the new `DetachBehaviour::DetachRun`: on the chord it flushes pending input, sends `RunDetach`, sends **no signal**, and returns `0` without draining. `lns exec` keeps its prior `SignalAndDrain` behaviour (exec sessions aren't independently registered, so "leave running + re-attach" doesn't map onto the current model).

## Disconnect policy (acceptance criterion)

A **deliberate** detach (chord) leaves the run running. An **accidental** disconnect — the `lns` process dying without the chord — still cancels the attached run (`WriteFailed` + not-detached → `CancelAndDeregister`, unchanged). Documented in `docs/running-workloads.md` and `docs/cli-reference.md`.

## Tests

- Layer 3: `request_detach` (registry), `PumpOutcome::Detached` → `BackgroundDrain`, `pump_responses` returns `Detached` on signal / ignores a dropped sender, `handle_request(RunDetach)` accept + unknown-run error, `plan_feed` Trigger sends `RunDetach` (no SIGHUP) for `DetachRun`, `exit_code_after_detach(DetachRun)` returns 0 without draining, IPC round-trips.
- Smoke: `crates/lns-cli/tests/smoke/interactive-shell.exp` contract #3 rewritten — chord → exit `0`, run still listed by `lns sandbox ls`, `lns sandbox attach` re-joins (manual, needs a live microVM).
- Gate: `make lint`, `make complexity`, and the affected-crates coverage gate (lns-ipc / lns-service / lns-cli all 100% on changed files) pass locally.

### DIFF
--- a/crates/lns-cli/src/service/orchestrator.rs
+++ b/crates/lns-cli/src/service/orchestrator.rs
@@ -779,12 +779,20 @@ async fn send_one_shot(socket: &Path, request: &Request) -> Result<()> {
     let mut stream = UnixStream::connect(socket)
         .await
         .with_context(|| format!("connecting to {} for one-shot", socket.display()))?;
+    write_and_await_ack(&mut stream, request).await
+}
+
+async fn write_and_await_ack<S>(stream: &mut S, request: &Request) -> Result<()>
+where
+    S: AsyncReadExt + AsyncWriteExt + Unpin,
+{
     let frame = encode_frame(request).context("encoding one-shot request")?;
     stream
         .write_all(&frame)
         .await
         .context("writing one-shot request")?;
-    let _ = read_frame_bytes_async(&mut stream).await;
+    // Awaiting the reply is load-bearing for RunDetach: it blocks until the service has fired detach_tx, so the CLI never closes the run stream first and trips WriteFailed → CancelAndDeregister.
+    let _ = read_frame_bytes_async(stream).await;
     Ok(())
 }
 
@@ -1795,6 +1803,34 @@ mod tests {
         );
         assert!(matches!(control, PumpControl::Detach));
         assert!(pending.is_empty());
+    }
+
+    #[tokio::test(start_paused = true)]
+    async fn write_and_await_ack_blocks_until_the_service_responds() {
+        let (mut client, mut server) = tokio::io::duplex(4096);
+        let mut ack = std::pin::pin!(write_and_await_ack(
+            &mut client,
+            &Request::RunDetach { run_id: 7 },
+        ));
+
+        assert!(
+            timeout(Duration::from_millis(50), &mut ack).await.is_err(),
+            "the detach round-trip must not complete before the service acknowledges — \
+             a fire-and-forget RunDetach would let the CLI close the run stream first and \
+             reintroduce terminate-on-detach",
+        );
+
+        let req = read_frame_bytes_async(&mut server)
+            .await
+            .expect("server reads the request");
+        let decoded: Request = decode_frame(&mut &req[..]).expect("decode request");
+        assert_eq!(decoded, Request::RunDetach { run_id: 7 });
+
+        let frame = encode_frame(&Response::DetachAccepted).expect("encode ack");
+        server.write_all(&frame).await.expect("server writes ack");
+
+        ack.await
+            .expect("the round-trip completes once the service acknowledges");
     }
 
     #[test]
--- a/crates/lns-cli/src/service/orchestrator.rs
+++ b/crates/lns-cli/src/service/orchestrator.rs
@@ -220,6 +220,7 @@ pub async fn run_image(args: RunArgs, debug: bool) -> Result<i32> {
         run_id,
         tty,
         detach_chord,
+        DetachBehaviour::DetachRun,
         quiet,
     )
     .await
@@ -269,7 +270,16 @@ pub async fn exec_image(args: ExecArgs) -> Result<i32> {
     };
     crate::log::debug!(run_id = run_id, "exec session opened");
 
-    drive_attached_session(stream, Some(socket), run_id, tty, detach_chord, args.quiet).await
+    drive_attached_session(
+        stream,
+        Some(socket),
+        run_id,
+        tty,
+        detach_chord,
+        DetachBehaviour::SignalAndDrain,
+        args.quiet,
+    )
+    .await
 }
 
 pub async fn kill(args: KillArgs) -> Result<()> {
@@ -315,6 +325,7 @@ pub async fn ls() -> Result<()> {
 pub enum DetachBehaviour {
     SignalAndDrain,
     LeaveRunning,
+    DetachRun,
 }
 
 pub(crate) async fn drive_attached_session<S>(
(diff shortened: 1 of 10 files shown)


crates/lns-service/src/ipc/mod.rs
```diff
@@ -108,6 +123,15 @@ pub async fn handle_request(request: &Request, started_at: Instant) -> Response
...
             }
         }
+        Request::RunDetach { run_id } => {
+            if crate::run_registry::request_detach(*run_id) {
+                Response::DetachAccepted
+            } else {
+                Response::Error {
+                    message: format!("no active run with id {run_id}"),
+                }
+            }
```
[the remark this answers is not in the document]
chenhunghan: Fixed in fa16807. `request_detach` now returns a `DetachOutcome` enum (`Detached` / `NotAttached` / `NotFound`), and the `RunDetach` handler maps a still-registered-but-already-detached run (signal consumed, or pump already returned) to `"run N is not attached"`, reserving `"no active run with id N"` for a genuinely absent run.

crates/lns-cli/src/service/orchestrator.rs
```diff
@@ -727,11 +739,13 @@ fn plan_feed(
...
             let mut requests: Vec<Request> = drain_pending(run_id, pending).into_iter().collect();
-            if matches!(detach, DetachBehaviour::SignalAndDrain) {
-                requests.push(Request::RunSignal {
+            match detach {
+                DetachBehaviour::SignalAndDrain => requests.push(Request::RunSignal {
                     run_id,
                     signal: SignalKind::Hup,
-                });
+                }),
+                DetachBehaviour::DetachRun => requests.push(Request::RunDetach { run_id }),
```
heikki-laitala, line 747: Nit (maintainability): correctness here quietly depends on `send_one_shot` *awaiting* the `DetachAccepted` response (the `read_frame_bytes_async` at the end of `send_one_shot`) before this stdin pump returns and the CLI exits. That round-trip is what guarantees the service has fired `detach_tx` — and thus `pump_responses` returns `Detached` → `BackgroundDrain` — before the CLI closes its stream. If a future refactor turned `send_one_shot` into fire-and-forget, the stream could close first and the still-pending frame write would `WriteFailed` → `CancelAndDeregister`, silently reintroducing the terminate-on-detach bug this PR fixes. None of the new unit tests cover that ordering (they verify the request is produced and the oneshot fires, but not the await-before-exit invariant). A one-line comment at the `send_one_shot` response read noting it's load-bearing for detach would help — and if there's an easy way to assert it, worth a regression test.

chenhunghan: Good catch — addressed in 4fc9161. Extracted the post-connect `write request + await reply` core of `send_one_shot` into `write_and_await_ack`, which makes the await an explicit, testable seam, and added a regression test (`write_and_await_ack_blocks_until_the_service_responds`) that asserts the round-trip does **not** complete until the peer acknowledges — so a future fire-and-forget refactor fails the test rather than silently reintroducing terminate-on-detach. Also left the one-line load-bearing note at the reply read as you suggested.

heikki-laitala (approved):

PR #120, opened 2026-07-03
fix(dashboard): stop text corrupting after long uptime by replaying texture deltas eframe drops on hidden-window passes

chenhunghan:
## Problem

After the service runs for a while, all text and icons in the egui audit dashboard (and any other window) render garbled — glyphs shifted and smeared — and only `lns service stop/start` recovers. The dashboard is fine right after startup.

## Root cause

eframe runs a full egui pass for every viewport it wakes, **including the permanently-hidden ROOT viewport** (our approval window drives `App::logic()` while hidden). Each pass's `Context::end_pass` drains the context-global `TexturesDelta` into that pass's `FullOutput` — but eframe's `run_ui_and_paint` skips `paint_and_update_textures` entirely for non-visible windows, so the drained delta is **dropped on the floor** (unfixed on egui `main` as of today).

epaint's font atlas accumulates glyphs while the dashboard/approval cards render (changing relative timestamps, new sandbox names, icons). When the atlas crosses 80% fill, the *next* pass's `Fonts::begin_pass` throws the whole atlas away and rebuilds it. If that next pass is a hidden-ROOT pass (tray menu event, dashboard-open request, decision handler — all request ROOT repaints), the full-texture delta describing the rebuilt atlas is lost. From then on the GPU copy of the font texture no longer matches the CPU atlas that glyph UVs are computed against, so every glyph in every viewport samples the wrong texels — permanently, because the CPU side believes the delta was delivered.

## Fix

`ui::texture_delta_guard` — an `egui::Plugin` registered on the tray context (and both preview examples) that mirrors what eframe should do:

- `on_begin_pass` records whether eframe will paint this pass, using the exact same signal eframe's skip decision reads (`ViewportInfo::visible()` from the pass's raw input).
- `output_hook` moves the `TexturesDelta` of unpainted passes into a stash (eframe then has nothing to drop), and replays the stash ahead of the next painted pass's own deltas.

Ordering is safe by construction: sets apply oldest-first, frees are only ever delayed past the paint they must not precede, and `TextureManager` never reuses freed ids.

## Testing

- 7 Layer 3 tests pin the stash/replay behavior on a raw `egui::Context` (hidden pass hands the backend nothing; next painted pass replays stashed sets before its own and drains the stash; frees survive; unknown visibility fails open as painted).
- `make lint`, `make complexity`, `make coverage-affected` green — new file at 100% (81/81).
- `audit_dashboard` example launches and runs cleanly with the guard installed.

## Also in this PR

`make build` copied the re-signed binaries over the existing `bin/lns` / `bin/lns-service` inodes; macOS caches a vnode's code signature after first exec, so the overwritten binary gets SIGKILLed on the next launch (`zsh: killed`). The copy step now unlinks the destination first.

### DIFF
--- a/crates/lns-service/src/ui/texture_delta_guard.rs
+++ b/crates/lns-service/src/ui/texture_delta_guard.rs
@@ -1,0 +1,159 @@
+use eframe::egui;
+use egui::epaint::textures::TexturesDelta;
+
+/// Recovers the texture updates eframe discards for passes of unpainted windows (emilk/egui: `run_ui_and_paint` skips `paint_and_update_textures` when the window is hidden, losing the pass's drained `TexturesDelta` and desyncing the GPU font atlas until restart) by stashing them and replaying them ahead of the next painted pass.
+#[derive(Default)]
+pub struct TextureDeltaGuard {
+    pass_painted: Vec<bool>,
+    stashed: TexturesDelta,
+}
+
+pub fn install(ctx: &egui::Context) {
+    ctx.add_plugin(TextureDeltaGuard::default());
+}
+
+impl egui::Plugin for TextureDeltaGuard {
+    fn debug_name(&self) -> &'static str {
+        "lns_texture_delta_guard"
+    }
+
+    fn on_begin_pass(&mut self, ui: &mut egui::Ui) {
+        let painted = ui.ctx().input(|i| i.viewport().visible().unwrap_or(true));
+        self.pass_painted.push(painted);
+    }
+
+    fn output_hook(&mut self, output: &mut egui::FullOutput) {
+        let painted = self.pass_painted.pop().unwrap_or(true);
+        if !painted {
+            self.stashed
+                .append(std::mem::take(&mut output.textures_delta));
+        } else if !self.stashed.is_empty() {
+            let mut deltas = std::mem::take(&mut self.stashed);
+            deltas.append(std::mem::take(&mut output.textures_delta));
+            output.textures_delta = deltas;
+        }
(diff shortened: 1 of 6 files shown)


crates/lns-service/src/ui/texture_delta_guard.rs
```diff
@@ -0,0 +1,159 @@
...
+    fn on_begin_pass(&mut self, ui: &mut egui::Ui) {
+        let painted = ui.ctx().input(|i| i.viewport().visible().unwrap_or(true));
+        self.pass_painted.push(painted);
+    }
+
+    fn output_hook(&mut self, output: &mut egui::FullOutput) {
+        let painted = self.pass_painted.pop().unwrap_or(true);
+        if !painted {
+            self.stashed
+                .append(std::mem::take(&mut output.textures_delta));
```
heikki-laitala, line 29: Nit (optional, no action needed): `TexturesDelta::append` extends without coalescing same-id sets. The ROOT viewport here is *permanently* hidden and only ever drains via some other viewport's painted pass, so if the font atlas crosses 80% fill and rebuilds several times between two dashboard/tray paints, the stash accumulates multiple full-atlas `set` entries for the same `TextureId::Managed(0)`, all replayed at once on the next painted pass. Correct (last set wins, ordering preserved) but redundant. Bounded by rebuild frequency between paints, so practically negligible — flagging only because I traced `append` and noticed it. Not worth changing unless you observe stash bloat during long hidden stretches.

heikki-laitala (approved):

Issue #124, opened 2026-07-06
PRD (draft): Launch configured agents from OCI artifacts (typed artifact model)


chenhunghan:
> **Status: DRAFT PRD.** Captures agreed decisions at a high level. Implementation
> detail (module boundaries, file layout, sequencing) is intentionally deferred to a
> follow-up action plan.

## Problem Statement

Today `lns run <image>` runs a single OCI image. There is no way to distribute or
launch a *configured* AI agent — e.g. Claude Code with a specific set of skills and
settings — as one versioned, pullable unit from an OCI registry.

Anyone who wants "Claude Code specialized for research" must hand-assemble the
skills and settings on every machine. There is no way to:

- pull a ready-to-run, configured agent by a single reference (`…/claude:research`),
- version a skill (or the whole configured agent) independently and pin exactly what ran,
- share that configuration across a team reproducibly,
- do any of the above without leaking the secrets that agent settings tend to contain.

## Solution

Introduce **typed OCI artifacts** so a configured agent is a single pullable,
versioned artifact, adopting the model in
[`lensapp/lens-artifact-spec`](https://github.com/lensapp/lens-artifact-spec) as
the source of truth.

`lns run ghcr.io/lensapp/claude:research` pulls a **bundle (`AgentSystem`)** that
composes a **Sandbox** (the base image + isolation) + an **Agent** (the invocation) +
**FileSets** (skills/settings) + a **Policy** + **Integrations**, assembles them into a
microVM, and launches — with network access governed exactly as any other run. Plain
`lns run alpine` is unchanged. `lns` also **builds, signs, and pushes** these artifacts.

The model is generic: `lns` is an assembler with no per-agent knowledge. Compliant
agents read a standard mount layout; agents that haven't adopted it (Claude Code
today) are configured via **FileSets** mounted to their expected paths
(`/root/.claude/…`). The same mechanism launches a `codex:…` bundle.

## User Stories

1. As an agent user, I want to `lns run ghcr.io/lensapp/claude:research`, so that I get a Claude Code instance already configured with research skills and settings without assembling anything.
2. As an agent user, I want `lns run alpine` and existing image runs to behave exactly as before, so that the new artifact types are additive and never surprise me.
3. As an agent user, I want to pin a configured agent by digest (`…@sha256:…`), so that I can reproduce exactly the bytes I ran.
4. As an agent user, I want to pull a specific version by tag (`claude:research-1.4.0`), so that I can choose a vetted release of the configuration.
[...]

PR #131, opened 2026-07-07
feat: collapse the lns CLI to one noun — the sandbox (two-tier surface)

chenhunghan:
Collapses the `lns` CLI to **one user-facing noun — the sandbox** — on a two-tier surface, and makes the sandbox a first-class OCI artifact: authored as `./lns.yaml`, published with `lns push`, pulled and run anywhere. Amends the PRD in #124.

A **sandbox** is a `./lns.yaml` (`apiVersion: lns.run/v1`, `kind: Sandbox`) pinning a base image plus optional command, env, network policy, integrations, credential slots, resources, workdir, volumes, ports, and filesets. It is either **cached** or **running**. The top level carries only docker-familiar verbs, each an exact shortcut into a complete `lns sandbox` namespace (`lns run` = `lns sandbox run`, and so on); lns-native-only verbs (`validate`, `ls`, `prune`) live only under `lns sandbox`.

## The surface

```
authoring     lns init            lns sandbox validate      lns inspect [path]
distribution  lns push <ref>      lns pull <ref>            lns tag <src> <dst>
running       lns run [ref]       (no ref → ./lns.yaml)     lns ps
lifecycle     lns exec / kill / stop / logs / attach / inspect
cache         lns rm <ref>        lns sandbox ls            lns sandbox prune --force
```

## The definition — one example `lns.yaml`

`lns init` scaffolds this shape (with empty lists); every field below is applied at launch:

```yaml
apiVersion: lns.run/v1
kind: Sandbox
metadata:
  name: reviewer
spec:
  image: docker.io/library/alpine:3.20  # required base image; digest-pinned by `lns push`;
                                        # sets no USER → workload runs as the unprivileged sandbox user, not root
  command: sh                           # default workload command; `lns run -- <cmd>` still wins
  workdir: /workspace                   # guest working directory (absolute; created if missing)
  env:                                  # non-secret env seeded into the workload
    APP_ENV: dev
  resources:                            # VM boot size; per-run --cpus/--mem win;
    cpu: 1                              # the service clamps over-ceiling or overflowing values
    memory: 512Mi
  policy:                               # inline network policy, layered into the guest gate;
    defaultVerdict: ask                 # local lns-policy.yaml denies stay dominant,
    allowedRoutes:                      # approvals persist to the invoking directory's overlay only
      - match: api.github.com
        verdict: allow
  integrations:                         # declaring is arming: ids resolve against the machine catalog
    - github                            # at launch (placeholder seeded, routes allowed);
                                        # an id the catalog doesn't know refuses the launch pre-boot
  credentials:                          # slots remap an integration's env name; required + unbound
    - name: openai                      # refuses pre-boot, pointing at `lns integration connect openai`
      env: LLM_API_KEY                  # (a *denied* credential refuses distinctly from a never-bound one)
      required: true
  volumes:                              # the consumer's files: live host binds and named volumes,
    - type: bind                        # duplicate targets refused offline
      source: .
      target: /workspace
  filesets:                             # files shipped INSIDE the artifact (agent settings, skills):
    - path: ./skills                    # a `path` is packed + digest-pinned by `lns push`, a pre-published
      mountPath: /opt/skills            # digest-pinned `ref` materializes at run; symlinks + floating refs refused
      owner: root                       # pins shipped inputs; the default (workload) transfers the mount
                                        # to the run-as user so seeded tool state stays rewritable
  ports:                                # local ./lns.yaml run: auto-published on loopback (compose semantics —
    - container: 8080                   # the definition in your directory is policy you wrote);
      host: 3000                        # pulled sandbox: disclosure only until `-P`; explicit `-p` wins conflicts
```

Semantics that don't fit a comment: a definition shipping **neither** policy nor integrations plans with no baseline (the directory overlay governs verbatim); a declared `oauth` integration with no machine grant blocks the launch on a host-side sign-in streamed into the run's pre-phase; everything above is disclosed by `lns inspect` and the run summary.

**Fileset ownership.** Each fileset entry takes `owner: workload|root` (default `workload`): materialization ships a `/.lens/fileset-owned` manifest inside the runtime layer and `lns-init` chowns the listed paths right after resolving the run-as ids — only the guest can resolve a named image user — so a pulled artifact seeds rewritable tool state (`~/.claude.json`-style configs) with no per-machine step. `owner: root` pins skills/prompts an agent must not rewrite mid-run; `lns push` preserves the field when pinning. A fileset `mountPath` into `/.lens` is refused outright: a published artifact must not overlay the sandbox's own runtime (supervisor, guest tools).

**Bundled `claude-code-subscription` integration.** Claude Code's subscription auth reads `CLAUDE_CODE_OAUTH_TOKEN` (a `claude setup-token` OAuth token, `sk-ant-oat01-…`) and sends it as `Authorization: Bearer` to `api.anthropic.com` — a distinct credential from the API-key `anthropic` entry (`ANTHROPIC_API_KEY`, `sk-ant-api03-…`). A new bundled catalog entry lets a sandbox declare `integrations: [claude-code-subscription]` and arm the token per-machine (reactive card or `lns integration connect`) with no per-machine `lns integration add`; its placeholder keeps the `oat01` shape so Claude Code emits it (the boundary swaps-on-match), and its `tokenFallback` points the card at `claude setup-token`. One lns-policy contract test pins the shape. This makes a "preconfigured Claude Code sandbox" — base image + onboarding-skip fileset + this integration — pushable and runnable by anyone with their own subscription.

**Declared-only credential seeding.** A policy-bearing run previously seeded a placeholder env var for **every** bundled credential integration (applied *and* merely-connectable), so `ANTHROPIC_API_KEY` (and the rest) appeared in workloads that never declared them — noise that trips tools which detect *which* auth env vars are set (e.g. Claude Code refused to use the subscription token when `ANTHROPIC_API_KEY` was also present). Now only **declared** integrations (`spec.integrations` + credential slots) seed their env var; connectable ones are marked detect-only — still on the wire so the MITM knows their placeholder and an on-domain connect offer still fires, but carrying `env_var: None` so the guest seeds nothing. `Provider` gains a required `seeds_env()`; `DefProvider::detect_only()` opts a connectable out. (The guest skipping a `None` env var is asserted by the wire contract; the live-microvm re-verify is deferred until a virt host is available.)

**Same-domain connectable suppression.** A connectable (undeclared) integration still armed its injection from the machine-global credential store, so a leftover stored value bled into runs that never declared it — and collided when it shared a domain. A stray `anthropic` value injected `Authorization: Bearer <it>` on `api.anthropic.com` over a declared `claude-code-subscription`, and being armed it never gated, so no card fired and Anthropic 401'd. `resolve_connectable_integrations` now drops any catalog entry whose route domain an applied integration already owns: the declared credential wins the endpoint, and no same-domain sibling's machine-global value can inject over it.

## Commands added, updated, removed (vs `main`)

Both spellings of every verb parse identically: implicit `--` before a workload command, `-it`/`-ti` bundling, `-h` = hostname on `run`.

### Added

| Command | Flags | Behavior |
|---|---|---|
| `lns init` | — | Scaffold `./lns.yaml` with every spec field present and neutral defaults — it validates and runs as written. Refuses to overwrite. |
| `lns sandbox validate` | — | Validate `./lns.yaml` offline (schema + cross-field, including fileset path checks); lists every problem and exits non-zero. |
| `lns push <REF>` | `--dry-run` | Build `./lns.yaml` and upload it as a sandbox artifact in one step; `spec.image` and every `path` fileset are digest-pinned in the published config. `--dry-run` validates, packs, and builds offline, printing the digests and blob sizes that would publish — nothing uploads. Auth refusals say whether no login is stored or the stored one was rejected, with a sign-in recipe (`ghcr.io` gets the `gh` CLI path). |
| `lns pull <REF>` | — | Fetch a published sandbox into the local cache: classifies the manifest, caches the definition, prefetches `spec.image` so a pulled sandbox runs offline. Refuses a plain image. |
| `lns tag <SOURCE> <TARGET>` | — | Re-reference a cached sandbox under a new tag, within one repository. |
| `lns ps` | — | List running sandboxes with their CPU and memory. |
| `lns stop <RUN>` | `-t <SECONDS>` | Top-level shortcut for the existing `sandbox stop` (SIGTERM, then SIGKILL after the timeout, default 10 s). |
| `lns logs <RUN>` | `-f` | Top-level shortcut; a non-follow replay now terminates cleanly (terminal-ownership fix). |
| `lns attach <RUN>` | `--detach-keys <CHORD>` | Top-level shortcut for `sandbox attach`. |
| `lns inspect [TARGET]` | — | State-aware across the lifecycle: no target or a path (`.`, `lns.yaml`, `./dir`) renders the local definition offline; a run id prints live state as JSON; a cached ref prints the artifact's kind and definition (image, workdir, mounts, declared ports, filesets, integrations, over-broad-policy flag). |
| `lns rm <REF>` | — | Remove a cached sandbox (or an exited run's record) and free now-unreferenced layers; refuses a running target. |
| `lns sandbox run [REF]` | the full `lns run` set | Completes the namespace: parses and dispatches exactly like the top-level shortcut. |
| `lns sandbox ls` | — | Repurposed: lists **cached** sandboxes in the local store (alias `list`); the old active-runs listing is `lns ps`. |
| `lns sandbox prune` | `-f`/`--force` (required) | Repurposed: removes every cached sandbox not held by a running one. No interactive prompt. |

### Updated

- **`lns run [REF] [-- COMMAND…]`** — `REF` is now optional (defaults to `./lns.yaml`) and takes a **sandbox**: a registry coordinate for a published `kind: Sandbox` (directly runnable) or a path to a local definition (`.`, `lns.yaml`, `./dir`, relative/absolute; a path-named definition roots its binds and filesets at its own directory, compose-style). A plain image ref is refused with an `lns init` hint. Flag changes:
  - `-P`, `--publish-declared` (**new**) — publish the definition's `spec.ports` on loopback; automatic for a local `./lns.yaml` run, opt-in for a pulled sandbox.
  - `-p`, `--publish` — on a container-port conflict with declared ports, the explicit `-p` wins.
  - `-w`, `--workdir` — default is now `spec.workdir`, then the image `WORKDIR`.
  - `-u` / `--sandbox-user` / `--sandbox-uid` — when the image sets no `USER`, the default is now the unprivileged `sandbox` user (uid 65534), not root.
  - `COMMAND` after the ref no longer needs `--` (docker-style `lns run … echo hi`); with no `REF`, `lns run -- echo hi` overrides the local definition's command.
- **`lns exec` / `lns sandbox exec`** — the same docker-style argv normalization; the `sandbox` spelling no longer deadlocks (terminal ownership).
- **`lns kill <RUN> [--signal <SIG>]`** — unchanged surface, now a delegating shortcut through the sandbox namespace like `stop`/`rm`.
- **`lns integration connect <ID> [--policy <PATH>]`** — for credential-kind ids, now the machine-scoped **bind vehicle**: raises the value-decision card (use the host-detected value, type one, or deny) and persists every completed decision, deny included; a concurrent duplicate bind for the same id is refused; a headless service fails the bind closed (new `LNS_HEADLESS` knob). This is the fix a required-credential launch refusal points at.
- **`lns config`** — settable keys narrowed to the run gap-fillers `run.cpus`, `run.mem`, `run.registry`; legacy `run.env` / `run.volume` / `run.publish` values are warn-ignored.

### Removed

- **`lns image`** (`pull` / `ls` / `rm` / `prune`) — replaced by the sandbox-artifact verbs: `lns pull`, `lns sandbox ls`, `lns rm`, `lns sandbox prune`.
- **Top-level `lns ls`** (active runs) — replaced by `lns ps`.
- **`lns sandbox stats`** and **`lns sandbox rename`**.

## Under the hood

- **New `lns-artifact` crate** — the flat `kind: Sandbox` schema: spec types, YAML parser, offline validator (schema + cross-field), registry reference handling, and the deterministic artifact assembly `lns push` uploads. The artifact `Kind` enum is `{ Sandbox, FileSet }`.
- **One planning path for local and published runs** — `lns run` with no ref sends the local definition on the wire (`RunImageArgs.definition`, serde-default) and the service plans it through the same path as a published artifact, so policy, integrations, credentials, and resources apply identically; explicit `-- cmd` / `-e` still win. `lns-init` seeds the sandbox passwd/group lines ahead of any image alias so `whoami`/`id` resolve to the sandbox identity.

## Retired

- **The bundle model**: the `AgentSystem` / `Agent` artifact kinds, the component graph and `--with` overrides, the signature stub, and the dead standalone `Policy` / `Integration` kinds — no compat shim, nothing was published. The audit activity renames `bundle_run` → `sandbox_run`.
- **Secret-shape scanning** — deliberately: format matching is not a security boundary. The placeholder model is — real secrets stay outside the workload by construction.

## Fixes riding along

- **Policy transport drop (#127)**: `defaultTransport` and per-route `transport` now always serialize on the host→guest wire, so an ask-default policy no longer degrades to deny-all in the guest and a single direct-transport route no longer clears every route.
- **Terminal ownership**: `sandbox`, `logs`, and `attach` own the terminal, so streaming verbs (`lns sandbox exec`, non-follow `lns logs`) no longer deadlock against the dispatcher's stdio locks.
- **Credential card guidance**: when no host credential is detected, the plain credential card now names the host CLI that mints the token (`$ claude setup-token` for `claude-code-subscription`, via a new optional `tokenFallback.command` in the shipped catalog) — previously the plain credential card surfaced no fallback at all, so it dead-ended at an unexplained blank field.
- **Fail-closed guardrails**: a policy-bearing sandbox refuses to run unsupervised, concurrent same-id credential binds refuse atomically, and duplicate host ports are rejected offline.

## Tests

Layer 2 (Gherkin) + Layer 3 green at 100% per-file coverage across the workspace; `make lint` + `make complexity` clean. Layer 1: `make e2e` green, including a loopback-registry harness (+ selftest and test-only producer) that drives push → pull → run round-trips virt-free; `make e2e-microvm` **59/59** on a live Vz host (declarative launch, policy gate, in-guest integration/credential arming, streaming, plain-image refusal). A coverage audit of the new surface then added three more `@microvm` streaming scenarios — live `lns attach` (output relay + exit-code adoption), `lns logs -f` following until the run exits, and `lns sandbox run` booting like its shortcut — verified 7/7 within their feature on the same host (62 total), and activated the staged shortcut-equivalence contract (both spellings of pull/ps/stop/kill/rm/inspect issue identical service requests). CI now uploads the lcov artifact when the coverage gate fails so a floor miss is diagnosable from any machine.

## Review round (2026-07-20)

Addressed @jansav's review — six threads, each fixed with tests and resolved:

- **Policy merge (critical)**: `merge_effective` now treats every deny-by-default layer as a ceiling (intersection under deny-dominance), so a user's `defaultVerdict: deny` clamps a pulled artifact's allows the same way the artifact's `deny` default already clamped the user's — closing a silent exfil channel that defeated the user's lockdown.
- **Volume runtime-namespace guard**: volume targets get the same `/.lens` (and `/`) shadow guard as filesets; `workdir` keeps the plain absolute-path check (it isn't a mount, and `/` is a legitimate working directory).
- **Path-shaped run targets**: a configured default registry no longer corrupts `lns run .` / `../x` / `/abs` / `lns.yaml` into a registry reference — qualification is gated on the same `is_definition_path` predicate `run::target::resolve` uses.
- **Credential-domain suppression**: the connect-offer suppression routes through the gate's own wildcard- and case-insensitive `host_matches_pattern`, so an applied `api.host` suppresses a connectable `*.host` / `API.host`, not just a byte-identical pattern.
- **rm/inspect error masking**: both fall through to the cached artifact only on the run-not-found conditions (`no such run` / `no active run with id`); any other daemon error now surfaces verbatim instead of masquerading as `no such image`.
- **Sandbox schema hardening**: env keys containing `=`/whitespace/control chars are rejected, and cpu/memory are positivity-checked at parse (the service's resolver keeps the unit grammar + host memory ceiling).

## Review round (2026-07-21)

Addressed a second review pass. Seven inline threads fixed with tests and resolved:

- **Base image removable while a by-reference sandbox runs** (`image_store`): a run launched by reference (auto-pull-on-run) now persists the sandbox→base `ImageRecord`, so `rm`/`prune` can't delete the base out from under a live or cached sandbox.
- **Fileset exec bit dropped** (`lns-artifact` build): a packed fileset now carries each file's real mode instead of a forced `0o644`, so a shipped `*.sh` survives push→pull executable.
- **Shipped-policy warning blind to `*.tld` / moderate CIDRs** (`artifact/policy`): the over-broad warning now flags a whole-TLD wildcard (`*.com`) and any `/16`-or-broader CIDR the gate honors.
- **Connectable suppression keyed on routes, not injection domains** (`credential_flow`): the same-domain suppression now keys on routes ∪ injection domains, closing a Bearer bleed reachable under a custom catalog.
- **Runtime `-v` mounts bypassed the `/.lens` guard** (`lns-ipc`): `validate_volume_target` now refuses `/` and `.lens`, source-independent with the definition-level guard.
- **Malformed digest accepted as pinned** (`distribute`): a floating fileset `ref` with a truncated `@sha256:` now fails at push via the strict validator, not at pull.
- **Push auth error masked connectivity failures** (`build/push`): only a genuine auth refusal earns the login recipe; DNS/connection/server errors surface verbatim.

Also from the same pass (not inline threads):

- **Uncapped buffered blob pull** — `CountingSink` now caps a buffered blob at its declared size (clamped to the global ceiling), so a registry that under-declares then over-streams can't OOM the service.
- **Run proceeded after a failed image-index write** — the write is now propagated instead of warn-and-continue, so a run never leaves just-installed layers orphaned for a later prune.
- **Sandbox-nested argv normalization** — `lns sandbox --log-level debug run …` now skips the value-consuming global's value, matching the top-level spelling.
- **Docs** — corrected two claims that `lns push` digest-pins `spec.image`; it publishes the reference verbatim, and the docs point authors at pinning by digest themselves.

Deferred, called out explicitly: the `content_store` sweep still has no production caller. Wiring it needs a complete referenced-digest keep-set across cached filesets, guest tools, and the supervisor; a wrong keep-set risks deleting in-use content, so it stays tracked below rather than shipped half-correct this round.

## Coverage hygiene (2026-07-21)

Two files the coverage gate ignored wholesale — `artifact/real.rs` and `build/push.rs` — were hiding pure, already-tested logic behind their I/O-leaf ignore (the tests ran but didn't count, and one `declared_view_ports` out-of-range arm had no test at all). Split each into its inherent leaf and the logic it hid, moving the latter into measured modules at 100%: `artifact/inspect.rs` (`project_inspection` + `declared_view_ports`, the manifest→`ArtifactInspection` projection), `artifact/fileset.rs` (`validate_fileset_artifact` + `validate_fileset_layer_sizes`), and `build/push_auth.rs` (`select_auth` + the auth error/recipe helpers). The leaves keep only real registry/FS I/O and stay ignored with narrowed reasons.

## Review fixes (2026-07-21)

Addressed review findings on the sandbox-artifact path:

- **Artifact-declared integrations no longer arm at launch (security).** A published sandbox's `spec.integrations` were force-armed — their catalog routes appended past the effective deny-ceiling and, for a credential already bound on this machine, injected on egress with no re-consent — so an untrusted artifact declaring e.g. `github` could open `github.com` past a local `defaultVerdict: deny` and act as the user. Declaring is now disclosure only: `merge_effective` applies just the user's overlay integrations, and an artifact-declared id the user hasn't connected in this directory falls into the connectable set, offered as a live connect on first use (accepting it arms it and records it in `lns-policy.yaml`). The launch sign-in gate narrows to required `spec.credentials` slots. Rewrote the `declared_integrations` Layer 2 suite, the two e2e features that asserted arming, and the integration docs.
- **Credential-slot env-var names are validated** like `spec.env` keys (reject `=`, whitespace, control chars) rather than only checked non-empty.
- **A lost base-image retention record now fails the run** instead of warn-and-continue, so a concurrent `prune`/`rm <base>` can't reclaim the base under a live sandbox — matching the hardened `image::pull` → `record` sibling.
- **A non-positive declared layer size is refused before any fetch**, closing the buffered-blob-cap bypass where a registry declaring every layer size `0` evaded the aggregate ceiling yet still got the full per-blob cap.

## Review fixes (2026-07-22)

- **A push-scope refusal at upload now reaches the sign-in recipe.** `auth_error` was wired only onto the pre-flight `client.auth()`, but a bearer registry (ghcr.io, Docker Hub) grants `auth()` a reduced-scope token and only refuses at `push_blob`/`push_manifest_raw` with a 401/403 — which surfaced as a bare `pushing blob …: ServerError` instead of the "missing push scope" recipe written for exactly that case. A shared `is_auth_refusal` classifier now also treats `ServerError { code: 401 | 403 }` as a refusal, and the upload phases route through it, so the recipe fires for the common ghcr-token-without-`write:packages` case.
- **A crafted pulled fileset digest no longer panics the run summary.** `fileset_display` shortened a pinned digest with a byte slice (`&digest[..12]`), which aborts when byte 12 falls inside a multi-byte UTF-8 char. A pulled `SandboxView`'s fileset `ref` isn't digest-validated on the consume side, so `lns run <ref>` of an artifact declaring e.g. `repo@sha256:aaaaaaaaaaaééé` crashed (exit 101) in the preflight before launch. It now truncates by characters, not bytes.

## Deferred follow-ups

- `lns diff` — needs full pulled-definition + local grant reading; no clean small slice yet.
- `lns inspect <cached ref>` re-pulls the manifest instead of reading the local store, so a local-only `lns tag` can't be inspected offline.
- composefs/content-cache sweeps, IPC `Image*` → sandbox renames, audit scoping by sandbox name, bulk exited-run prune.
- The volume-seeding / server-image microvm specs stay parked under `crates/e2e-tests/specs/microvm/` until `lns image import` lands.

Closes nothing on its own; amends #124.


crates/lns-service/src/artifact/policy.rs
```diff
@@ -0,0 +1,231 @@
...
+    };
+    for layer in &layers {
+        for rule in &layer.network.allowed_routes {
+            if rule.verdict == Verdict::Deny {
+                push_unique(rule);
+            }
+        }
+    }
+    for (index, layer) in layers.iter().enumerate() {
+        if baseline_denies_by_default && index == 0 {
```
jansav: **Critical — a pulled artifact's `allow` rules override the user's `defaultVerdict: deny` lockdown.**

The merge suppresses the *overlay's* (user's) allow rules only when the *baseline* (author) is deny-by-default. There is no symmetric guard: when the **user** sets `defaultVerdict: deny`, the baseline's allow rules are still unioned in ahead of the deny fallback. `effective_running_policy` (`supervisor/adapter.rs:474`) passes the untrusted artifact policy as `baseline` and the local `lns-policy.yaml` as `overlay`.

Scenario: user hardens local policy to `defaultVerdict: deny` + `allow: [api.trusted.example]`; they pull and run a sandbox shipping `allow: [attacker.example]`. Effective routes become `[allow api.trusted.example, allow attacker.example]`, default `Deny` → first-match gate permits `attacker.example`. A *narrow* author allow isn't even flagged by `guardrail_flags`.

**Why:** silent data-exfil channel that defeats a user's explicit lockdown — the exact "run untrusted artifacts" threat model this PR introduces. The doc comment says the baseline is "merged *under*" the overlay, but the code lets it win. Fix: a user's `deny` default must clamp baseline allows the same way the baseline's `deny` default already clamps the user's allows.

crates/lns-artifact/src/sandbox.rs
```diff
@@ -0,0 +1,774 @@
...
+    }
+    doc.spec
+        .policy
+        .validate_local_transport()
+        .context("sandbox policy")?;
+    if let Some(workdir) = &doc.spec.workdir {
+        spec::validate_mount_path(workdir).context("workdir")?;
+    }
+    let mut targets = BTreeSet::new();
+    for volume in &doc.spec.volumes {
```
jansav, line 169: **Should-fix — volume and workdir targets bypass the `/.lens` runtime-namespace guard.**

Filesets are checked by `overlaps_runtime_namespace` (`:229`), but the volume loop here and `workdir` (`:156`) only call `validate_mount_path`, which accepts `/` and `/.lens/...` (it just requires absolute + no `..`). A pulled sandbox can declare a volume targeting `/.lens/guest-tools/bin` (the busybox `lns-init`/broker exec) or `/`, shadowing the guest runtime.

**Why:** inconsistent validation — a pulled artifact can break the guest boot (DoS). Not a privilege crossing given microVM containment, but the fileset guard should apply to volumes and workdir too.

crates/lns-cli/src/service/orchestrator.rs
```diff
@@ -15,25 +15,35 @@ use tokio::time::timeout;
...
+        let args = RunArgs::from_arg_matches(matches)?;
+        launch_run(args, ctx.debug).await
     })
 }
 
+pub async fn launch_run(mut args: RunArgs, debug: bool) -> Result<i32> {
+    args.env = crate::run::env_file::merged_run_env(&args.env_file, &args.env)?;
+    let config_path = crate::config::default_config_path()?;
+    let defaults = crate::config::load_run_defaults(&config_path)?;
+    let args = crate::config::apply_run_defaults(args, defaults);
```
jansav, line 38: **Should-fix — default registry qualification corrupts path-shaped `lns run` refs.**

`apply_run_defaults` registry-qualifies `args.image` here, *before* `run::target::resolve` classifies it at `:42`. With `run.registry ghcr.io` set (or `--registry`), `lns run .` becomes `ghcr.io/.` (`image_has_registry(".")` is false), which is then misclassified as a registry `Reference` — the local `./lns.yaml` is silently never run. Same for `/abs/project`, `..`, and `lns.yaml`.

**Why:** breaks documented path-form usage whenever a default registry is configured. Classify the target first; only registry-qualify a `RunTarget::Reference`.

crates/lns-service/src/credential_flow/integrations.rs
```diff
@@ -97,18 +124,75 @@ pub fn resolve_applied_integrations(
...
 
     let mut out = ConnectableIntegrations::default();
     for integ in catalog {
         if owned.contains(integ.id.as_str()) {
             continue;
         }
+        if integ
+            .routes
+            .iter()
+            .any(|r| owned_domains.contains(r.match_pattern.as_str()))
```
jansav: **Should-fix — same-domain credential suppression is exact-string, but the injection matcher is wildcard- and case-insensitive.**

`owned_domains.contains(r.match_pattern.as_str())` compares by byte equality, while core's `domain_matches` / `injection_matches` lowercase and honor `*` / `*.suffix` / mid-segment wildcards. A connectable whose pattern isn't byte-identical to an applied integration's (e.g. applied `api.anthropic.com` vs connectable `*.anthropic.com`, or just `API.anthropic.com`) escapes this suppression, so a leftover machine-global stored value can arm injections on the applied integration's domain — the exact clobber the comment above says it prevents.

**Why:** the suppression must use the same matcher the gate/injection uses, or it silently fails open for wildcard/case-variant catalog patterns.

crates/lns-cli/src/sandbox/mod.rs
```diff
@@ -271,39 +573,68 @@ async fn rm<W: std::io::Write>(
...
+        {
+            bail!(
+                "{} is a running sandbox; stop it first with `lns stop {}`",
+                args.run,
+                args.run
+            )
+        }
+        // An exited run is a spent record; a reference the service knows no run for is a cached sandbox.
+        Response::RunInspect { .. } => remove_run(svc, &args.run, out).await,
+        Response::Error { .. } => remove_cached(svc, &args.run, out).await,
```
jansav: **Minor — `rm` (and `inspect` at `:694`) treat *any* `InspectRun` error as "not a run, try the cache".**

A `Response::Error { .. }` falls through to `remove_cached` / `inspect_cached` regardless of the message, so a transient daemon error on a run named e.g. `reviewer` surfaces as "no such image: reviewer" instead of the real error. The "unknown run" case is the common one and fine; matching on the specific not-found condition (rather than any error) would avoid masking genuine failures.

crates/lns-artifact/src/sandbox.rs
```diff
@@ -0,0 +1,774 @@
...
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub struct SandboxSpec {
+    #[serde(default)]
+    pub image: String,
+    #[serde(default)]
+    pub command: Option<String>,
+    #[serde(default)]
+    pub workdir: Option<String>,
+    #[serde(default)]
+    pub env: BTreeMap<String, String>,
```
jansav, line 78: **Minor — input-validation gaps in the sandbox schema (low impact given containment).**

1. `env` keys (here) and credential `env` names (`:176`, only `.trim().is_empty()` checked) flow verbatim into the guest env via `format!("{key}={value}")`. A hostile artifact with an `env` key containing `=` or a newline yields a malformed/ambiguous entry; `LD_PRELOAD` / `PATH` are accepted. A key charset check would close this.
2. `resources` (`:80`, cpu/memory) is parsed but never range-validated — negative/zero/garbage values validate clean. Worth a bounds check alongside the port validation already present.

jansav (commented):
**Problem:** Collapse the CLI to one noun (the sandbox) and make a sandbox a first-class, distributable OCI artifact — authored as `./lns.yaml`, pushed, pulled, and run anywhere. This deliberately introduces a new threat surface: running untrusted published artifacts.

**Approach:** New `lns-artifact` schema crate; one service-side plan/run path for local and pulled definitions; an author-baseline network policy merged under the user's local overlay; filesets packed into the artifact with a guest-side chown; declared-only credential seeding. The cold-pull digest chain, the guest chown / PID-1 path, and deny-first route ordering are genuinely well-built (verified against oci-client internals and `lns-init`).

**Product alignment:** The artifact model fits the distribution direction, but finding 1 violates the core principle that access control is the *user's* to set — an untrusted author's policy is made authoritative over the user's `deny` lockdown.

**Verdict: Needs fixes.** One Critical (policy-merge authority inversion), three Should-fix, and a couple of Minor input-validation gaps. Inline comments below.

---

**Scoping notes**

- **Credential arming is out of scope here (accepted risk).** A pulled artifact's declared `integrations` can silently arm a credential you bound in another project, with no per-workload prompt, because credential decisions are machine-global. The durable fix is binding credential decisions to sandbox identity — tracked as a separate follow-up. Flagging only so the interim exposure is a conscious decision, not an accident.
- **Two adjacent hardening gaps are pre-existing** (unchanged by this PR) but become more reachable now that untrusted pulls are first-class, worth a separate pass: (a) the manifest/config size guard in `image/real.rs` runs *after* `pull_manifest_and_config` has buffered the document, so it can't prevent the OOM it's named for; (b) the manifest cache returns cached bytes without re-hashing against the pin, unlike the layer cache which re-verifies on read.

crates/lns-service/src/artifact/policy.rs
chenhunghan: Fixed — `merge_effective` now treats **every** deny-by-default layer as a ceiling: an allow survives only if every such layer carries it (intersection under deny-dominance). So a user's `defaultVerdict: deny` clamps a pulled artifact's allows exactly the way the artifact's own `deny` default already clamped the user's — `attacker.example` no longer rides through the user's lockdown.

This subsumes the old one-directional clamp (the existing "overlay must not widen a deny-by-default baseline" test still passes) and, unlike a naive symmetric skip, correctly *keeps* a host both layers agree on when both deny by default. New tests cover the exfil scenario, the both-layers-agree case, and the disjoint both-deny case. (commit 69d46c54)

crates/lns-artifact/src/sandbox.rs
chenhunghan: Fixed for volumes — volume targets now go through the same `overlaps_runtime_namespace` guard as filesets, so a pulled sandbox can't mount a volume over `/.lens` (or `/`) and shadow the guest `lns-init`/broker. Test covers `/`, `/.lens`, and `/.lens/guest-tools/bin`. (commit 489373a5)

I deliberately left `workdir` on the plain absolute-path check. It isn't a mount and can't shadow the runtime, and the shadow guard rejects `/` — which is a perfectly legitimate working directory, so applying it to `workdir` would break a common config for no safety gain. Happy to add a narrower `workdir` guard if you have a concrete boot-break in mind.

crates/lns-cli/src/service/orchestrator.rs
chenhunghan: Fixed — `apply_run_defaults` now guards the registry-qualification with the same `is_definition_path` predicate `run::target::resolve` uses, so a path-shaped target (`.`, `..`, `/abs/project`, `lns.yaml`, `…/lns.yaml`) reaches `resolve` unqualified and only a genuine registry reference is qualified. With `run.registry ghcr.io` set, `lns run .` now runs the local `lns.yaml` instead of becoming `ghcr.io/.`. Test covers each path shape. (commit e4ed394e)

crates/lns-service/src/credential_flow/integrations.rs
chenhunghan: Fixed — the suppression now routes through the gate's own `host_matches_pattern` (made `pub(crate)` and lowercasing both sides, as DNS names are), checked in both directions. So an applied `api.anthropic.com` now suppresses a connectable `*.anthropic.com` *or* `API.anthropic.com`, not just a byte-identical pattern — closing the fail-open you described. Tests cover the wildcard-cover, case-variant, and distinct-domain (no over-suppression) cases. (commit 5581d14c)

crates/lns-cli/src/sandbox/mod.rs
chenhunghan: Fixed — `rm`/`inspect` now fall through to the cached-artifact path only on the run-not-found conditions, and surface any other daemon error verbatim.

While wiring this I found the service actually emits *two* not-found messages: `no such run: …` when the handle resolves to nothing, and `no active run with id …` when it resolved but the run is gone. Both mean "not a live run → try the cache"; the predicate matches both and bails on anything else (transient/capability/policy errors). Tests pin that a transient `InspectRun` error bubbles up without touching the cache. (commits 60c26b7f, fcc80f61)

crates/lns-artifact/src/sandbox.rs
chenhunghan: Fixed both, with one scoping note on each:

**env keys** — a key that is empty or contains `=`, whitespace, or a control character is now rejected at parse, closing the malformed/ambiguous `KEY=VALUE` entry. I did *not* add an `LD_PRELOAD`/`PATH` name denylist: the base image can set those regardless, so blocking them by name in the definition buys nothing.

**resources** — cpu/memory are now positivity-checked at parse (parity with the port checks): a bare integer must be ≥ 1, and a `500m`/`2Gi`-style value's numeric part must be non-zero. I intentionally didn't re-implement the full unit grammar in `lns-artifact`: the service's `resolve_size` owns it and already treats zero/negative/garbage/over-ceiling as a safe fallback to defaults, bounded by a 256 GiB host-protection ceiling. Duplicating the suffix vocabulary here would just be a drift hazard, so this stays the fail-fast layer for the obvious cases while the authoritative sizing + ceiling remain in the service. (commit 287b0f9d)

crates/lns-service/src/image_store/mod.rs
```diff
@@ -151,36 +229,112 @@ pub async fn remove_with<F: Fs, C: Caches>(
...
+        candidate.reference != reference && candidate.dependencies.contains(&reference)
+    }) {
+        bail!(
+            "image {reference:?} is required by cached sandbox {:?}",
+            owner.reference
+        );
+    }
+    let active_references: HashSet<String> = records
+        .iter()
+        .filter(|record| holder(active, &record.reference).is_some())
```
jansav, line 274: **Should-fix — a running-by-reference sandbox's base image is removable by `rm`/`prune` (auto-pull-on-run path).**

`holder()` (`:123`) matches a running run whose `image` equals the target — but for a sandbox run `RunSummary.image` is the *sandbox* reference (`ipc/adapter.rs` sets `image_label = args.image`), never the base digest, so this active-holder filter never covers the base. And a bare `lns run <ref>` records the run only in the audit chain (`artifact/real.rs:44` → `audit::record_sandbox_run`) — it never writes a sandbox `ImageRecord`. The dependency guard at `:232` matches `artifact_record_for`, which is written **only** by `lns pull` (`:392`), so it has nothing to match either.

Result: while `lns run <sandbox-ref>` (never previously pulled) is live, `lns rm <base@sha256>` / `lns sandbox prune --force` delete the base record, its manifest-cache entry, and its layer blobs, and misreport reclaimed bytes. The live VM survives (content-store-backed), but the sandbox's cached/offline state is destroyed. The new tests (`remove_…_held_by_a_running_workload`, `prune_keeps_…_active_sandbox`) only cover the pulled-first path, where a sandbox record exists.

crates/lns-artifact/src/build.rs
```diff
@@ -0,0 +1,391 @@
...
+
+fn tar_layer(entries: &[FileEntry]) -> Result<Vec<u8>> {
+    let mut sorted: Vec<&FileEntry> = entries.iter().collect();
+    sorted.sort_by(|a, b| a.path.cmp(&b.path));
+    let mut builder = tar::Builder::new(Vec::new());
+    for entry in sorted {
+        let mut header = tar::Header::new_gnu();
+        header.set_entry_type(tar::EntryType::Regular);
+        header.set_size(entry.data.len() as u64);
+        header.set_mode(0o644);
```
jansav: **Should-fix — packed fileset files are always `0o644`, so a fileset can never ship an executable.**

`FileEntry` carries no mode and `tar_layer` forces `0o644` on every entry. A *local* path-fileset run preserves the real mode (`crates/lns-service/src/artifact/real.rs:102`, `mode & 0o777`), so this is a local-vs-published parity break: a shipped executable — e.g. a `.claude/hooks/*.sh` in this PR's flagship Claude-Code example — runs locally but fails with EACCES after `push`→`pull`. Determinism doesn't require flattening; the source mode is itself deterministic. Capture and re-emit the source mode.

crates/lns-service/src/artifact/policy.rs
```diff
@@ -0,0 +1,297 @@
...
+pub fn guardrail_flags(policy: &Policy) -> Vec<GuardrailFlag> {
+    let mut flags = Vec::new();
+    if policy.network.default_verdict == Verdict::Allow {
+        flags.push(GuardrailFlag::PermissiveDefaultVerdict);
+    }
+    for rule in &policy.network.allowed_routes {
+        if rule.verdict != Verdict::Allow {
+            continue;
+        }
+        if rule.match_pattern == "*" {
```
jansav: **Should-fix — the over-broad-shipped-policy warning is blind to the patterns the gate honors.**

`guardrail_flags` only flags a literal `*` allow, and `is_broad_cidr` (`:24`) only flags CIDR prefix ≤ 8 — but the gate matches `*.com`/`*.io` and honors `/9`+. A pulled artifact shipping `allow *.com` under the user's default `ask` overlay is unioned in (no ceiling) and silently pre-approves every `.com` host: no prompt, and no `⚠ over-broad shipped policy` line. This is advisory-only (a `deny`-default user still clamps it via the merge), but it's the sole automated disclosure against a broad shipped allow, and it misses exactly the wildcard/CIDR shapes worth warning about. Match `*.`-leading patterns and tighten the CIDR threshold.

crates/lns-service/src/credential_flow/integrations.rs
```diff
@@ -97,18 +124,82 @@ pub fn resolve_applied_integrations(
...
+}
+
+/// The catalog integrations a run can offer to connect: every entry (credential or oauth) not already applied and not colliding with an applied integration's domain.
 pub fn resolve_connectable_integrations(
     policy: &Policy,
     catalog: &[Integration],
 ) -> ConnectableIntegrations {
     let owned: HashSet<&str> = policy.integrations.iter().map(String::as_str).collect();
+    // An applied integration owns its route domains; a connectable that overlaps one must not ride the same run — otherwise its machine-global stored value would inject over the applied integration's credential (e.g. a leftover `anthropic` value clobbering a declared `claude-code-subscription` on api.anthropic.com).
+    let owned_domains: Vec<&str> = catalog
```
jansav: **Should-fix — same-domain connectable suppression keys on routes, not injection domains.**

The overlap check compares the connectable's `routes` against the applied integration's `routes`, but a credential is injected on its *injection* domains, and `Catalog::validate` doesn't require injection-domain ⊆ routes. So a user-authored `~/.lns-integrations.yaml` integration whose bearer injection targets `api.anthropic.com` while its `routes` point elsewhere, plus a leftover stored value, is **not** suppressed and arms a `Bearer` over a declared same-domain credential — the exact bleed class this PR set out to close, just with route ≠ injection domains.

Not reachable with the shipped catalog (bundled injection domains ⊆ routes), so it needs custom-catalog misconfiguration — hence Should-fix, not Critical. Fix: key the overlap on injection domains, or enforce injection-domain ⊆ routes in validation (defense-in-depth either way).

crates/lns-ipc/src/protocol.rs
```diff
@@ -385,9 +508,8 @@ pub fn validate_volume_target(target: &str) -> Result<(), String> {
         ));
     }
     if target.chars().any(cmdline_unsafe_char) {
-        return Err(format!(
-            "invalid volume target {target:?}: must not contain whitespace, quotes, or control characters"
-        ));
+        let reason = "must not contain whitespace, quotes, or control characters";
+        return Err(format!("invalid volume target {target:?}: {reason}"));
```
jansav, line 512: **Should-fix — runtime `-v` mounts bypass the `/.lens` runtime-namespace guard.**

`validate_volume_target` guards absolute / unsafe-chars / `..`, but not the `/.lens` (or `/`) runtime namespace — unlike a definition-declared volume, which `overlaps_runtime_namespace` refuses in `lns-artifact` (added this review round). So `lns run -v cache:/.lens …` (or a same-user raw-IPC `RunImageArgs`) mounts over the runtime tree (supervisor, guest tools), the broker's exec fails, and the run aborts.

It fails closed and isn't reachable from a pulled artifact, so it's not a privilege crossing — but the guard's whole purpose is that `/.lens` can never be shadowed; make the invariant source-independent.

crates/lns-cli/src/sandbox/distribute.rs
```diff
@@ -0,0 +1,388 @@
...
+                "ref": pinned,
+                "mountPath": fileset.mount_path,
+                "owner": owner_str(fileset.owner),
+            });
+            packed.push(PackedFileset {
+                built,
+                reference: pinned,
+            });
+        } else if let Some(declared) = &fileset.reference
+            && !declared.contains("@sha256:")
```
jansav: **Should-fix — `contains("@sha256:")` accepts a malformed digest as "pinned".**

This substring check treats `@sha256:abc` (3 hex chars) as digest-pinned — weaker than the strict 64-hex `is_digest_pinned_image` in `lns-artifact/src/spec.rs`. A bogus fileset pin publishes successfully and only fails downstream at pull, contradicting the docs' "a published sandbox whose fileset ref is not digest-pinned is refused." Reuse the strict validator here so a malformed pin is refused at push.

crates/lns-cli/src/build/push.rs
```diff
@@ -0,0 +1,156 @@
...
+        ClientProtocol::Http
+    } else {
+        ClientProtocol::Https
+    };
+    let client = oci_client::Client::new(ClientConfig {
+        protocol,
+        ..Default::default()
+    });
+    let auth = registry_auth_for(&reference);
+    client
```
jansav, line 36: **Should-fix — every `client.auth` failure is reported as "no login is stored".**

`auth_refused` is applied to *any* error from `client.auth`, including DNS failure, connection-refused, and TLS errors. A user hitting a registry that's simply unreachable is told to fix authentication and pointed at a `lns login` recipe; the real (transport) error only reaches the debug stream. Distinguish an auth denial (401/403) from a transport error and surface the latter as connectivity.

jansav (commented):
**Problem:** Collapse the `lns` CLI to one noun — the sandbox — on a two-tier surface, and make a sandbox a first-class OCI artifact (`./lns.yaml` authored, `lns push`ed, pulled, and run anywhere), retiring the bundle/image model.

**Approach:** New `lns-artifact` crate (flat `kind: Sandbox` schema, offline validator, deterministic assembly); one planning path for local and published runs; docker-familiar top-level verbs as exact shortcuts into `lns sandbox`; declared-only credential seeding with same-domain suppression; policy merge as an intersection under deny-dominance.

**Product alignment:** Strong fit — "policy you run into", "real secrets stay outside the workload", and "ephemeral by default" are preserved.

The security-critical machinery holds up under scrutiny: the policy merge is fail-closed in both directions, #127's transport serialization is fixed, credential seeding is genuinely declared-only, the guest chown is path-guarded and guest-confined, content is sha256-verified before it becomes addressable, artifact classification refuses a plain image by exact media-type, and `tag` can't escape its repo. No Critical defects.

**Verdict: Needs fixes.** 7 Should-fix items inline; 2 more Should-fix below that didn't land on a diffed line; 1 docs correction. Each carries its own reachability/severity caveat.

---

**Should-fix — not inlined (target line outside the diff hunks):**

- **Content store is never garbage-collected** — `crates/lns-service/src/content_store/mod.rs`: `sweep()` has no production caller (`#![allow(dead_code)]`). Every run installs the supervisor, guest tools, and fileset blobs into the content store, and pulled FileSet artifacts stream there too; `rm`/`prune` reclaim only the layer + build caches. Under "the sandbox you'll actually leave running" this is a monotonic disk leak. If the deferral is intentional, make it explicit.
- **`crates/lns-service/src/image/real.rs:209`** — a run proceeds after a failed index-record write (`warn … continuing`), so just-installed layers exist in no record and a later `prune`/`rm` can sweep still-referenced content, forcing a re-pull.

**Docs correction (behavior accepted, like a Helm chart's floating base tag):** `docs/running-workloads.md` (~L361, "the same publish-time pinning `spec.image` gets") and the PR body state `lns push` digest-pins `spec.image`. It doesn't — `build_artifact` serializes the config verbatim and `pack_path_filesets` rewrites only `spec.filesets`. Filesets *are* pinned; `spec.image` keeps its authored tag. (The `lns init` scaffold does **not** make this claim, so only the docs/PR text need the edit.)

**Reviewed and sound:** policy merge deny-dominance, `#127` transport serialization, declared-only credential seeding + wildcard/case same-domain suppression, guest fileset chown (guest-confined, `lchown`, `..`/NUL-guarded), resource-clamp overflow handling, port publish-consent, content addressing + verification, manifest-cache poisoning inertness, `tag` repo constraint, path-vs-registry run-target resolution.

**Accepted / not filed:** unbounded in-memory base-layer download (a registry lying about layer size can OOM the service — risk accepted; the fileset path already shows the bounded pattern via `CountingWriter` if you reconsider); `nested_workload` divergence on a mid-placed value-consuming global (too minor).

crates/lns-service/src/image_store/mod.rs
chenhunghan: Fixed in c67a9bcf. The run path (`peek_and_plan`) now persists the sandbox→base `ImageRecord` on launch — the same linkage `lns pull` writes — so a by-reference sandbox that was never explicitly pulled keeps its base out of `rm`/`prune` reach while it's live or cached. `image_store::record_artifact_run` is Layer-3 tested: `artifact_run_record_links_the_running_sandbox_to_its_normalized_base` pins the base dependency, and `a_recorded_artifact_run_protects_its_base_from_removal` pins that `remove_with` then refuses the base.

crates/lns-artifact/src/build.rs
chenhunghan: Fixed in be51976b. `FileEntry` now carries the file's real mode (`& 0o777`), `tar_layer` writes it instead of a hard `0o644`, and the host walk reads it via the dir-entry mode; the guest expansion already honored the tar mode. Pinned by `build_fileset_preserves_each_entrys_mode_in_the_tar_layer` (exec bit survives packing) and `walk_carries_each_files_mode_so_the_exec_bit_survives_packing`.

crates/lns-service/src/artifact/policy.rs
chenhunghan: Fixed in defaa89c. `guardrail_flags` now flags a wildcard sitting directly on a TLD (`*.com`) as well as a bare `*`, and `is_broad_cidr` warns on `/16`-or-broader (was `/8`) — covering the `*.com` / `/9` cases the gate honors. A scoped wildcard like `*.example.com` and a normal subnet like `/24` stay unflagged. Pinned by `a_whole_tld_wildcard_is_broad_but_a_scoped_one_is_not`, the widened `a_narrow_cidr_is_not_flagged_as_broad`, and `guardrail_flags_a_whole_tld_wildcard_the_gate_would_honor`.

crates/lns-service/src/credential_flow/integrations.rs
chenhunghan: Fixed in b85072ae. The overlap now keys on each integration's routes **and** its credential/oauth injection domains (`claimed_domains`), so a connectable that injects on an applied integration's injection domain is suppressed even when that domain isn't among its declared routes. Pinned by `connectable_excludes_an_entry_colliding_with_an_applied_injection_domain`.

crates/lns-ipc/src/protocol.rs
chenhunghan: Fixed in 975dce86. `validate_volume_target` now refuses `/` and any target whose first real segment is `.lens`, so a runtime `-v` mount can't shadow the runtime tree — the `/.lens` invariant is now source-independent, matching the definition-level guard. Pinned by `validate_volume_target_refuses_shadowing_the_lens_runtime_namespace` (and it still accepts `/.lensfoo`).

crates/lns-cli/src/sandbox/distribute.rs
chenhunghan: Fixed in 291e88c5. The `contains("@sha256:")` substring check is replaced with the strict `lns_artifact::spec::is_digest_pinned_image`, so `@sha256:abc` (or any non-64-hex digest) is refused at push instead of failing only at pull. Pinned by `push_refuses_a_declared_fileset_ref_with_a_malformed_digest`.

crates/lns-cli/src/build/push.rs
chenhunghan: Fixed in a9417448. Only `AuthenticationFailure` / `UnauthorizedError` earn the "sign in" recipe now; any other `OciDistributionError` (DNS/connection/server) surfaces verbatim under a "could not reach … to authenticate the push" context, so a connectivity failure no longer misdirects to fixing authentication. Pinned by `a_transport_failure_surfaces_the_error_without_the_login_recipe` and `an_authentication_failure_gets_the_login_recipe`.

jansav (approved):
**Problem:** Collapse the `lns` CLI to one noun — the sandbox — and make sandboxes first-class OCI artifacts (authored as `lns.yaml`, `push`ed, `pull`ed, and run anywhere). A pulled artifact is attacker-controlled, so the whole change is a new trust boundary.

**Approach:** New `lns-artifact` crate (flat `kind: Sandbox` schema + offline validator + deterministic assembly), one planning path for local and published runs, a deny-dominant policy merge, disclosure-not-arming for artifact-declared integrations, and strict digest-pinning for filesets.

**Product alignment:** Central to "run OCI images locally; control access into and out of the sandbox." The trust boundary is, on the whole, carefully built.

## Verdict: Approve ✅

The trust-boundary hardening is solid. Verified end-to-end: fileset path-traversal / control-char / setuid guards, strict 64-hex digest validation, the `merge_effective` deny-ceiling intersection, credential-domain suppression, the base-image retention bridge, the `/.lens` volume/fileset guards, blob size caps (an under-declared-then-over-streamed blob can't OOM the service), blob + DiffID digest verification, and push auth-vs-connectivity classification. All prior review-round findings map 1:1 to landed fixes; no new coverage-ignore markers; clean conventional commits.

Three **consent gaps** remain — each a case where an artifact-declared capability is applied to the consumer without an opt-in. We agreed to resolve them in a **follow-up** that adds a single first-run confirmation gate for any pulled artifact (one place that discloses and consents to everything the artifact will do), rather than bolting a separate gate onto each surface here.

### Follow-up scope

**1. Host binds — highest priority.** `crates/lns-cli/src/service/orchestrator.rs:207` — for a pulled `Reference`, `resolve()` applies the artifact's declared `spec.volumes` as real host binds unconditionally. Declared *ports* are gated (`local_definition || publish_declared`, line 216-220) but *mounts* are not, and `validate_bind_source` has no cwd confinement. So an artifact with `spec.volumes: [{type: bind, source: "/", target: "/host", readOnly: false}]` mounts the host root **read-write** into the untrusted guest with no prompt (and, in `-q`/`-d`, no disclosure). Recommend gating declared binds **off by default** for pulled artifacts (mirror the ports gate); the first-run confirmation then re-enables them behind consent.

**2. Base image not pin-enforced.** `crates/lns-artifact/src/sandbox.rs` (`parse`) accepts any non-empty `spec.image` and `peek_and_plan` (`crates/lns-service/src/artifact/real.rs:33`) never checks it, so a `@sha256`-pinned artifact can boot a floating, swappable base (`image: evil.com/rootfs:latest`) pulled trust-on-first-use. This is a **consumer-side** check — an attacker never runs `lns push`, so it belongs in the pull/run gate as enforce-or-warn, not in push validation. (`lens.dev/v1alpha1` already enforces base pinning.)

**3. Silent network allows.** A pulled artifact's narrow `allowedRoutes: [{match: …, verdict: allow}]` merges into the effective policy and, under the default `ask`, pre-authorizes egress with only a generic "ships a network policy" disclosure. Subsumed by (1) **provided** the first-run gate enumerates the shipped policy's allow-routes and declared integrations — not just binds/ports.

Everything else is good to ship.

PR #140, opened 2026-07-08
perf(lns-service): pool the OCI registry client so repeated registry contacts skip the auth handshake

### LINKED ISSUE #132: perf: warm runs of a tag reference pay a full registry round-trip (~1.7s) despite a complete local cache
## Problem

On a fully warm cache, `lns run alpine -- true` takes ~2.9s on macOS/Vz, and ~1.7s of that is registry traffic for an image the CLI itself reports as `✓ image cached`.

`CachingRegistry::pull_manifest_and_config` (`crates/lns-service/src/image/manifest_cache.rs`) only serves cache hits for digest-pinned references:

```rust
// Only digest-pinned references are immutable; caching a tag would freeze it past upstream republishes.
if reference.digest().is_none() {
    return self.inner.pull_manifest_and_config(reference).await;
}
```

So a bare `alpine` does the full anonymous-token + manifest-index + platform-manifest + config-blob chain against Docker Hub on **every run**, and this sits on the boot critical path (the `tokio::join!` in `run/orchestrator.rs` can't complete before it).

### Measured (warm cache, arm64 macBook, Vz)

| Reference form | Wall time |
|---|---|
| `lns run alpine -- true` | ~2.9s (of which ~1.7s between "auditing" and "image cached") |
| `lns run -- true` (imageless, no registry) | ~0.75–1.4s |

## Proposal

Adopt Docker-style pull semantics for tag references:

- Cache tag manifests too, keyed by the tag, with a freshness policy — either serve-stale-and-revalidate-in-background (warn/update if the tag moved), or a short TTL (e.g. 15 min).
- Add `--pull=always|missing|newer` (default `missing` or the TTL behaviour) as the escape hatch.
- Where a synchronous revalidation is wanted, a single HEAD request for the tag's current digest (compared against the cached digest) replaces the current 3–4 sequential round-trips.

This alone takes the warm alpine run from ~2.9s to ~1.2s.

## Related quirk found while verifying

Pinning by the *platform* digest errors out: the manifest cache stores the platform-manifest digest, but the puller compares it against the index digest —

```
✗ manifest digest mismatch for alpine@sha256:45e0... — expected sha256:45e0..., received sha256:e474...
```

This blocks the user-side workaround of digest-pinning to skip the fetch, so it's worth fixing alongside.

chenhunghan:
## What

Pool the OCI registry client per **(registry, credential)** in `lns-service` so the bearer token and HTTPS connection survive across pulls. `image::pull` takes a client from the pool instead of constructing a fresh `oci_client::Client` every time. No user-visible behavior change.

## Why — this complements #135, it doesn't overlap it

**#135 fixes #132 by avoiding registry round-trips** on warm tag runs (tag-manifest cache, 60 s TTL, HEAD-revalidate, `--pull`). This PR is the orthogonal, freshness-neutral half: **it makes the round-trips that still happen cheap.**

A fresh `oci_client::Client` per pull discards the bearer-token cache and connection pool, so every registry contact re-runs the auth-token handshake. Measured on this branch (Apple Vz, docker.io, `alpine` by tag):

| registry contact | time |
|---|--:|
| full manifest+config, **fresh** client (auth + 2 GETs) | ~1.72 s |
| full manifest+config, **pooled** token (2 GETs only) | ~0.76 s |
| ⇒ auth-token handshake alone | **~1.0 s** |

Once #135 lands, the registry is still contacted — and re-authenticated — on exactly these paths:

- `pull=auto` after the 60 s TTL lapses → a HEAD revalidation,
- `pull=always` → a HEAD / refetch every run,
- the first pull of a service session / cold cache.

Pooling the client removes the ~1.0 s handshake from all of them (the token expires naturally; the connection is kept alive). Concretely, #135's revalidation HEAD drops from ~1.17 s (fresh-client, measured) to ~0.2 s (pooled token, derived: 1.17 s − ~1.0 s auth).

`oci_client::Client` is `Clone` over `Arc<RwLock<..>>` auth + a `TokenCache` — designed to be shared. It's keyed by a **SHA-256 credential fingerprint** (never the plaintext secret) so a `lns login` change gets a fresh client instead of reusing a stale token (`store_auth_if_needed` only stores creds if absent — a naive global client would be a stale-credential bug).

## Standalone measurement (on `main`, before #135)

On `main` every warm run does a full fetch, so the standalone effect is large:

| scenario (wall-clock boot, median n=10) | before | after |
|---|--:|--:|
| `alpine` tag, warm | 2436 ms | 1490 ms (−39%) |
| imageless | 547 ms | 539 ms (no registry path) |

Read these as *evidence of the handshake cost this removes*, not a standalone headline: once #135 makes most warm runs cache hits, this PR's benefit narrows to the revalidate / `always` / cold paths above — which is the intended composition.

## Interaction & merge order

Independent of #135; both touch `image::pull`, so whichever lands second rebases. The pooled client simply feeds the (post-#135) `CachingRegistry` — no design conflict. If #135 folds this in directly, close this.

## Testing

- **Layer 3** unit tests for the pool (`image/client_pool.rs`): factory runs once per key, per-key isolation, credential fingerprinting (hides secret/token, re-keys on user/secret/token change, separates registries).
- `make lint && make complexity` pass; `make coverage`: `image/client_pool.rs` 100% (60/60), `image/mod.rs` 100%; `image/real.rs` remains the IGNORE'd production-wiring adapter. Measured end-to-end on a real macOS/Vz host against an isolated service.

## Security

Touches the registry-pull path. Credentials never appear in the pool key (SHA-256 fingerprint); a credential change forces a fresh client, so a rotated/removed credential is never silently reused. No change to the microVM boundary, policy, audit chain, or host IPC.

Refs #132. Complements #135.

### DIFF
--- a/crates/lns-service/src/image/client_pool.rs
+++ b/crates/lns-service/src/image/client_pool.rs
@@ -55,17 +55,19 @@ mod tests {
     fn get_or_create_runs_the_factory_once_per_key() {
         let pool: ClientPool<u32> = ClientPool::new();
         let calls = AtomicU32::new(0);
-        let a = pool.get_or_create("k".into(), || {
-            calls.fetch_add(1, Ordering::SeqCst);
-            42
-        });
-        let b = pool.get_or_create("k".into(), || {
-            calls.fetch_add(1, Ordering::SeqCst);
-            99
-        });
-        assert_eq!(a, 42);
-        assert_eq!(b, 42, "a warm key returns the cached client, not a new one");
-        assert_eq!(calls.load(Ordering::SeqCst), 1, "factory runs only on the miss");
+        // fetch_add returns the pre-increment count, so the factory both records that it ran and yields 0 on the miss vs 1 if the hit wrongly re-ran it.
+        let a = pool.get_or_create("k".into(), || calls.fetch_add(1, Ordering::SeqCst));
+        let b = pool.get_or_create("k".into(), || calls.fetch_add(1, Ordering::SeqCst));
+        assert_eq!(a, 0, "the miss runs the factory");
+        assert_eq!(
+            b, 0,
+            "the hit returns the cached client, not a fresh factory run (which would be 1)"
+        );
+        assert_eq!(
+            calls.load(Ordering::SeqCst),
+            1,
+            "the factory runs only on the miss"
+        );
     }
 
     #[test]
@@ -73,7 +75,11 @@ mod tests {
         let pool: ClientPool<u32> = ClientPool::new();
         assert_eq!(pool.get_or_create("a".into(), || 1), 1);
         assert_eq!(pool.get_or_create("b".into(), || 2), 2);
-        assert_eq!(pool.get_or_create("a".into(), || 3), 1, "key a still cached");
+        assert_eq!(
+            pool.get_or_create("a".into(), || 3),
+            1,
+            "key a still cached"
+        );
     }
 
     #[test]
@@ -98,7 +104,10 @@ mod tests {
         let other_user = auth_fingerprint(&RegistryAuth::Basic("bob".into(), "s1".into()));
         let other_secret = auth_fingerprint(&RegistryAuth::Basic("alice".into(), "s2".into()));
         assert_ne!(base, other_user, "a different user re-keys the client");
-        assert_ne!(base, other_secret, "a re-login with a new secret re-keys the client");
+        assert_ne!(
+            base, other_secret,
+            "a re-login with a new secret re-keys the client"
+        );
     }
 
     #[test]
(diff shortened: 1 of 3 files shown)


chenhunghan:
Related to #132 (same warm-tag registry-latency problem).

To be clear on scope: this PR does **not** implement #132's proposal (tag-manifest caching, HEAD-revalidation, `--pull`, or the platform-digest-mismatch fix). It's a complementary, freshness-neutral change — reusing the registry client so the bearer token + connection pool survive across pulls, eliminating the repeated auth handshake (warm `alpine` 2.44s → 1.49s here).

#132's mechanism (HEAD-revalidated tag cache + `--pull`) is still worth doing on top of this; the two are additive (token reuse also makes #132's HEAD cheaper). The platform-digest pinning quirk noted in #132 is untouched here.

PR #156, opened 2026-07-22
feat: confirm a pulled sandbox's host binds and filesets before the run starts

chenhunghan:
## Problem

A published sandbox artifact can declare host binds (`spec.volumes` with `type: bind`), named volumes (`type: volume`), and filesets. When another user runs it by reference, those are honored with no gate: a relative bind source like `.` resolves against the consumer's cwd and is virtiofs-shared into the guest; an author-chosen volume name attaches from the consumer's global volume store (`<cache_root>/volumes/<name>.img` — unscoped, so it can name another project's persisted data) read-write across runs; and filesets materialize author-published files the workload will read. The only UX today is an informational run summary (suppressed by `--quiet`) — the run proceeds immediately.

## Change

`lns run <ref>` now discloses every host bind (resolved host path, target, mode), named volume (name, target, mode), and fileset (source, mount path) the pulled sandbox will mount, and asks `Continue? [y/N]` before anything is mounted or started. Default is No.

- **Fails closed without a terminal**: a non-interactive or detached run of a pulled sandbox that declares any of them refuses with an error naming the sandbox and pointing at the escape hatch.
- **`--yes`** accepts the mounts without prompting (scripted/CI callers).
- Local `./lns.yaml` runs are unaffected — the user authored those declarations.
- The confirmation prints its own disclosure, so `--quiet` cannot hide it.
- Runs where the pulled artifact declares none of the three are unchanged (no prompt).

## What the user sees

Interactive run of someone else's published sandbox:

```
$ cd ~/some-project
$ lns run ghcr.io/acme/reviewer:1.0.0
lns run
  Image:     ghcr.io/acme/reviewer:1.0.0 (resolving…)
  ...

ghcr.io/acme/reviewer:1.0.0 mounts into the workload:
  Host bind: /Users/you/some-project → /workspace — the workload can read and write this host directory
  Volume:    review-cache → /cache — the workload can read and write this machine's persistent volume
  Fileset:   ghcr.io/acme/reviewer-skills@sha256:9f2ab314c0de… → /root/.agent/skills — author-published files the workload will read
Continue? [y/N]:
```

- Enter (or `n`) — the default — aborts: `Error: run declined; nothing was mounted or started`. Nothing is shared, attached, or booted.
- `y` proceeds into the existing flow (per-secret KEEP/DROP scan on the bind, then boot).

Piped/detached (no terminal), without `--yes`:

```
$ lns run ghcr.io/acme/reviewer:1.0.0 </dev/null
...
ghcr.io/acme/reviewer:1.0.0 mounts into the workload:
  Host bind: /Users/you/some-project → /workspace — the workload can read and write this host directory
  ...
Error: ghcr.io/acme/reviewer:1.0.0 mounts the paths above into the workload and there is no terminal to confirm — run interactively, or pass --yes to accept them
```

Scripted opt-in: `lns run --yes ghcr.io/acme/reviewer:1.0.0` skips the prompt entirely.

## Implementation

The prompt logic lives in a new pure module `crates/lns-cli/src/run/pull_confirm.rs` (Layer 3 tested at 100%, no IGNORES entry), wired into `run_image` for `RunTarget::Reference` targets only, ahead of the existing per-secret KEEP/DROP scan.

## Tests

- **Layer 3**: full behavior matrix in `pull_confirm.rs` (no-mounts silence, disclosure lines per kind, default-No, explicit decline, fail-closed pointing at `--yes`, `--yes` silence).
- **Layer 1, virt-free** (`features/run_pull_confirm.feature`): a real pushed bind+volume sandbox and a fileset sandbox each fail closed without a terminal, disclose their mounts, and never reach `started run`.
- **Layer 1, `@microvm`**: the pulled-published-sandbox scenario now runs with `--yes` (required — the harness has no TTY, so the gate would otherwise fail it closed) and additionally asserts the gate stays silent; verified locally on Vz through a real boot with the registry kept online.

## Pre-existing gap surfaced (not introduced here)

The `@microvm` scenario "a pulled published sandbox launches offline from the consumer project" (landed on main today in fdfb2961, never yet run by a nightly) fails **before this PR's gate**: the CLI preflight inspects the tag reference, and tags always re-resolve against the registry (`image/manifest_cache.rs` — by design, so republishes stay visible), so the deliberately-offline registry 503s. Offline run-by-tag of a pulled sandbox needs a daemon-side tag→digest fallback (e.g. via the sandbox index) — separate fix. With the registry online, the scenario passes end-to-end including the `--yes` assertions.

Follow-up worth filing separately: scope the volume store per reference so an author-chosen volume name can't collide with another project's data.

### DIFF
--- a/crates/lns-cli/src/service/orchestrator.rs
+++ b/crates/lns-cli/src/service/orchestrator.rs
@@ -245,6 +245,23 @@ pub async fn run_image(
 
     let (volumes, bind_specs) = crate::cli::split_mounts(&args.mounts);
     let interactive = host_binds_interactive(args.detach, crate::raw_mode::stdin_is_tty());
+    if published.is_some() {
+        let reference = target.image();
+        let mounts = crate::run::pull_confirm::PulledMounts {
+            reference: &reference,
+            binds: &bind_specs,
+            filesets: &args.filesets,
+        };
+        let stdin = std::io::stdin();
+        let mut input = stdin.lock();
+        crate::run::pull_confirm::confirm_pulled_mounts(
+            &mounts,
+            args.assume_yes,
+            interactive,
+            &mut input,
+            &mut std::io::stderr(),
+        )?;
+    }
     let resolved_binds = resolve_host_binds(&bind_specs, interactive)?;
     if !quiet {
         let dispositions = crate::run::summary::format_bind_dispositions(&resolved_binds);
--- a/crates/lns-cli/src/service/orchestrator.rs
+++ b/crates/lns-cli/src/service/orchestrator.rs
@@ -250,6 +250,7 @@ pub async fn run_image(
         let mounts = crate::run::pull_confirm::PulledMounts {
             reference: &reference,
             binds: &bind_specs,
+            volumes: &volumes,
             filesets: &args.filesets,
         };
         let stdin = std::io::stdin();
--- a/crates/lns-cli/src/service/orchestrator.rs
+++ b/crates/lns-cli/src/service/orchestrator.rs
@@ -127,13 +127,6 @@ struct PublishedTarget {
     filesets: Vec<(String, String)>,
 }
 
-#[derive(Debug, PartialEq, Eq)]
-struct PublishedMountDisclosure {
-    volumes: Vec<lns_ipc::VolumeMount>,
-    binds: Vec<lns_ipc::BindSpec>,
-    filesets: Vec<(String, String)>,
-}
-
 fn published_target(
     reference: &str,
     inspection: lns_ipc::ArtifactInspection,
@@ -163,22 +156,6 @@ async fn preflight_published(socket: &Path, reference: &str) -> Result<Published
         image: reference.to_string(),
     };
     await_published_preflight(reference, real::send_request(socket, &request)).await
-}
-
-fn published_mount_disclosure(
-    published: Option<&PublishedTarget>,
-    cwd: &Path,
-) -> Result<Option<PublishedMountDisclosure>> {
-    let Some(published) = published else {
-        return Ok(None);
-    };
-    let resolved = crate::run::declarative::resolve(&published.defaults, cwd, None, Vec::new())?;
-    let (volumes, binds) = crate::cli::split_mounts(&resolved.mounts);
-    Ok(Some(PublishedMountDisclosure {
-        volumes,
-        binds,
-        filesets: published.filesets.clone(),
-    }))
 }
 
 async fn await_published_preflight(
@@ -227,6 +204,11 @@ pub async fn run_image(
         (_, Some(published)) => published.defaults.clone(),
         _ => crate::run::declarative::Defaults::default(),
     };
+    let consumer_mount_targets: Vec<String> = args
+        .mounts
+        .iter()
+        .map(|mount| mount.target().to_string())
+        .collect();
     let resolved = crate::run::declarative::resolve(
         &defaults,
         target.project_dir().unwrap_or(&cwd),
@@ -258,7 +240,6 @@ pub async fn run_image(
         (_, Some(published)) => published.filesets.clone(),
         _ => Vec::new(),
     };
-    let pull_confirm_mounts = published_mount_disclosure(published.as_ref(), &cwd)?;
     let quiet = args.quiet;
     let resolved_policy = if quiet {
         let (path, _source) = crate::run::summary::resolve_policy(args.policy.as_deref(), &cwd)?;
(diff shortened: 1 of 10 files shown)


chenhunghan:
@copilot code review against main

crates/lns-cli/src/service/orchestrator.rs
```diff
@@ -158,6 +165,22 @@ async fn preflight_published(socket: &Path, reference: &str) -> Result<Published
...
 }
 
+fn published_mount_disclosure(
+    published: Option<&PublishedTarget>,
+    cwd: &Path,
+) -> Result<Option<PublishedMountDisclosure>> {
+    let Some(published) = published else {
+        return Ok(None);
+    };
+    let resolved = crate::run::declarative::resolve(&published.defaults, cwd, None, Vec::new())?;
```
jansav: **Disclosure ignores `-v` overrides, so the prompt can name a mount that won't happen.**

`published_mount_disclosure` re-resolves `published.defaults` with `Vec::new()` as the explicit-mount set, so it always discloses the author's *full* declared set. But the actual run (the `resolve(...)` a few lines above the call site, then `args.mounts = resolved.mounts`) resolves with the consumer's `-v` mounts, and `declarative::resolve` drops any author-declared mount whose target the consumer overrode.

So `lns run <ref> -v /tmp/safe:/workspace` — when the author declared a bind at `/workspace` — prints `Host bind: /Users/you/some-project → /workspace — the workload can read and write this host directory` even though the cwd is never mounted (only `/tmp/safe` is).

**Why:** This prompt exists to state "what it will mount on this machine" so the consumer can consent to reality. Here it names a specific host path with a specific access mode that is factually wrong for that run — a consumer could decline a safe run, or accept believing their cwd is exposed when it isn't. It's fail-safe (never under-discloses a mount that *does* happen), but it's a real inaccuracy in the one prompt this PR adds, and no test pins the override case. Fix: disclose the author-origin mounts that actually survive the merge (resolve with the consumer's explicit mounts as the override filter, then exclude the consumer's own `-v` entries from the disclosed set) rather than resolving against an empty explicit set.

crates/lns-cli/src/service/orchestrator.rs
```diff
@@ -158,6 +165,22 @@ async fn preflight_published(socket: &Path, reference: &str) -> Result<Published
     await_published_preflight(reference, real::send_request(socket, &request)).await
 }
 
+fn published_mount_disclosure(
```
jansav: **Pure logic placed in a coverage-IGNORED file escapes the ratchet.**

`published_mount_disclosure` is a pure function, but `orchestrator.rs` is in the IGNORES table (`scripts/coverage-floor.sh`), so although you've tested it, its lines won't be held to the 100% floor as the file evolves — a future change to this logic could silently lose coverage. Move it beside `pull_confirm.rs` (or another gated module) so the new logic lands under the coverage gate the same way `pull_confirm.rs` does.

jansav (commented):
**Problem:** A pulled-by-reference sandbox's author-declared host binds, named volumes, and filesets were mounted into the workload with no consent gate — a relative bind like `.` silently virtiofs-shares the consumer's cwd, an author-chosen volume name attaches from the consumer's unscoped volume store, and filesets materialize author files. Only a `--quiet`-suppressible summary existed.

**Approach:** A new pure module `pull_confirm.rs` discloses every bind/volume/fileset and prompts `Continue? [y/N]` (default No) before anything is mounted, wired into `run_image` for `RunTarget::Reference` targets only, ahead of the KEEP/DROP scan. Fails closed with no terminal, `--yes` bypasses, local `lns.yaml` runs are unaffected, and the disclosure is independent of `--quiet`. Well-tested: Layer 3 behavior matrix at 100% (no IGNORES), plus virt-free Layer 1 scenarios.

**Product alignment:** Squarely on the "control access into the sandbox" and "real secrets stay outside the workload" principles — a pulled artifact can no longer silently reach into the consumer's machine.

The core security property holds: the disclosure never *hides* a real mount. Two items to address below.

**Verdict:** Needs fixes.

crates/lns-cli/src/service/orchestrator.rs
chenhunghan: Fixed — the disclosure is now computed from the mounts that actually survive the merge: `run_image` captures the consumer's explicit targets before `resolve()`, and `artifact_declared_mounts` filters the resolved set down to the author-origin entries, so `-v /tmp/safe:/workspace` neither discloses nor prompts for the dropped declared bind. The override case is pinned by `a_consumer_override_of_a_declared_target_is_not_disclosed`, plus tests for the declared-survivor and consumer-added-mount cases. (commit 542ea552)

crates/lns-cli/src/service/orchestrator.rs
chenhunghan: Fixed — the pure logic moved into `run/pull_confirm.rs` as `artifact_declared_mounts`; the orchestrator now only captures the explicit-mount targets and delegates. `pull_confirm.rs` holds at 100% under the full-workspace coverage gate (verified with `make coverage`), so the new logic sits inside the ratchet like the rest of the module. (commit 542ea552)

jansav (approved):

PR #158, opened 2026-07-22
docs(examples): add Codex CLI sandbox recipes for ChatGPT sign-in and API-key auth

chenhunghan:
## Summary

`docs/examples/` had one agent recipe (Claude Code). This adds the equivalent for OpenAI's Codex CLI — twice, because Codex has two auth models with genuinely different sandbox shapes:

### 1. `codex-chatgpt-subscription/` — Sign in with ChatGPT

Codex's ChatGPT-plan login is a browser OAuth flow with a `localhost:1455` callback, which two existing lns features carry:

- `ports: [container: 1455]` — a local `./lns.yaml` run publishes declared ports on host loopback, and the guest-side forwarder connects to guest-loopback, so the host browser's redirect lands on Codex's callback server inside the guest.
- A `codex-home` named volume over the workload home — Codex manages its own token set in `~/.codex/auth.json` (it decodes the JWT claims and runs the refresh flow itself), so there is no env var or header for the boundary to placeholder-inject. The volume persists the login, the trust decision, and the ~100MB binary: sign in once, restarts boot straight into the agent. `lns volume rm codex-home` signs out / forces a binary update.

The base is `curlimages/curl` (alpine + curl) rather than `node:lts-alpine`: Codex is a single static Rust binary fetched from its GitHub releases (no npm), and busybox `wget` can't do HTTPS through the egress proxy while `curl` can.

### 2. `codex-openai-api-key/` — usage-based billing, stateless

The thin variant: no callback port, no volume, and the real key never enters the guest. A `spec.credentials` slot seeds the `OPENAI_API_KEY` placeholder at boot with nothing pre-armed, the boot command pipes it into `codex login --with-api-key`, and the first request to `api.openai.com` raises the approval card where the key is bound per-machine — plain `lns run` with no setup command.

### Seed-config gotcha both examples encode

The workload runs as the image's `USER` (`curl_user`, home `/home/curl_user`), so a fileset seeded at `/home/sandbox` (the claude-code pattern) is invisible. The subscription example seeds at `/opt/codex-seed` and copies into `$HOME` at boot (the volume would shadow a home-mounted fileset); the API-key example mounts the fileset at `/home/curl_user` directly. The seeded `config.toml` pre-trusts `/workspace` and turns Codex's own sandbox off — the microVM is the sandbox.

## Docs

- Both example directories: `lns.yaml`, seed `config/.codex/config.toml`, `README.md` (sign-in mechanics, publish-to-registry caveats — a pulled sandbox's declared ports need `-P`, a consumer's first run raises the credential card).
- `docs/README.md` and `docs/running-workloads.md` link both examples.

## Test instructions

1. ChatGPT variant, first run: copy `docs/examples/codex-chatgpt-subscription/{lns.yaml,config}` into a project, `lns run`. Expect: binary downloads through the `github.com`/`release-assets.githubusercontent.com` routes, no trust prompt, onboarding offers **Sign in with ChatGPT**; opening the printed URL in the host browser completes via `localhost:1455` and Codex reaches the prompt.
2. ChatGPT variant, restart: `lns run` again. Expect: no download, no sign-in, no trust prompt — straight to the agent (verified: 12.5s first boot vs 2.2s warm).
3. Sign-out: `lns volume rm codex-home`, then `lns run`. Expect: fresh download + sign-in again.
4. API-key variant: copy `docs/examples/codex-openai-api-key/{lns.yaml,config}` into a project, `lns run` with no prior `lns integration connect`. Expect: `codex login --with-api-key` reports `Logged in using an API key - sk-LNSPL***00000`; the first model request pauses on the `openai` approval card, and after binding a key the request completes.
5. Offline check: `lns sandbox validate` in each example directory passes.


docs/examples/codex-openai-api-key/lns.yaml
```diff
@@ -0,0 +1,32 @@
...
+    allowedRoutes:
+      - match: github.com
+        verdict: allow
+      - match: release-assets.githubusercontent.com
+        verdict: allow
+      - match: api.openai.com
+        verdict: allow
+  credentials:
+    - name: openai
+      env: OPENAI_API_KEY
```
jansav, line 25: I think we should prefer connectors for these. Allows users to use them easier for other sandbox artifacts as well.

docs/examples/codex-openai-api-key/README.md
```diff
@@ -0,0 +1,71 @@
...
+lns run
+```
+
+No setup command: the `spec.credentials` slot seeds the placeholder at boot,
+and the first request that carries it to `api.openai.com` pauses on an
+approval card where you bind your real key — use the value detected on the
+host or paste one. The decision is per-machine, stored in
+`~/.lns-credentials.json`, and remembered for every later run; the key never
+lives in the guest, the project, or a published artifact. To bind it ahead of
+the first run instead (or to re-bind later), `lns integration connect openai`
```
jansav, line 49: `lns integration connect` doesn't exist — the noun was renamed to **connector** a while back, and the subcommand registers as `connector` (`crates/lns-cli/src/connector/mod.rs:147`). Every other doc says `lns connector connect <id>` (`docs/credentials.md:35`, `docs/cli-reference.md:295`).

**Why:** this is the one imperative instruction in the "bind it ahead of the first run" path, and a reader copy-pasting it gets an unrecognized-subcommand error.

docs/README.md
```diff
@@ -22,6 +22,12 @@ You drive everything through one binary: the `lns` CLI.
   ports, interactive vs. detached sessions.
 - **[Example: Claude Code](examples/claude-code/)** — a complete agent recipe:
   manifest, seed config, network allowlist, and credential wiring.
+- **[Example: Codex CLI (ChatGPT subscription)](examples/codex-chatgpt-subscription/)** —
+  the same recipe shape for OpenAI's Codex CLI, signed in with a ChatGPT plan
+  through a published callback port.
+- **[Example: Codex CLI (OpenAI API key)](examples/codex-openai-api-key/)** —
+  the stateless variant: usage-based billing via the `openai` integration, the
```
jansav, line 29: Same pre-rename vocabulary as the `lns integration connect` line — "integration" was renamed to "connector", so this needs adapting here plus `docs/examples/codex-openai-api-key/README.md:5` and `docs/examples/codex-chatgpt-subscription/README.md:85`.

**Why:** `docs/` is the source of truth for terminology; the manifest key is `connectors`, the CLI noun is `connector`, and `docs/connectors.md` is the page these sentences point at. A second name for the same concept in the two newest examples is what future readers will copy.

jansav (commented):
**Problem:** `docs/examples/` had only a Claude Code recipe; this adds two Codex CLI recipes covering its two genuinely different auth shapes — browser OAuth needing a published callback port plus a persistent volume, vs. a stateless boundary-injected API key.

**Approach:** Two sibling example directories, each with `lns.yaml`, a seed `config/.codex/config.toml`, and a README; both encode the image-`USER` home gotcha (`/home/curl_user`) and turn Codex's own sandbox off because the microVM is the boundary.

**Product alignment:** Fits "policy you run into, not write" and "real secrets stay outside the workload" — the API-key variant is a good demonstration of the placeholder model, and the subscription variant is honest about where the placeholder model can't apply.

I checked the manifests against the schema and runtime rather than the prose: `spec.credentials[{name, env}]` / `ports[{container}]` / `filesets[{path, mountPath}]` all parse and no mount targets collide; the `openai` catalog entry is `OPENAI_API_KEY` + `bearer_header` on `api.openai.com`, matching the slot; local-definition auto-publish and the `-P`-for-pulled caveat are exactly what `declarative_ports.feature` pins, and the guest forwarder dials `127.0.0.1`, so the `localhost:1455` callback claim holds; a pristine named volume is seeded from the image dir and chowned to the run-as ids, so `codex-home` over `/home/curl_user` is writable by `curl_user`; and filesets materialize as workload-owned guest writes, so `codex login` can write `auth.json` under the `/home/curl_user` fileset. The recipes are sound.

**Verdict:** Needs fixes — both findings are the same thing: these examples are written against the pre-rename "integration" vocabulary.

PR #166, opened 2026-07-23
feat: scope connector access to per-workload grants

jansav:
## Problem

Connecting a connector was effectively global. The secret value is machine-global (`~/.lns-credentials.json`), and while "connected" is recorded per-directory in `lns-policy.yaml`, one project directory hosts many sandboxes/workloads — so a connector connected for one workload silently armed for **every** workload run there. Worse, `lns-policy.yaml` is committable, so a cloned repo listing `connectors: [github]` plus your machine-global token could spend your credential with no prompt.

## What this does

Scopes a connector **grant** to `(project × workload × connector)`, remembered in a new machine-local sidecar `~/.lns-workload-grants.json` (never committed; precedent: `~/.lns-host-bind-decisions.json`). The secret **value** stays machine-global — the password-manager model: one vault entry, many per-use permission decisions.

A connected connector now arms only when the running workload holds a fresh allow-grant; otherwise it is demoted to a first-use offer card (reusing the existing declared-not-connected machinery), and a denial is remembered per-workload.

**Workload identity:** a local definition keys by its (canonicalized) directory — every `-f`/`-- cmd` variant of one project shares it, and editing the file does not re-prompt; a published image keys by `repo@digest`, so a republished or swapped digest re-prompts. A run that resolves neither is refused rather than falling into a shared bucket.

### Behavior
- **First run of a new workload** in a project with a connected connector → one first-use offer card disclosing the env var + injection domains. Accept arms and remembers it; deny suppresses it for the rest of the run and future runs.
- **Re-runs** of a remembered workload → no prompt.
- **A different or foreign workload** in the same project → its own offer; it does not inherit another workload's grant.
- **Required credential slots** (`spec.credentials`) gate the same way: a bound slot injects only where the workload holds a grant. A required `oauth` slot's launch-blocking sign-in *is* the grant — completing the device flow records it before the boot gate runs, so the workload starts armed instead of meeting a second card seconds later.
- **Upgrade migration:** the first run of a project's *own definition* grandfathers its already-connected connectors, so existing primary workflows are never re-prompted; foreign workloads still get the card. (History-free — there is no durable run log.)
- **CLI:** `lns connector grants [--all]` lists remembered grants, `lns connector revoke <id>` forgets a connector's grants in a project, and `lns connector disconnect` now clears them too. `lns connector connect` reports when the project still holds a standing decline that binding a value cannot lift, and names the `revoke` that does.
- **Approval window:** an oauth card for an already-bound connection now offers **Reconnect** alongside spending it, so the wrong account or a server-side revocation isn't a dead end whose only exit is a standing per-workload deny.

### Safety
- A card **deny** no longer writes a machine-wide `CredentialEntry::Deny`; it is a per-workload sidecar deny (honored by verdict alone, so it sticks even for a connector with no injection wiring).
- A grant persists only once the value it was given for actually reaches the credential store, so a failed store write can't leave an allow behind that arms the next run silently.
- Grant writes are read-modify-write under an exclusive sidecar lock, so a concurrently running sandbox's grants are never clobbered. The wait for that lock is bounded (a wedged holder in another process can no longer park a tokio worker on the path that answers credential cards), and a kernel that refuses to lock at all — a network mount — surfaces as an error rather than a guard nobody holds.
- **Disconnect wins the race with a card.** The sidecar counts forgets per connector; a decision persists only if the count hasn't moved since the card went up. So a `lns connector disconnect` landing mid-decision arms that run in front of the developer and records nothing, instead of resurrecting the grants the command just reported forgetting. A required slot's minutes-long boot sign-in is covered by the same counter, captured before the gate opens.
- Two credential slots naming one connector are rejected at parse: arming is per id, so one slot's grant would otherwise inject the secret under an env var the consent card never named. The gate correspondingly requires *every* provider resolved under an id to match the grant's disclosure before it arms.
- A decision is remembered in-memory even when the sidecar cannot be persisted, and the user is told on the approval window; a corrupt/unreadable sidecar fails safe (re-offer) and is never overwritten.
- The armed reconciler re-reads the sidecar on every reload rather than holding its boot snapshot, so a reconnect after a disconnect meets a fresh card. The gate keeps the project's full shareable `connectors:` list on the session policy, so a mid-run persist round-trips losslessly; the watcher-reload route deriver gates the same way, while a connector connected live this run survives the reload it triggers and is revoked on disconnect.
- The grant snapshot (env var + injection domains, re-offered if the catalog redefines them) is the first concrete step toward the `(id, env_var, domain)`-bound approvals direction.

### Also in here
- `fix(coverage)`: `coverage-strip-ast` stopped demanding coverage of lines that hold no expression — a macro-argument line carrying only literals/punctuation, and `} else {`. Both were gated with no test able to answer them.

## Docs
`docs/credentials.md` gains a **Workload grants** section; `docs/connectors.md` and `docs/cli-reference.md` are updated for the connect-≠-arm split and the new subcommands.

## Deferred to follow-ups
- **`lns connector connect --workload <ref>`** — headless pre-grant, so a CI-style run can be granted without an interactive card.
- **`lns connector deny --everywhere`** — an explicit writer of the machine-wide `CredentialEntry::Deny` the card no longer writes.
- **Routing the network-offer card's raise through the credential session**, so it pins its forget count when the card goes up like the credential path does, rather than comparing against the run's start-of-life count.
- **`project_key` parent-canonicalization** for the deleted-policy edge case.
- Snapshot fidelity for a malformed `ApiKeyHeader` injection, and the leak-attempt-deny-overwrites-allow semantic — both currently fail safe.

## Testing
Full workspace gate green: `make lint`, `make complexity`, `make coverage`. Every touched file is at 100% line coverage. Per-workload gating, grant recording, the disconnect races, the migration seed, and the CLI behaviors are pinned by Layer-2 Gherkin (`connector_grants.feature`, plus the credential/declared-connector features) and Layer-3 tests, with synthetic fixtures throughout.


chenhunghan (approved):
**Problem:** A connector "connected" in a directory armed its machine-global secret for every workload run there — and since `lns-policy.yaml` is committable, a cloned repo listing `connectors: [x]` could spend your credential with no prompt.

**Approach:** Grants scoped to (project × workload × connector) in a machine-local, flock-serialized sidecar, pinned to the connector's disclosure snapshot so a redefinition invalidates rather than inherits. Arming is re-gated at boot and on every reload; disconnect-vs-card races are settled by a forget counter compared inside the locked update. The race handling (mid-sign-in disconnect, unreadable sidecar failing closed, grants held back when the value write fails) is carefully reasoned and each case is pinned by a test.

**Verdict:** Ship it. ✅

One note on the description, not the code: the **Upgrade migration** bullet ("the first run of a project's own definition grandfathers its already-connected connectors") describes behavior that isn't in the diff — and rightly so. With no durable history, a rule that auto-grants a project's own definition would also auto-grant a cloned repo's own definition, which is exactly the attack this PR closes; the shipped behavior (one first-use card per existing workload after upgrade, then never again) is the safe one, and `docs/credentials.md` already describes it. Suggest replacing that bullet with an honest upgrade note so post-upgrade re-prompts aren't read as a bug — the deferred `lns connector connect --workload <ref>` remains the safe path if fleet upgrade friction ever matters.

PR #170, opened 2026-07-24
feat: seed a declared connector's placeholder env at launch

chenhunghan:
## Problem

Since 219ac29c ("offer artifact-declared integrations reactively instead of arming them"), a `spec.connectors` id was pure disclosure: nothing was seeded into the workload env. An agent CLI that checks for its key variable at startup (e.g. `claude` checking `ANTHROPIC_API_KEY`) exits before making its first request — so the reactive on-domain connect offer that replaced arming never gets the chance to fire.

## Change

**1. Declaring a connector seeds its placeholder env var** while staying unarmed:

| Tier | Placeholder env seeded | Routes armed | Value armed at boundary |
|------|------------------------|--------------|------------------------|
| Connected in directory overlay | yes | yes | yes |
| **Artifact-declared, not connected** | **yes** (this PR) | no — connect offer on first use | **no — consent at first use** |
| Merely connectable (undeclared catalog) | no | no | no |

The detect-only flip moves from `supervisor/adapter.rs` into `resolve_connectable_with_declared`, where the declared set is already known: a declared id keeps `seed_env`, everything else becomes detect-only at resolve time.

**2. Env seeding and value arming are separate axes** (review finding: seeding alone would have let an untrusted artifact spend an already-bound machine credential by declaring the connector and pre-allowing its route — the placeholder travelled, the store-armed injection swapped in the real value, and the route cleared policy with no card).

The wire expansion (`credential_flow/registry.rs`) now takes the run's **armed ids** — overlay-connected connectors and credential slots at boot — and serializes unarmed injections for every other provider regardless of store state, so the real value never reaches the wire struct without per-directory consent. The armed set lives on the `CredentialSession` and is granted **live** at every consent tail (`arm_connected`, a connect-now decision, `connect_connector_now`); the policy emitter receives it per emit, so accepting a connect card arms the credential on the next frame without a relaunch. `is_armed_for_request` requires the grant too, closing the store-only auto-allow in `submit_pending` for unconsented ids.

What 219ac29c fixed stays fixed: no route past the local ceiling, no value without re-consent. `-e` handling is untouched.

## Tests

- **Guards (green before the fix):** the declared rig exposes the run's wire credentials; a connected connector's stored value must arm at launch. The reactive connect-then-arm chain was already pinned by the oauth/pkce connector features; a new session test pins that a live connect grants arming *before* the policy emit (the regression a boot-frozen flag would cause).
- **Exploit pinned shut (red → green):** Layer 2 — stored machine value + declared connector + artifact-pre-allowed route + empty overlay ⇒ placeholder seeded, route allowed, **boundary injection unarmed**, connector still offered; plus the undeclared-connectable sibling. Layer 3 — stored and host-detect values outside the armed set stay unarmed; an unconsented pending request still prompts instead of auto-allowing.
- Env-seeding scenarios from the first commit (declared seeds placeholder, phantom-placeholder boundary) unchanged.

No `lns.yaml` schema, wire-format, catalog, or policy-file changes.


chenhunghan:
@copilot+gpt-5.6-sol code review in terms of security, post your comments on pr

chenhunghan:
**Problem:** Declared connectors need to seed placeholders so CLIs start and trigger reactive connection.

**Approach:** Connector resolution now preserves `seed_env` for declared connectors while keeping undeclared catalog entries detect-only.

**Product alignment:** The intent fits reactive consent, but the implementation can expose a stored credential without that consent.

**Verdict:** Needs fixes.

### Critical — can expose a bound credential without consent

**Declared connectors are serialized with already-armed injections** (`crates/lns-service/src/credential_flow/connectors.rs:240`)

A seeded `DefProvider` is later resolved against the machine credential store, and stored, host-detected, or OAuth values become armed injections. If the connector's destination is already permitted by the sandbox policy, no network connect card appears: the placeholder is replaced and the request leaves with the real credential.

An untrusted artifact can therefore declare a connector and allow its domain to spend an existing machine-bound credential without per-directory consent. Environment seeding must be separated from injection arming so declared-but-unconnected connectors remain unarmed until accepted.

Please add a behavioral scenario covering: stored credential + declared connector + pre-allowed destination + connector absent from the directory overlay.

All CI checks and the 157-scenario service behavior suite pass; the current tests do not model stored credential resolution for declared connectors.

crates/lns-service/src/supervisor/adapter.rs
```diff
@@ -507,14 +513,14 @@ pub(super) async fn start(
         .collect();
     let offerable = build_offerable(&connectable, &catalog);
     let connectable_routes = Arc::new(connectable.routes);
+    // Only the boot-consented ids (overlay-connected + credential slots) may arm a machine-stored value; a connect grants more live.
+    let armed_ids: HashSet<String> = applied
```
jansav: The security-load-bearing composition — `armed_ids` = applied-provider ids, run providers = applied ∪ connectable — now exists twice as inline logic: here and re-implemented in the Layer 2 rig's `launch()` (`tests/behaviours/steps/declared_connectors.rs:136`). The rig calls the same pub `resolve_*` functions for everything else, but this derivation is a hand-maintained mirror, and the "boundary injection stays unarmed" scenarios assert against the mirror. The only tests that execute this copy (`supervisor/mod.rs` `start_if_policy_*`) assert nothing about arming.

**Why:** if `start()` later widens `armed_ids` — e.g. someone "fixes" the extra consent card that now appears when a connectable id has a stored value by adding connectable ids at boot — every test stays green while the exploit this PR closes reopens. Extract the composition into one pub helper (or a small `RunProviders { providers, armed }` built from `AppliedConnectors` + `ConnectableConnectors`) and call it from both `start()` and the rig, so the Layer 2 scenario pins production.

jansav (commented):
**Problem:** Since 219ac29c a declared connector was pure disclosure — no placeholder seeded — so an agent CLI that checks its key env var at startup exits before the request that would trigger the reactive connect offer. This PR seeds the placeholder for declared connectors and, forced by that, splits env seeding from value arming so a malicious artifact can't declare a connector plus a pre-allowed route to silently spend a machine-bound credential.

**Approach:** The detect-only flip moves into `resolve_connectable_with_declared` where the declared set is known. A per-run `armed` id set on `CredentialSession` (boot: overlay-connected + slots; grown live at every consent tail) gates the single production site that turns a stored value into a wire injection, plus the `submit_pending` auto-allow. Every consent tail grants before its policy emit, and the ordering is pinned by a test. The exploit is pinned red→green at Layer 2 with Layer 3 siblings for stored/host-detect values outside the armed set.

**Verdict:** Needs fixes — one finding on where the exploit-closing test anchors.

crates/lns-service/src/supervisor/adapter.rs
chenhunghan: Extracted in 42035cf0: `run_providers()` in `credential_flow/connectors.rs` is now the single composition — `RunProviders { providers, armed, connectable_ids }` — called by both `start()` and the Layer 2 rig's `launch()`, so the "boundary injection stays unarmed" scenarios assert against the derivation production actually runs. A unit test additionally pins the composition itself (only applied ids in `armed`), so widening it at boot fails a named test rather than silently reopening the exploit.

crates/lns-service/src/supervisor/adapter.rs
```diff
@@ -368,6 +384,13 @@ const OAUTH_REFRESH_SKEW_SECS: u64 = 60;
 const WINDOW_NOT_INSTALLED: &str = "approval window state was not installed at boot; tray::run_tray must run before any policy-bearing run starts";
 
 /// The per-connector oauth wiring a run hands to its credential subsystem: device-flow configs, display names, and token fallbacks, all keyed by connector id.
+/// The run's consent boundary at boot: which ids may arm a stored value (`armed`), which of those survive a policy reload (`reload_grants` — the credential slots), and which are offered for a live connect (`connectable_ids`).
```
jansav: `CredentialConsent` was inserted between `OauthWiring`'s doc comment and the struct, so `CredentialConsent` is now documented as "the per-connector oauth wiring…" and `OauthWiring` has no doc at all. Wrong documentation on the run's consent-boundary type — move the oauth-wiring sentence back onto `OauthWiring`.

crates/lns-service/src/approval_flow/session.rs
```diff
@@ -136,6 +141,11 @@ impl ApprovalSession {
     }
 
     /// Installs the connector-route deriver once at boot so a watcher reload re-applies a connected connector's routes instead of dropping them; idempotent, the first wins.
+    /// Installs the armed-reconciler once at boot; idempotent, the first wins.
```
jansav: Same insertion mistake as in adapter.rs: the connector-route-deriver doc ("re-applies a connected connector's routes…") now sits on `set_armed_reconciler`, and `set_connector_route_deriver` is undocumented. A reader of the reconciler installer gets the deriver's contract — move the deriver sentence back onto `set_connector_route_deriver`.

jansav (commented):
**Problem:** Since declared connectors became reactive-offer-only, a `spec.connectors` id seeded nothing into the workload env — an agent CLI that checks its key variable at startup exits before the first request, so the reactive connect offer never fires. Seeding alone would have reopened an exploit: declared connector + artifact-pre-allowed route would let a machine-stored credential inject with no consent card.

**Approach:** Env seeding and value arming become separate axes. A declared connector keeps `seed_env` at resolve time (undeclared connectables go detect-only), while a per-run armed set on `CredentialSession` — applied overlay/slot ids at boot, grown live at every consent tail, reconciled down on policy reload with slot grants preserved — gates whether the wire expansion ever resolves a stored value. Traced adversarially: the floor's `connectors` can't re-arm via reload (`merge_effective` takes only the overlay's), the armed value is emitted before a connect releases network holds (pinned by ordering tests), the drain-under-lock makes cross-subsystem releases exactly-once, and the `expand_credentials_*` signature change makes it impossible for a call site to forget the armed set. Exploit closure pinned red→green at Layer 2 and Layer 3. Full `cargo test -p lns-service` is green (160 scenarios).

**Verdict:** Needs fixes — two misattached doc comments, both two-line moves; everything functional and the security model held up.

crates/lns-service/src/supervisor/adapter.rs
chenhunghan: Fixed in 0db0f262 — moved the oauth-wiring doc back onto `OauthWiring`; `CredentialConsent` keeps its own consent-boundary doc.

crates/lns-service/src/approval_flow/session.rs
chenhunghan: Fixed in 0db0f262 — the route-deriver doc is back on `set_connector_route_deriver`; `set_armed_reconciler` now carries its own contract.

crates/lns-service/src/credential_flow/session.rs
```diff
@@ -17,8 +17,8 @@ use lns_policy::connectors::TokenFallback;
 
 pub type FrameSink = mpsc::UnboundedSender<HostFrame>;
 
-/// Invoked after a state-changing decision so a follow-up `Policy` frame arms the matching injection.
-pub type PolicyEmitter = Box<dyn Fn(&CredentialStateFile) + Send + Sync>;
+/// Invoked after a state-changing decision so a follow-up `Policy` frame arms the matching injection; receives the run's armed ids because a stored value may only arm for a consented connector.
```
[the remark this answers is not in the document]
chenhunghan: Fixed in 0db0f262 — updated to "a resolved value (stored, oauth, or host-detect) may only arm for a consented connector."

crates/lns-service/src/credential_flow/registry.rs
```diff
@@ -10,18 +12,20 @@ pub fn detect_for_with(id: &str, custom: &[DefProvider]) -> Option<String> {
         .and_then(|p| p.detector().detect())
 }
 
-/// Every provider becomes one `Credential` even when unarmed so the MITM can detect every known placeholder in outbound requests; host detection resolves against the run's providers.
+/// Every provider becomes one `Credential` even when unarmed so the MITM can detect every known placeholder in outbound requests; a machine-stored value arms only for ids in `armed` — the connectors consented for this run — so a declared or connectable connector stays unarmed until connected.
```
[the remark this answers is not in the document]
chenhunghan: Fixed in 0db0f262 — the comment now says a resolved value (stored/oauth/host-detect) arms only for ids in `armed`.

crates/lns-service/src/supervisor/adapter.rs
```diff
@@ -368,6 +384,13 @@ const OAUTH_REFRESH_SKEW_SECS: u64 = 60;
...
+struct CredentialConsent {
+    armed: HashSet<String>,
+    reload_grants: HashSet<String>,
+    connectable_ids: HashSet<String>,
+}
+
 struct OauthWiring {
     configs: HashMap<String, crate::oauth::OauthConfig>,
```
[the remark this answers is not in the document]
chenhunghan: Fixed in 0db0f262 — oauth-wiring doc moved back to `OauthWiring`.

crates/lns-service/src/supervisor/adapter.rs
```diff
@@ -508,13 +508,7 @@ pub(super) async fn start(
...
     let mut custom = applied.providers;
-    // Connectable (undeclared) connectors stay known for an on-domain connect offer but are detect-only, so they never seed a phantom placeholder into the workload env.
-    custom.extend(
-        connectable
-            .providers
-            .into_iter()
-            .map(DefProvider::detect_only),
-    );
+    custom.extend(connectable.providers);
     let custom_providers = Arc::new(custom);
```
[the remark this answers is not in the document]
chenhunghan: Fixed in 0db0f262 — comment now reads "resolve as connectable — detect-only unless definition-declared — so their use offers a live connect."

jansav (approved):

PR #172, opened 2026-07-27
fix: only a decision the developer actually made is recorded

chenhunghan:
One rule, three places it was broken: **only a decision the developer actually made gets recorded.** A closed credential card, a closed network card, and a failed sign-in each recorded a verdict nobody picked.

## The bug

Closing a credential card with the ✕ wrote a permanent `{"kind":"deny"}` rule to `~/.lns-credentials.json`. After that, `submit_pending` short-circuits and the card **never appears again** — not in this run, not in any later run, and not in any other project, since that store is machine-wide. The workload's requests just fail silently at the boundary. The only way out was hand-editing the JSON.

One click on the pile's close-all did it to every credential in the stack at once.

Nothing pinned it: `close_action` and `close_all` live in `tray.rs`, which is coverage-exempt, and `credential_flow.feature` had no scenario for closing a card. The behaviour was never specified — it was wired.

## What changed

**A closed card is the absence of a decision.** No rule, no grant, no audit line, and the next use asks again. On the wire it reads as `Timeout` — the guest already treats that as "fail this one, keep gating".

**The close paths can no longer express a verdict.** `CardAction::DismissNetwork` / `DismissCredential` carry only an id, so there is nowhere to put a decision the developer didn't make. `close_all` routes through the same `close_action` mapping instead of deciding cards inline.

**Dismissal is its own variant.** `CredentialDecisionRequest` gains `Dismiss` alongside `Allow` / `AllowBound` / `Deny` / `DenyAlways` / `Timeout`. It records no grant, persists no rule, and earns no audit line — the three things a dismissal must not do.

**A deny reaches only as far as the card that asked** (hardening — no observable change). `Deny` meant a per-workload decline in the run flow but a machine-wide `CredentialEntry::Deny` in the bind flow, picked by which decision channel was attached. A prompt now carries a `DenyScope`; only the bind card is `Machine`, so it alone can emit `DenyAlways` and bind a standing refusal. The run flow narrows either variant to the workload sidecar, making it structurally incapable of widening a refusal.

**The dismissal set is closed as a type.** `close_action` returns a `Dismissal` rather than the 12-variant `CardAction`, so `apply_dismissal` is exhaustive and a new card kind is a compile error instead of a silently dropped request that hangs to the approval timeout.

**Sign-in failures stop reading as refusals.** `connect_oauth` recorded `Decision::Deny` for any unsuccessful sign-in. Now only a denial in the browser records an approval; cancel, expiry, and transport errors record none. This is the same rule as the dismiss fix applied to sign-ins rather than cards, which is why it rides along here.

**Network cards get the same treatment.** Closing one no longer logs a `deny_once` the developer never picked, and `dismiss_request` sends `Timeout` on the wire rather than `DenyOnce`. The guest fails the request closed either way, but core's `Decision::audit_reason()` writes the distinction into the JSONL trail, so a dismissed card now reads there as `decision-timeout` instead of `user-denied-once`.

## Reconciled with main's grant sidecar

This branch was opened before the per-workload grant sidecar landed on `main` (`a3a57727`, `d804661a`, `2afdf3c2`, `8be2654e`), which rewrote the same enum. The rebase keeps **main's** deny semantics rather than this branch's original ones:

- A card deny stays sticky, remembered per `(project, workload, connector)` in the grant sidecar, and audits as `Decision::Deny`.
- `main` had already removed the machine-wide `{"kind":"deny"}` write from the card-deny path — the bug this PR fixes on the *dismiss* path.
- The branch's original "card deny is once-only, standing deny only via `lns connector connect`" model is **dropped**, along with the two scenarios and the one unit test that pinned the once/always distinction.

What survives from this branch is the dismiss contract and the sign-in-outcome split; the deny model is main's.

## Tests

Failing tests landed before the fix (commit `f38d7c1f`), so the diff is red → green.

`egui_kittest` (dev-only, no wgpu/snapshot features) closes the widget-wiring gap: the test hovers a real card, finds the ✕ through the AccessKit tree, clicks it, and asserts what fired. That needed `close_button` and `pile_pill` to register `WidgetInfo` — two interactive controls that were unlabelled and invisible to assistive tech, fixed in its own commit.

Review round: four follow-up commits address five reviewer threads (exhaustive dismissal type, `DenyScope`/`DenyAlways` split, the wire-decision contradiction, one doc comment) — see the resolved threads for detail.

Five new Gherkin scenarios across `credential_flow`, `approval_flow`, and `connector_cli`, plus tightened audit assertions in `oauth_connector` / `pkce_connector`. Both Layer 2 rigs now capture ledger events so a scenario can assert the audit chain stayed silent.

## Not in scope

There is still no CLI verb to clear a machine-wide standing deny — `lns connector disconnect` cancels per-workload grants but doesn't touch a `{"kind":"deny"}` entry in `~/.lns-credentials.json`. `lns connector deny` / `lns connector forget` are worth adding; filing separately.

## Verification

`make lint` and `make complexity` are clean, and `cargo test --workspace --exclude e2e-tests --all-targets` passes — 1818 lns-service unit tests and 168/168 Layer 2 scenarios.

`make coverage` has **not** been run locally — relying on CI. The new `Dismiss` / `DenyAlways` match arms in `session.rs` / `bind.rs` and the `Dismissal` mapping in `tray.rs` are the likeliest places for the 100%-per-file gate to bite (`tray.rs` itself is IGNORE'd).


crates/lns-service/src/tray.rs
```diff
@@ -804,28 +820,33 @@ fn pile_scroll_offset(
...
+        CardAction::DismissNetwork { id } => {
+            state.dismiss(id);
+        }
+        CardAction::DismissCredential { id } => {
+            state.decide_credential(id, CredentialDecisionRequest::Dismiss);
+        }
+        CardAction::CancelSignIn { credential_id } => {
+            state.cancel_sign_in(credential_id);
+        }
+        _ => {}
```
jansav: **`apply_dismissal`'s catch-all reintroduces the silent-failure class this PR just closed**

The one function that is now the single dismissal path ends in `_ => {}`. Add a `StackItem` variant, wire it into `close_action`, and close-all silently drops it — no compile error, and the held request hangs to the 300s timeout. `tray.rs` is coverage-exempt, so nothing catches it.

**Why:** The PR's whole thesis is "make the type prevent the mistake." An exhaustive match here does that; the wildcard undoes it for the next card kind.

crates/lns-service/src/credential_flow/bind.rs
```diff
@@ -54,9 +54,13 @@ pub fn resolve_bind_decision(request: CredentialDecisionRequest) -> BindResoluti
             };
             BindResolution::Persist(entry, decision)
         }
-        CredentialDecisionRequest::Deny => {
+        // The bind card holds no request, so a deny here can only mean the machine-wide one; the reactive card's deny persists nothing.
+        CredentialDecisionRequest::DenyOnce => {
```
jansav: **`DenyOnce` persists a permanent machine-wide deny in the bind flow**

`resolve_bind_decision(DenyOnce)` → `Persist(CredentialEntry::Deny, Denied)`. The variant name asserts "once"; the comment above is the only thing distinguishing the two flows, and both cards render through the same `render_credential_card` with the same "Deny" button — they differ solely by which `decision_tx` was registered.

**Why:** This is the same failure shape as the bug being fixed: one durability decision routed by which channel happens to be attached, not by the type. Misroute it and the ✕-writes-a-permanent-deny bug returns via the Deny button. Give the bind flow its own variant (`DenyAlways`) so the type carries the distinction.

jansav (commented):
**Problem:** Closing an approval card with ✕ wrote a permanent, machine-wide `{"kind":"deny"}` to `~/.lns-credentials.json`, after which `submit_pending` short-circuited forever — in every project, recoverable only by hand-editing JSON. Close-all did it to the whole stack at once.

**Approach:** A closed card becomes the absence of a decision. `CardAction::DismissNetwork`/`DismissCredential` carry only an id, so the close paths structurally cannot express a verdict; `close_all` routes through the same `close_action` mapping. A card deny is now once-only — standing denies stay reachable only via `lns connector connect`. Sign-in cancel/expiry/transport errors stop recording as refusals. `egui_kittest` closes the widget-wiring gap by clicking the real ✕ through the AccessKit tree.

**Product alignment:** Correct. "Policy you run into, not write" is broken by an unreachable card you can't get back, and a machine-wide rule written by a ✕ violates "one directory = one project."

The core move — deleting the verdict from the type — is right, and the test-first ordering with the a11y labels as their own commit is exactly how this should have been done. Full `lns-service` suite verified green locally (1752 unit + 166 scenarios).

**Verdict: needs fixes.** Two inline, plus two findings on lines outside the diff:

### Should fix

**3. The tray.rs IGNORES reason is now false**
`scripts/coverage-floor.sh:90` (and `:92`)

The reason reads "the egui card renderers' click/interaction state needs a live frame + pointer input (no headless egui_kittest harness in-crate)." This PR adds exactly that harness and drives real clicks with it.

**Why:** CLAUDE.md makes the table the tracker — the reason names the design move that gets the file to 100%, and the move just happened. Leaving a stale justification means the next reader can't tell what actually still blocks it.

**4. Stale doc comment on the variant this PR redefined**
`crates/lns-service/src/tray.rs:515`

`/// Dismiss every card at once …, denying or cancelling each held request.` It no longer denies anything.

crates/lns-service/src/tray.rs
chenhunghan: Fixed — `close_action` now returns a `Dismissal` (Network / Credential / SignIn / Inform) instead of the 12-variant `CardAction`, and `apply_dismissal` matches it exhaustively, so wiring a new card kind into `close_action` is a compile error at the one place that has to handle it. `From<Dismissal> for CardAction` keeps the UI action pipeline unchanged.

While in there: informs and sign-ins now route through `apply_dismissal` too, rather than reaching past it into `WindowState`, so no arm of the new match is dead.

You were right that no test would have caught it. `close_all_takes_down_every_kind_of_card_in_the_stack` now builds a stack holding every card kind and asserts close-all empties it — I mutation-checked it by neutering the `SignIn` arm, and it fails as it should. (commit bd309d3)

crates/lns-service/src/credential_flow/bind.rs
chenhunghan: Fixed, and you called the shape correctly — durability was decided by which channel was attached, not by the type.

One note on scope: the variant you're looking at was named `DenyOnce` when you reviewed, but this branch has since been rebased onto main, where the per-workload grant sidecar landed (a3a5772, d804661, 2afdf3c, 8be2654). That rebase renamed it to plain `Deny` and removed the machine-wide write from the *run* flow — so the `Deny → Persist(CredentialEntry::Deny)` in `resolve_bind_decision` is inherited from main rather than introduced here. Your point survived the rename intact, and arguably got worse: the name no longer even hinted at the difference.

The fix goes a step past adding the variant, because a variant alone doesn't produce itself — the tray couldn't tell the two card kinds apart. So:

- `CredentialPendingPrompt` carries a `DenyScope`. `bind_prompt` is the only `Machine` one; everything else is explicitly `Workload`.
- The tray maps scope → request, so the bind card's Deny emits `DenyAlways` and a run card's emits `Deny`.
- `resolve_bind_decision` persists on `DenyAlways` and *refuses* a run-scoped `Deny` outright.
- The run flow narrows both `Deny` and `DenyAlways` to the workload sidecar, so it is now structurally incapable of writing a machine-wide deny — a future misroute fails safe instead of escalating. Pinned by `the_run_flow_cannot_widen_a_deny_past_the_workload_even_when_asked_to`. (commit 3c75dc5)

crates/lns-service/src/approval_flow/session.rs
```diff
@@ -257,6 +257,16 @@ impl ApprovalSession {
...
     }
 
+    /// Fails a held request because its card was closed: no rule, and no audit line claiming a deny-once the developer never picked.
+    pub fn dismiss_request(&self, id: &str) -> DecisionOutcome {
+        if self.remove_pending(id).is_none() {
+            return DecisionOutcome::UnknownId;
+        }
+        self.notifier.dismiss(id);
+        self.send_decision_frame(id, Decision::DenyOnce);
+        DecisionOutcome::Resolved
```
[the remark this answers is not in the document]
chenhunghan: Fixed — `dismiss_request` now sends `Decision::Timeout`. You're right that it contradicted the stated contract, and it was self-inconsistent two ways: the credential dismiss path already sent `CredentialDecisionKind::Timeout`, and the genuine approval timeout sends `Decision::Timeout`, which left the network dismiss as the only closer claiming a verdict.

The guest fails the held request closed on either value, so observable behaviour is unchanged — this is about what the frame *claims*. Two tests pinned the old value and now pin the new one (`a_dismissed_card_fails_the_request_closed_and_records_no_approval`, and the same assertion one layer up in `supervisor::adapter`). (commit 632edcc)

crates/lns-service/src/credential_flow/bind.rs
```diff
@@ -54,9 +54,13 @@ pub fn resolve_bind_decision(request: CredentialDecisionRequest) -> BindResoluti
...
             BindResolution::Persist(CredentialEntry::Deny, CredentialBindDecision::Denied)
         }
+        CredentialDecisionRequest::Dismiss => {
+            BindResolution::Failed("the value card was dismissed before a decision was made".into())
+        }
```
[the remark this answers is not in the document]
chenhunghan: Fixed by introducing `DenyAlways`, as suggested.

Scope note: the `DenyOnce` name you reviewed is gone — this branch was rebased onto main, where the per-workload grant sidecar landed and renamed it to plain `Deny`. So the misleading-name half of your comment resolved itself, but the substantive half didn't: `resolve_bind_decision` persisting a machine-wide `CredentialEntry::Deny` is inherited from main, and one variant still meant two different durabilities depending on the receiving flow.

`DenyAlways` is now the only request that binds a standing refusal, and it can only originate from a card whose `DenyScope` is `Machine` — which is just the bind card. A run-scoped `Deny` arriving at `resolve_bind_decision` is now refused rather than persisted, and the run flow narrows `DenyAlways` down to the workload rather than honouring it, so the "future call site accidentally persists a permanent deny" path you flagged is closed from both ends. (commit 3c75dc5)

crates/lns-service/src/tray.rs
```diff
@@ -2049,6 +2088,165 @@ mod tests {
         assert_eq!(deny_decision(true), Decision::DenyAlways);
     }
 
+    /// Drives `render_stack` through headless egui passes, hovers the stack so hover-only affordances appear, clicks the control labelled `label`, and returns whatever that click fired. `expanded` fans the pile out first so its header controls exist.
```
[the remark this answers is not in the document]
chenhunghan: Fixed — down to one sentence: "Clicks the control labelled `label` in a headless `render_stack` and returns what it fired, hovering first so hover-only affordances exist and fanning the pile out when `expanded`." (commit 542672e)

docs/credentials.md
```diff
@@ -55,7 +55,11 @@ Decisions are made interactively, at either of two moments:
 
 Either way you choose to use the value Lens Sandbox detects on the host, store a
 specific value at the boundary, or deny (requests carrying the placeholder then
-fail at the boundary). The decision is remembered for next time.
+fail at the boundary). A bound value is remembered for next time. A deny from an
+approval card is not — it fails that request, and the next use asks again.
```
jansav: **The doc change states the opposite of what the code does, and contradicts the next paragraph.**

"A deny from an approval card is not [remembered] — it fails that request, and the next use asks again." But the code does `remember_grant(&credential_id, GrantVerdict::Deny)`, and `workload_denied()` then suppresses the card for the rest of this run *and future runs*. Two paragraphs down, the same file says the opposite: "Declining a **reactive** first-use card … is remembered as a per-workload deny instead."

**Why:** This is the branch's original once-only deny model, which the PR description says was dropped in favour of main's sticky grant. The sentence about closing a card is correct and worth keeping — the sentence about deny is a rebase leftover that now tells developers a security decision isn't remembered when it is.

crates/lns-cli/tests/behaviours/steps/connector_cli.rs
```diff
@@ -407,3 +407,21 @@ fn listed_as_kind(world: &mut BehaviourWorld, id: String, kind: String) {
         "expected {id} listed as {kind}, got line: {line}"
     );
 }
+
+#[given("the approval-window value card will be dismissed")]
+fn value_card_dismissed(world: &mut BehaviourWorld) {
```
jansav: **This scenario asserts its own fixture string, not the dismissal.**

The Given hardcodes `"the value card was dismissed before a decision was made"` and the Then asserts the output contains `"dismissed"`. Nothing links either end to `resolve_bind_decision(Dismiss)`, which is where that wording actually lives.

**Why:** Change the reason in `bind.rs` to anything without the word "dismissed" and this stays green while the scenario still claims to pin "Dismissing the value card reports the dismissal". `bind.rs` already unit-tests the exact reason, and `connector/mod.rs:1132` already pins that a `BindOutcome::Failed` bails non-zero — so as written the scenario is decorative. Either drop it, or have the fake derive its reason from `resolve_bind_decision`.

crates/lns-service/tests/behaviours/steps/credential_flow.rs
```diff
@@ -910,3 +926,143 @@ fn decision_allow_assert(frames: &[HostFrame]) -> Result<(), String> {
...
+        })
+        .count();
+    if failed < 2 {
+        return Err(format!("expected two failed requests, got {failed}"));
+    }
+    Ok(())
+}
+
+#[then("a future request carrying either placeholder fires a fresh credential card")]
+fn then_either_placeholder_fires_fresh_card(world: &mut BehaviourWorld) -> Result<(), String> {
```
jansav: **The close-all scenario names two providers and checks one.**

`then_either_placeholder_fires_fresh_card` submits only `FIXTURE_ID`, but the scenario set up `"some-provider"` **and** `"some-other-provider"`.

**Why:** The bug being fixed was specifically that *one* click denied *every* credential in the stack. A step that checks one of the two providers can't distinguish "close-all decided nothing for any of them" from "close-all decided nothing for the first one" — which is the whole claim of the scenario.

jansav (commented):
**Problem:** Three paths recorded a verdict the developer never picked — closing a credential card wrote a permanent machine-wide `{"kind":"deny"}` to `~/.lns-credentials.json` (unrecoverable except by hand-editing JSON), closing a network card logged a `deny_once`, and any unsuccessful sign-in audited as a refusal.

**Approach:** A dismissal becomes its own thing rather than a deny: `CardAction::DismissNetwork`/`DismissCredential` carry only an id, `CredentialDecisionRequest::Dismiss` records no grant/rule/audit line and sends `Timeout` on the wire, and `close_action` returns a closed `Dismissal` type so `apply_dismissal` is exhaustive. `DenyScope` on the prompt makes the run flow structurally unable to widen a refusal past the workload. `egui_kittest` closes the widget-wiring gap that let this ship untested (`tray.rs` is coverage-exempt).

**Product alignment:** Directly serves "policy you run into, not write" — the previous behaviour turned an accidental ✕ into an invisible, permanent, machine-wide policy edit.

Verified independently: core's `gate.rs` treats `Decision::Timeout` as deny with no negative caching, so a dismissed request fails closed and the next one re-asks — the wire claim holds. 1818 unit tests, 169 lns-service and 295 lns-cli Layer 2 scenarios pass locally. The new dev-dep is clean: `egui_kittest` 0.34.2 matches `egui` 0.34.2 and pulls only `kittest`/`accesskit_consumer`, no wgpu.

**Verdict:** Needs fixes. Three findings, all in the "should fix" band — one wrong doc statement and two tests that claim more than they check. Nothing wrong with the approach.

docs/credentials.md
chenhunghan: Fixed — and you're right that it was rebase debris: I updated this prose in the `.feature` header when reconciling onto main's grant sidecar and missed the doc.

The deny sentence is gone. The paragraph now says a bound value *and* a deny are both remembered, and leaves the reach to the paragraph immediately below that already explains it (proactive card → `~/.lns-credentials.json`, reactive card → per-workload grant). The dismissal sentence — the part that was correct — stays, now stating explicitly that nothing is recorded, the held request fails, and the next use asks again.

I also swept the rest of `docs/` for the same leftover; the other "asks again" mentions are the disconnect-while-open case and `connector revoke`, both accurate. (commit cf8ecb0)

crates/lns-cli/tests/behaviours/steps/connector_cli.rs
chenhunghan: Dropped it — your first option.

I looked at deriving the fake's reason from `resolve_bind_decision` and it isn't reachable: `lns-cli` depends on `lns-ipc` and `lns-policy`, not `lns-service`, and the reason crosses as an opaque `Response::CredentialBindFailed { reason }`. So at this boundary a dismissal is *indistinguishable* from a timeout or any other bind failure — which is why the scenario could only ever assert its own fixture. A scenario titled "reports the dismissal" claims a distinction the CLI provably cannot make.

Narrowing it to the pass-through contract wouldn't have earned its place either: that's already pinned at `connector/mod.rs:1121` (failed bind → non-zero, reason surfaced, connector not recorded), and the `Cli`-to-bind-failure wiring is pinned by the sibling "fails clearly when the service is unavailable" scenario. It would have been a third pin of the same path wearing a dismissal label.

Dismissal stays pinned where it's observable: `bind.rs:236` for the wording, and `credential_flow.feature` service-side for the behaviour. Both step definitions went with it, so `no_dead_steps` stays honest. (commit 8d0ae75)

crates/lns-service/tests/behaviours/steps/credential_flow.rs
chenhunghan: Fixed, and chasing it turned up something worse than the asymmetry you spotted.

The step now takes both ids from the scenario and asserts each gets a fresh card. But when I mutation-tested that by making `close_all` handle only the first card in the stack, **the scenario still passed** — because the `When` step looped over the pending cards calling `record_decision(Dismiss)` itself. It never ran the pile's fan-out at all, so it was pinning the loop the test wrote, not the one the ✕ runs. As written it couldn't have caught the original bug even with both providers checked.

It couldn't reach the real thing either: `close_all` was private, and the rig dropped the delivery channel the window queues decisions on (production drains it in `credential_delivery_loop`, also private). So:

- `tray::close_all` is now `pub`, with a one-line note saying why.
- The rig gained `apply_queued_card_decisions`, applying queued deliveries the way the service's delivery loop does — the harness completing wiring it had left dangling, rather than the test deciding anything.
- The `When` step clicks close-all and lets production fan out.

Same mutation now fails at `both held requests are failed at the boundary`. (commit 061027d)

jansav (commented):
Verified locally in an isolated worktree: `cargo test -p lns-service --all-targets` passes — 169/169 Layer 2 scenarios, all unit tests green.

**Problem:** Three paths recorded verdicts nobody picked — closing a credential card wrote a permanent machine-wide `{"kind":"deny"}` that silenced the card in every project forever, closing a network card logged a `deny_once`, and any failed sign-in audited as a refusal.

**Approach:** Make a dismissal unrepresentable as a verdict: `DismissNetwork`/`DismissCredential` carry only an id, a new `Dismiss` request variant records nothing and reads as `Timeout` on the wire, `close_all` routes through the same `close_action` → `apply_dismissal` mapping as the per-card ✕, and a `DenyScope` on the prompt means only the bind card can emit `DenyAlways`. The new in-crate `egui_kittest` harness clicks the real ✕ through the AccessKit tree, so the widget wiring is finally pinned.

**Product alignment:** Squarely on "policy you run into, not write" — a stuck `deny` you could only escape by hand-editing JSON was the opposite. The `DenyScope` split also keeps blast radius where the card that asked put it.

**Verdict:** Needs fixes — two maintenance-level findings, both about the one hop that stayed unpinned.

crates/lns-service/src/tray.rs
```diff
@@ -538,6 +554,34 @@ pub enum CardAction {
     },
 }
 
+/// The refusal a card's "Deny" is entitled to ask for, so durability follows the card that asked rather than the flow that happens to answer.
+fn deny_request(scope: DenyScope) -> CredentialDecisionRequest {
```
jansav, line 558: **`deny_request` is the one untested hop, and it's the only thing that still writes a machine-wide deny.**

`bind.rs` pins that the bind prompt is `DenyScope::Machine`, and that `resolve_bind_decision(DenyAlways)` persists — but nothing pins the mapping between them. `deny_request` lives in `tray.rs`, which is coverage-exempt, so the gate can't see it either.

**Why:** invert the two arms and the bind card's Deny becomes `BindResolution::Failed("a deny scoped to one run cannot bind a standing refusal")` — `lns connector connect <id>` + Deny stops recording the refusal and reports a nonsense reason — with the whole suite still green. That is the same shape as the bug being fixed here (an unpinned tray mapping), and the kittest harness this PR just added makes it a one-liner: `click_labelled_control(machine_scope_card(), "Deny", false)`.

crates/lns-service/Cargo.toml
```diff
@@ -163,6 +163,7 @@ tokio-test = "0.4"
 # `regex` in as a dev-dep is cheap because cucumber already depends on
 # it transitively.
 regex = "1"
+egui_kittest = "0.34"
```
jansav, line 166: **The `tray.rs` coverage exemption now cites a blocker this PR removed.**

The IGNORES entry at `scripts/coverage-floor.sh:90` justifies exempting `tray.rs` with "the egui card renderers' click/interaction state needs a live frame + pointer input (**no headless egui_kittest harness in-crate**)". This PR adds exactly that harness and drives a real click through it, but leaves the entry and its reason untouched.

**Why:** CLAUDE.md makes that table the tracker — "the entry's reason names the design move that gets the file to 100%, and the entry leaving the table is the completion signal." Leaving a falsified reason in place means the next person reads a blocker that no longer exists and keeps a large file exempt on that premise. Rewrite it to the residual blocker (`eframe::App` / `run_native` main-loop wiring), which also makes the previous finding's home explicit.

crates/lns-service/src/tray.rs
chenhunghan: Fixed — and the mutation you described is exactly right: inverting the two arms left the whole suite green while `lns connector connect <id>` + Deny stopped recording the refusal and reported "a deny scoped to one run cannot bind a standing refusal".

Two kittest tests now click the real Deny button through the AccessKit tree, one per scope:

- `a_run_card_denies_only_for_this_workload` — `DenyScope::Workload` fires `Deny`
- `the_bind_card_denies_for_the_whole_machine` — `DenyScope::Machine` fires `DenyAlways`

Inverting the arms now fails both. (commit 83589c9)

crates/lns-service/Cargo.toml
chenhunghan: Fixed — the "no headless egui_kittest harness in-crate" clause is gone, since this branch added one.

The reason now names the residual blocker: `eframe::run_native` owns the main thread (and needs a live NSApp on macOS), so the `eframe::App` impl and tray/menu wiring can only run live, and reaching 100% needs the per-frame logic behind a seam the harness can drive. It also lists what the harness *does* now cover (close ✕, pile pill, Deny scope mapping) and adds `close_action`/`apply_dismissal`/`deny_request` to the host-tested residue — so the previous finding's home is explicit, as you suggested.

Ran the full `make coverage` after the edit: no failures. (commit 5c8806a)

jansav (approved):
Both findings verified as addressed at `5c8806a`.

**`deny_request` untested hop (`83589c9`) — fixed, and the test genuinely bites.** The two new tests click the real Deny button through the kittest/AccessKit harness on a `Workload`- and a `Machine`-scope card and assert the exact `CredentialDecisionRequest` fired. I verified by mutation rather than by reading: inverting the two arms of `deny_request` fails **both** tests (`left: Deny, right: DenyAlways`). The gap is closed by a test that detects the mutation, not one that merely executes the line.

**Falsified IGNORES reason (`5c8806a`) — fixed.** The "no headless egui_kittest harness in-crate" clause is gone, and the entry now names the residual blocker (`eframe::run_native` owns the main thread, so the `App` impl and tray/menu wiring only run live) plus the design move to 100%. Spot-checked the other direction too — `deny_request`, `close_action`, and `apply_dismissal` are all genuinely host-tested, so the reason isn't overclaiming.

**Gate, run locally at `5c8806a`:** `cargo fmt --check` clean, `cargo clippy -p lns-service --all-targets -- -D warnings` clean, `cargo test -p lns-service --all-targets` green — 1820 unit tests (up from 1818) and 169/169 Layer 2 scenarios.

One thing still unverified: `make coverage` has not been run locally. Every new `Dismiss`/`DenyAlways` arm in `session.rs` / `bind.rs` has a test, so I'd expect the per-file gate to pass, but CI is the only real check on that.

Ship it. ✅

PR #174, opened 2026-07-27
feat: declare developer tools in the sandbox definition, provisioned once per machine

### LINKED ISSUE #173: Declare developer tools in the sandbox definition, provisioned once per machine
## Problem

A sandbox author cannot declare the developer tools their workload needs. The base image must smuggle them in — dictating the image choice — or the workload reinstalls them over the network on every cold start (the microVM is ephemeral), paying boot latency and holding open policy routes that exist only for tool acquisition, not for the workload's actual job. The Claude Code example documents the workaround today: a `node:lts-alpine` base chosen solely for npm, a first-boot `npm install` on every cold start, and `registry.npmjs.org` held open just for it.

## Target crates

Multi-crate: `lns-artifact` (definition schema), `lns-cli` (authoring, validation, disclosure), `lns-service` (provisioning, cache, injection, audit), plus `docs/`.

## Approach

> Mechanisms marked **[validated]** were exercised end to end on 2026-07-27 with mise v2026.7.14 in real microVM guests (Alpine 3.20 musl and `debian:bookworm-slim` glibc, aarch64), driven through `lns run` itself. Full install+execute matrix:
>
> | Tool | glibc guest | musl guest |
> |---|---|---|
> | `python@3.12` | ✅ | ✅ |
> | `node@22` | ✅ | ✅ with pinned `libstdc++`+`bash` companions |
> | `go@1.24` | ✅ | ✅ |
> | `rust@1.88` | ✅ (needs bundled static curl) | — |
> | `java@21` | ✅ | — |
> | `terraform@1.9` | ✅ | — |
> | `jq@latest` / `ripgrep@latest` | ✅ | ✅ |
> | `gh@latest` | ✅ | — |
> | `uv@latest` | ✅ | ✅ |
> | `deno@2` | ✅ | ❌ no musl builds upstream |
> | `bun@1` | ✅ | ✅ with companions |
> | `@anthropic-ai/claude-code` | ✅ via declared node + `npm install -g` | — |

**Spec surface.** Add `spec.tools` to `lns.yaml` — a list of `name@version` entries (`node@22`, `python@3.12`, `go@1.24`, and anything else the engine's registry covers). A version is required — `@latest` is accepted and resolves at provision time — and engine backend prefixes (`aqua:`, `ubi:`, `npm:`) are refused: the spec carries portable tool names only, never engine syntax. Offline validation checks shape; resolution happens at run time.

**Install engine.** Tools are provisioned by **mise acting as a pinned install engine** — never linked as a library, never run as part of the workload. mise's static musl build runs dependency-free on both musl and glibc guests **[validated]** and honors the proxy environment **[validated]**.

**Engine versioning.** The mise binary is pinned in a checked-in manifest (version + per-arch URL + sha256, a `mise.toml` beside `kernels.toml`), fetched lazily into the service cache with a `.ready` sentinel — the `guest_tools` pattern. Bumps happen through operator tooling in the `bump-kernel` mold, verifying upstream's published checksums and minisign signature before rewriting the pin; the bump PR runs the tool-provisioning e2e scenarios, because registry semantics can drift between mise versions. Users never see or choose the engine version; mise's self-update and version phone-home are disabled (it phones home even on `--version` **[validated]**).

**Timing.** Fetching happens at launch planning on the host — after `lns run` is invoked, after image ingest (which determines the libc flavor), and strictly **before the workload microVM is created**; a provisioning failure refuses the launch before anything boots. `lns pull` additionally pre-provisions a published sandbox's (push-pinned) tools, preserving the documented promise that a pulled sandbox can start offline from the cached snapshot. `lns push` resolves fuzzy versions at publish time but downloads no tools.

**Cold path (first provision of a tool set on a machine).** The service boots a short-lived disposable provisioner guest containing the pinned mise, a pinned CA bundle installed at `/etc/ssl/certs/ca-certificates.crt` (slim images ship no CA store; mise ignores `SSL_CERT_FILE` but reads the canonical path **[validated]**), a pinned static curl (mise's rust backend bootstraps via rustup, which hard-requires curl or wget — busybox wget is not enough **[validated]**), and a writable staging share. It runs `mise install` for the declared set with fail-loud settings — notably `MISE_PYTHON_COMPILE=0` and `node.compile=false`: on any fetch failure mise otherwise *silently falls back to compiling from source*, which cannot work on a bare guest and buries the real error (this fallback impersonated a musl-support bug during our own validation until source review found the knob) **[validated]**. python and go installs are download → checksum → extract with no execution of the installed tool **[validated]**; node, bun, and deno execute the tool as a post-install check (node runs both `node` and `npm -v`) **[validated]**, so the provisioner rootfs comes in both libc flavors and always matches the flavor being provisioned — a cheap constraint since tool trees are already keyed by libc. The musl flavor bundles pinned companions (`libstdc++` for node/bun, `bash` for npm's wrapper script) **[validated: node and bun install and execute on Alpine with exactly these companions]**; the same companions ship alongside the tool tree into musl workload guests. python-build-standalone downloads are additionally sigstore-attestation-verified by mise **[validated]** — the provisioner's egress set must include `tuf-repo-cdn.sigstore.dev` and the `*.jdx.dev` version index alongside the tool backends' hosts.

**Ingest and warm path.** The staging tree is ingested into the service's content-addressed store and injected into workload guests as a **read-only layer** under the reserved runtime namespace, with the tools exposed via provision-time-resolved bin paths prepended to the workload's `PATH` (prepended at final env composition, since the image's own `PATH` wins otherwise). The workload guest never contains the engine — mise exists only in the provisioner. Warm runs — including restarts and rebuilt microVMs — execute no mise and touch no network: fetching is never part of the boot path, and tool file bodies aren't even copied at boot (the guest reads them lazily through the read-only content share). Each tool tree is cached by *resolved version + arch + libc* and tool sets are composed at injection time, so adding one tool to a project never re-fetches the others, and the content-addressed store dedupes identical trees across projects. The cache key deliberately excludes the engine version, so a mise bump never invalidates provisioned tools or triggers re-downloads; mise only runs again on the next cold provision.

**libc matching.** Installed builds must match the *workload* image's libc, not the provisioner's: musl image → musl builds, glibc image → gnu builds. Validated both ways: `python@3.12` (musl pbs build) installs and executes on Alpine, ~20 s including boot; `node@22` (official gnu binary) installs and executes with npm on Debian, ~6 s including boot **[validated]**. node was subsequently validated on musl as well (with the companions above), so both flavors are proven end to end.

**Trust and audit.** Acquisition is **audited system egress** — the same trust shape as pulling `spec.image`: declaration is consent, no approval card, no policy route. Provisioning events land in the per-run audit chain of the run that triggered them (the mounts precedent). The workload can neither tamper with the cached tools (read-only layer) nor reach the network to acquire them; everything a tool *does* at runtime stays inside the normal policy cage.

**Reproducibility and distribution.** First provision records the resolved exact versions in the machine cache (plus the provisioning mise version as debugging metadata, never as a constraint); later runs on that machine reuse them, so `node@22` never drifts between runs even when upstream releases a newer 22.x. There is deliberately no project-level lock file; cross-machine reproducibility comes from explicit versions in the spec and from publishing: `lns push` resolves fuzzy entries (`@22`, `@latest`) and embeds the exact pins in the published artifact — consumers run exactly what the author tested, disclosed via `lns inspect` and the run summary. Composition with the workload's own package manager works today: `npm install -g @anthropic-ai/claude-code` on a mise-provisioned node prints `2.1.220 (Claude Code)` **[validated]** — which is what lets the Claude Code example drop its node base-image coupling.

**What does NOT change.** The policy model, credentials/connectors, filesets/volumes, and the required `spec.image` are untouched; a definition without `tools` behaves byte-identically to today.

## Alternatives considered

- **Curated tool packs published by us as OCI artifacts** — rejected: we'd own a per-tool, per-arch, per-libc packaging pipeline forever; loses mise's registry breadth, which is the point.
- **Install at first boot inside the workload** (today's documented pattern) — rejected: per-run latency, policy routes held open for acquisition, and persisting installs would let a compromised workload poison future runs.
- **mise as a Rust library** — rejected: it isn't one; binary-only distribution, no stable embedding API.
- **Running mise on the host to fetch for the guest** — rejected: mise can't reliably install for a different OS/arch/libc than the machine it runs on.
- **Author-pinnable engine version in `lns.yaml`** — rejected: leaks the engine into the spec surface and splinters the test matrix; the recorded tool versions already carry all the reproducibility that matters.

## Decision record

Decisions made during intake, with the reasoning — so a reader doesn't have to reconstruct the discussion.

1. **Scope is any developer tool, not just language runtimes** (hence the field name `tools`, not `runtimes`): the same mechanism covers `terraform` or `rust@1.88` at zero marginal cost, and "runtimes" would under-sell it.
2. **No approval card for declared tools — declaration is consent.** `spec.image: debian` already triggers a policy-exempt system fetch with no card; a card for tools but not images would be incoherent. The one threat a card would cover (a workload editing its own bind-mounted `lns.yaml` so the *next* run fetches something the user never typed) equally rewrites `spec.policy`, `spec.volumes`, and `spec.env` — tools are the least dangerous line in a tampered spec. If spec-tamper resistance is wanted, it's one cross-cutting "spec changed since last run" mechanism, decided separately for the whole spec — explicitly out of scope here.
3. **Acquisition is system egress, audited, outside workload policy** — the docs define policy as governing what the *workload* may reach; provisioning runs no untrusted code (pinned engine + user-declared list) and carries no secrets or project mounts. Workload package-manager traffic (`npm i`, `pip install`) remains fully policy-caged; the network cage applies to all guest processes, so nothing in the workload gains a bypass.
4. **Resolved versions are recorded per machine (service cache), not as a project file.** A committed lock file next to `lns-policy.yaml` was considered and rejected to avoid new project-directory surface. Accepted consequence: a fuzzy version (`node@22`, `node@latest`) may resolve differently on different machines until published — cross-machine reproducibility comes from explicit spec versions and from `lns push`, which resolves fuzzy entries itself rather than refusing (push already has network) and embeds exact pins, mirroring how `path:` filesets are digest-pinned at push.
5. **Engine version is ours, pinned per release, off the hot path**: not in the tool-set cache key (a mise bump must not re-provision the world), recorded in the project versions file as metadata only, bumped via signed-and-checksummed operator tooling with e2e gating.
6. **libc matching is mandatory** (musl→musl, glibc→gnu). Without it, gnu binaries fail with a bewildering `not found` on musl/imageless-style rootfs, and always-musl degrades python (static musl python can't load native extensions — no numpy on Debian images).
7. **Provisioner runs in a disposable guest, results injected read-only** — a compromised workload can't poison the tool cache for future runs, and ephemeral-by-default is preserved (nothing tool-related persists in the workload's writable layer).
8. **Supported backends are the Rust-native ones** (core languages, aqua, ubi). Bash-based asdf plugins can't run on a bare provisioner guest and are refused by name.
9. **python2 is out** — no maintained standalone builds; mise's only path is source compilation (pyenv's python-build), which needs a C toolchain and bash we don't ship. Escape hatch: bring it via `spec.image`.
10. **Published sandboxes are first-class**: a pulled sandbox provisions its (push-pinned) tools on the consumer's machine, disclosed before anything runs — the definition stays portable.
11. **The workload guest is engine-free.** Tools are exposed through provision-time-resolved bin paths on `PATH`, not mise shims. Shims (mise injected into the workload for `mise ls` / `mise x` introspection) were considered and rejected: smaller guest surface wins, and the engine staying provisioner-only also means an engine-version bump touches nothing in workload guests.
12. **Entry syntax: version required, `@latest` allowed, backend prefixes refused.** A bare `node` is a validation error — the author states an intent (`node@22` or explicitly `node@latest`); `aqua:`/`ubi:`/`npm:` strings are engine internals and never appear in a portable artifact.
13. **Provisioning audit events land in the per-run chain of the triggering run** (the mounts precedent), not the durable connection ledger — `lns audit <run>` explains why that run touched the network.

## Upstream findings (re-validated against source, 2026-07-27 — neither is a clean "mise bug")

- **Source-compile fallback** — *confirmed in mise v2026.7.14 source, with nuance.* `node.rs::install_precompiled`: when the precompiled archive fetch returns **404** (not "any fetch failure" as this section originally said — other errors propagate) and `node.compile != false`, mise falls back to compiling from source; the only surface is a `warn!` when a flavor mismatch is detectable, otherwise the fallback is announced at `debug!` only. It is by-design behavior with documented off-switches (`MISE_NODE_COMPILE=0` / `MISE_PYTHON_COMPILE=0` — env names verified in settings.toml; both pinned in our provisioner). A UX report ("announce the fallback at warn/info by default") is defensible; a defect report is not.
- **npm bin-shim `.exe`** — *reattributed.* The shim generator (aube, github.com/jdx/aube, embedded as a crate) links `package.json` `bin` entries verbatim, and `@anthropic-ai/claude-code`'s registry manifest literally declares `bin: {"claude": "bin/claude.exe"}` as its **only** entry (verified against registry.npmjs.org, v2.1.220), relying on an install-time lifecycle script to swap in the platform binary from its `optionalDependencies`. So nothing "selects the Windows entry" — the failure is that the platform-swap postinstall doesn't take effect under mise's embedded-aube install. Root cause could sit in aube's lifecycle-script policy, mise's npm backend, or claude-code's packaging pattern; needs a runtime repro before filing anywhere. npm-backend tools stay out of the supported set either way.

Incidental product finding from the same validation, separate-issue candidate: the supervisor's boot-time proxy-CA install logs `failed to open /etc/ssl/certs/ca-certificates.crt for appending: No such file or directory` on CA-less images like `debian:bookworm-slim`, leaving the guest without the proxy CA.

## Code pointers

- `crates/lns-artifact/src/sandbox.rs` — `SandboxSpec` (`deny_unknown_fields`; new field + validation)
- `crates/lns-service/src/runtime_layer/mod.rs` — `for_run` is the existing read-only-layer injection seam
- `crates/lns-service/src/guest_tools/mod.rs` — the pinned-fetch pattern to copy for the mise binary
- `crates/lns-service/kernels.toml` + `crates/bump-kernel` — the pin-manifest + bump-tooling pattern to copy for the engine version
- `crates/lns-service/src/workload_env.rs` — final `PATH` composition (the image's own `PATH` wins via putenv; prepend here)
- `crates/lns-cli/src/sandbox/author.rs` — `SCAFFOLD` for `lns init`

## Acceptance criteria

The following Gherkin scenarios describe the expected behavior. If all scenarios pass, the feature is done.

```gherkin
Feature: Declared developer tools

  Scenario: Scaffold carries the tools field
    When I run "lns init"
    Then the generated lns.yaml contains "tools: []" with every other spec field unchanged

  Scenario: Declaring tools validates offline
    Given a lns.yaml declaring tools ["node@22", "python@3.12"]
    When I run "lns sandbox validate"
    Then validation succeeds without touching the network or the service

  Scenario: A malformed tool entry is refused with its cause
    Given a lns.yaml declaring tools ["node@"]
    When I run "lns sandbox validate"
    Then validation fails naming the entry and the expected "name@version" shape

  Scenario: A tool entry without a version is refused
    Given a lns.yaml declaring tools ["node"]
    When I run "lns sandbox validate"
    Then validation fails asking for an explicit version such as "node@22" or "node@latest"

  Scenario: A declared tool is available to the workload
    Given a lns.yaml declaring tools ["node@22"] over a base image that ships no node
    When the sandbox runs "node --version"
    Then it prints a node 22 version

  Scenario: A declared tool wins over the base image's copy
    Given a base image that ships node 20 and a lns.yaml declaring tools ["node@22"]
    When the sandbox runs "node --version"
    Then it prints a node 22 version

  Scenario: Tool acquisition needs no policy route and raises no approval
    Given a lns.yaml declaring tools ["node@22"] with defaultVerdict ask and no allowedRoutes
    When I run the sandbox for the first time
    Then the tools are provisioned without any approval card
    And the workload's own network requests still ask as usual

  Scenario: A provisioned tool set is reused without network
    Given tools ["node@22"] were provisioned by an earlier run on this machine
    When I run the sandbox again
    Then the run starts without downloading anything

  Scenario: Failed provisioning refuses the launch cleanly
    Given a lns.yaml declaring a tool whose download cannot complete
    When I run the sandbox
    Then the launch is refused naming the tool and the cause
    And a later run retries from a clean state

  Scenario: A tool unknown to the registry refuses the launch
    Given a lns.yaml declaring tools ["definitely-not-a-tool@1"]
    When I run the sandbox
    Then the launch is refused naming the unknown tool

  Scenario: First provision records resolved versions on this machine
    Given a lns.yaml declaring tools ["node@22"]
    When I run the sandbox for the first time
    Then the resolved exact version is recorded on this machine
    And later runs here use the recorded version even after upstream releases a newer 22.x

  Scenario: Publishing pins resolved tool versions
    Given a lns.yaml declaring tools ["node@22"]
    When I run "lns push ghcr.io/acme/agent:1.0.0"
    Then the published artifact carries the exact resolved versions

  Scenario: A pulled sandbox with tools starts offline
    Given a published sandbox declaring pinned tools
    And I ran "lns pull" on its reference while online
    When I run the sandbox with no network available
    Then it starts and the declared tools are available to the workload

  Scenario: A pulled sandbox's tools are disclosed before anything runs
    Given a published sandbox declaring tools
    When I run "lns inspect" on its reference
    Then each declared tool and its pinned version is listed
    And the run summary discloses them at launch

  Scenario: Tool provisioning is audited
    When tools are provisioned for a run
    Then the audit chain records what was fetched, from where, and its resolved version
```

Invariants that must not change:

- A definition with no `tools` produces byte-identical behavior to today.
- `spec.image` remains required; tools stack on top of it, never replace it.
- Everything a tool does at runtime (npm/pip/go traffic) stays inside the normal policy cage.
- Tools land read-only; changes the workload makes elsewhere still die with the microVM.

## Non-goals

- python2 (decision record #9).
- Project dependencies (the workload's own `npm install` / pip requirements — policy-gated as ever; volumes persist them).
- Imageless operation (per docs, "there is no imageless mode").
- Bash-based (asdf) backends (decision record #8).
- npm-backend tools until the aube bin-shim bug is fixed. deno on musl base images (upstream ships no musl builds). Any such refusal must be actionable: name the tool, the limitation, and both remedies — switch to a glibc base image, or bring the runtime via `spec.image` as today.
- Any change to the policy model or approval flow.
- Spec-tamper resistance (decision record #2) — a separate, whole-spec concern.

## Implementation checklist

Sequenced by dependency; check off as landed. The Gherkin scenarios above are the definition of done — this list is the route there.

**Spec surface (`lns-artifact`, `lns-cli`)**
- [x] `spec.tools` field on `SandboxSpec` + shape validation (version required, `@latest` allowed, backend prefixes and bare names refused with actionable messages)
- [x] `lns init` scaffold gains `tools: []`; `lns sandbox validate` covers the new refusals offline
- [x] `lns inspect` and the run summary disclose declared tools (and pinned versions when resolved)

**Engine distribution (`lns-service`)**
- [x] `mise.toml` pin manifest (version + per-arch URL + sha256) and lazy cached fetch with `.ready` sentinel (guest-tools pattern)
- [x] Companion pins: CA bundle, static curl, musl `libstdc++` + `bash` apks
- [x] `bump-mise` operator tooling (checksum + minisign verification, e2e-gated bump PRs)

**Provisioner (`lns-service`)**
- [x] Disposable provisioner guest, one rootfs per libc flavor, matching the flavor being provisioned; CA at `/etc/ssl/certs/ca-certificates.crt`, curl on PATH, companions present
- [x] Fail-loud mise config (`MISE_PYTHON_COMPILE=0`, `node.compile=false`, self-update/phone-home off) + provisioner egress set (`*.jdx.dev`, `tuf-repo-cdn.sigstore.dev`, backend hosts)
- [x] libc flavor detection from the ingested image (with musl↔glibc keying)
- [x] Staging-share ingest into the content store; per-tool cache keyed by resolved version + arch + libc; machine-level resolved-versions record (+ engine version as metadata)
- [x] Launch refusal paths: unknown tool, unsupported backend, deno-on-musl, failed download — each named, actionable, clean-state retry

**Workload integration (`lns-service`)**
- [x] Tool trees injected read-only via the runtime layer; resolved bin paths prepended in `compose_workload_env`; musl companions ride along; engine never enters the workload guest
- [x] Timing integration: provision at launch planning (post-ingest, pre-boot); `lns pull` pre-provisions pinned tools; `lns push` resolves fuzzy versions and embeds pins
- [x] Audit events in the triggering run's chain (what was fetched, from where, resolved version)

**Tests & docs**
- [x] Layer 2 features for validation/disclosure/refusals; Layer 3 for cache keying and record handling; `@microvm` e2e for install+execute on both libc flavors (live Vz pass: alpine/musl + bookworm-slim/gnu, 4/4 scenarios)
- [x] Docs: spec-field table row, both scaffold listings, tools section in running-workloads (incl. "acquisition is system fetch, not workload egress"), audit.md event note
- [x] Rewrite the Claude Code example: glibc slim base + `tools: [node@22]`, drop the npm route once verified

**Upstream / side issues**
- [x] ~~File mise issue: aube npm bin-shim selects Windows `.exe` entry~~ — re-validation reattributed this (see Upstream findings above); nothing to file as described
- [x] ~~File mise issue: silent source-compile fallback buries fetch errors~~ — re-validation softened this to a UX suggestion at most (see Upstream findings above); not filing as a bug
- [ ] Decide/file: supervisor proxy-CA append fails on CA-less images

## Open questions

- File the upstream mise issues (aube bin-shim `.exe` selection; silent source-compile fallback UX) and track them; decide whether the supervisor proxy-CA finding gets its own issue.

chenhunghan:
Closes #173.

A sandbox author can now declare the developer tools their workload needs — `spec.tools: [node@22, python@3.12]` — instead of smuggling them in through the base image or reinstalling them over the network on every cold start. Tools are provisioned once per machine by a pinned mise engine running in a disposable provisioner guest, ingested into the content-addressed store, injected read-only at `/.lens/tools/`, and prepended to the workload `PATH`. Acquisition is audited system egress with the same trust shape as pulling `spec.image`: declaration is consent — no approval card, no policy route — while everything a tool *does* at runtime stays inside the normal policy cage.

## What's in here

- **Schema + validation** (`lns-artifact`): `spec.tools` as portable `name@version` entries; a version is required (`@latest` allowed), engine backend prefixes and bare names are refused with actionable messages; shared `tools::ToolRef` parser.
- **Author surface** (`lns-cli`): `lns init` scaffolds `tools: []`; `lns sandbox validate` covers the refusals offline; `lns inspect` (offline, cached-ref) and the run summary disclose declared tools.
- **Publish-time pinning** (`lns-cli`): `lns push` resolves fuzzy versions against mise's public version index (new `ToolResolver` port) and embeds exact pins — the tool analogue of fileset digest-pinning — printing `pinned node@22 → node@22.11.0` per entry so the publisher sees what they shipped without reading it back out of the registry. `--dry-run` stays offline and notes when the published digest may differ.
- **Per-tool engine state** (`lns-service`): each declared tool installs against its own `MISE_DATA_DIR`/`MISE_CACHE_DIR`. Sharing one data dir let a tool's install code (arbitrary upstream code, root in the provisioner guest) plant `installs/<sibling>/<version>/` so the engine reported a *genuine* install of a tree it wrote — which the host then cached machine-wide under the sibling's provenance, escaping the project that declared the bad tool. The duplicate-marker check doesn't catch that, because nothing is forged.
- **Author-time tool gate** (`lns-artifact`): the registry snapshot moved from `lns-service` into the shared crate, so `lns sandbox validate` and `lns push` refuse an unprovisionable tool (unknown, or plugin-backed like `npm:prettier`) offline with the same wording the launch uses. Previously such a tool resolved, pinned, and published cleanly, and the author found out from a consumer's failed launch. The gate is deliberately **not** in `sandbox::parse`: the shipped registry shrinks and re-backends entries over time, so a consumer must still be able to `inspect` and `pull` an artifact whose tool this build can no longer install — the launch keeps its own refusal for that. The libc gate stays service-side, since it needs the base image's layers.
- **Bounded index query** (`lns-service`, `lns-cli`): the `@latest` lookup is wrapped in a 5s budget so the documented "falls back to the last version resolved here" holds against a blackholing proxy or captive portal, not just connection-refused; `lns push`'s resolver gained a request timeout and a body cap for the same reason. A resolved version is now a validated newtype (`SafeVersion`) rather than a `String`: it is checked at every boundary it enters the system from — the version index, the driver's output, `resolved.json`, and `manifest.json` — because each one becomes a host path component, a guest path component, and a word in the provisioner's shell driver. Validating at deserialize means a tampered cache file reads as unparseable, which the existing "ignore and re-resolve" path already handles; a future source cannot skip the check.
- **Hardlinks in tool trees** (`lns-service`): a type-1 tar entry (GNU and busybox tar both emit them for `nlink > 1`; JDK layouts ship them) resolves to the digest already ingested for its target instead of aborting the provision as an infrastructure fault.
- **Schema versions are enforced, not just stamped** (`lns-service`): a `resolved.json` written under a different `RECORD_SCHEMA_VERSION` is discarded and every tool re-resolved (rather than being read under today's semantics and re-stamped), and a cached tree whose `manifest.json` carries a different `MANIFEST_SCHEMA_VERSION` is a cache miss. Bumping either constant is now a real migration signal.
- **Warm runs don't queue** (`lns-service`): the read-only pass (record + cache lookup, `@latest` index check) runs outside the provision lock; only a set with something to install or record takes it. A fully cached launch no longer waits out an unrelated cold provision — which could be the full 600s driver timeout — while two runs installing the same cold set still serialize, since the lock is taken before a second planning pass that sees whatever the first one cached.
- **`@latest` means latest** (`lns-service`): a bounded line (`node@22`) resolves once per machine and stays pinned — that is what makes a warm run reproducible and offline-capable — but `node@latest` re-asks the version index on every run, host-side, with no guest boot. An unchanged answer is a cache hit; only a genuinely new release provisions. When the index is unreachable the record serves as last-known-good so offline start still works. `resolve_from_index` moved to `lns-artifact` so the service and `lns push` share one resolver.
- **Provisioning** (`lns-service`): registry snapshot (994 entries generated from the pinned mise release, incl. aliases) gates unknown/plugin-backed tools pre-boot; libc flavor detected from the image's layer tars (musl ↔ gnu keying, deno-on-musl refused with both remedies); disposable provisioner guest (pinned rootfs per flavor, CA at the canonical path, static curl, fail-loud engine env) stages tool trees through a writable virtiofs bind; per-tool cache keyed by resolved version + arch + libc with `.ready` sentinels; machine-level `resolved.json` pins first resolutions forever (engine version is metadata, never a cache key); failed provisioning refuses the launch naming the tool and cause, with structural clean-state retry.
- **Workload integration**: tool specs join the runtime layer (descriptor cache keys itself by tool set automatically), resolved bin paths prepend the composed `PATH` ahead of the image's copies, musl companion trees ride along, and the engine never enters the workload guest. Provisioning events land in the triggering run's audit chain (`lns-ocsf` `tool_provision` event).
- **Pull integration**: `lns pull` pre-provisions a published sandbox's pinned tools so the documented offline-start promise holds, and records each acquisition on the machine-level audit chain — a pull has no run chain to write to, so without this the most security-relevant path (running someone else's published sandbox) would be the one with no durable record.
- **Operator tooling**: `bump-mise` verifies upstream's `SHASUMS256.txt` + minisign signature before rewriting the engine pin and regenerates the registry snapshot; runbook at `runbooks/mise-bump.md`.
- **Docs**: spec-table row + new Tools section in running-workloads, scaffold listings, push/pull/inspect reference rows, audit note, and the Claude Code example rewritten onto `debian:bookworm-slim` + `tools: [node@22]` (its npm route remains only for the first-boot install of Claude Code itself — npm-backend tools stay a non-goal until the upstream aube shim bug is fixed).

## Test coverage

All 15 Gherkin scenarios from the issue are on disk: 12 pass in the Layer-2 suites (lns-cli 267 scenarios, lns-service 166 scenarios, green), and the 4 real-guest scenarios in `crates/e2e-tests/features/microvm_tools.feature` (`@microvm`) pass against live Vz guests — cold provision + audit + warm reuse (clean cache home), PATH-wins over the image's node 20 (`bookworm-slim`, gnu), and offline start of a pulled digest-pinned sandbox (alpine, musl). Layer-3 suites cover the parser, cache/record/registry/libc/driver internals, PATH composition, and push pinning.

## Deliberately out / follow-ups

- The `@microvm` tools scenarios passed on a live Vz host (4/4). The live pass surfaced and fixed three real-guest gaps: apk companions are concatenated gzip streams (MultiGzDecoder + ignore_zeros), rootfs pins must be per-arch platform-manifest digests (pulls verify the platform manifest, not the index), and mise's version phone-home is silenced by `CI=1` with all mise paths pinned under `/tmp`. A real end-to-end run of the rewritten Claude Code example remains open; `LNS_TOOL_INDEX_URL` should be confirmed against the live index in that run.
- Pre-existing (not from this PR): `microvm_declarative_launch`'s offline scenario fails on tag re-resolution — a tag reference always re-resolves by design (`manifest_cache.rs`), so the offline-start promise only holds for digest-pinned references. Needs its own issue/decision.
- Tool trees still accumulate: `ContentStore::sweep` has no production caller, so there is no keep-set to join and nothing reclaims a superseded tree's blobs. `@latest` re-resolving per run amplifies this. Deliberately not papered over with a keep-set API that nothing consumes — the sweep caller is the fix. Formerly: `tools::cache` exposes no GC keep-set yet: `ContentStore::sweep` has no production caller today, so there is no keep-set to join; wire tool digests in when a sweep caller lands.
- **Live-guest verified.** `make e2e-microvm` on a macOS Vz host: all 4 `@microvm` tools scenarios pass, including the two strengthened in review — warm reuse now asserts the second run's chain carries no `tool` event (the old stderr assertion could not fail, since progress never reaches a piped stderr), and offline start now runs from a clean cache home and asserts the run fetched nothing (deleting `pre_provision_for_pull` fails it). That covers the per-tool engine-state split and the companion filter against a real guest. Three scenarios outside this feature failed and are not reachable from these changes: two in `microvm_credentials` (one asserts a declared connector's placeholder env is *absent*, which contradicts the shipped seed-but-don't-arm model; the other has no clean-home Given and depends on the developer's own connector state) and `microvm_declarative_launch`'s offline 503, the pre-existing tag re-resolution gap below.
- ~~**Needs a live-guest re-run before merge:**~~ the per-tool engine-state fix changes the rendered driver script, and Layer 2/3 can only pin the script text — real-guest behavior there is covered solely by the `@microvm` suite. `make e2e-microvm` should be re-run on a virt-capable host. The theoretical risk it would catch: a tool whose install expects another tool present in the same data dir (low — every supported backend is download-only; `vfox:`/`asdf:` are refused).
- **Accepted, with the claim corrected rather than the gap closed:** every declared tool installs in one guest as root, so a tool that *deliberately* writes a sibling's engine state still can. Per-tool `MISE_DATA_DIR`/`MISE_CACHE_DIR` stops an install from accidentally resolving a version a sibling left behind; the duplicate-marker fault and the symlink-target refusal bound what a hostile tool can do to a sibling's cached tree. The docs and code comments now say a declared tool is trusted by the tools declared alongside it. Full isolation means one provisioner guest per tool — a boot each — and was judged not worth the cold-provision cost.
- **Residual, documented:** a VMM wedged *before* publishing a connector still cannot be stopped — the boot-timeout path gives a late connector a bounded window and uses it when it arrives, but `request_stop` needs the guest's own transport, so the error says the guest may still be running. Closing it needs a stop handle at the `VmmBackend` level.
- **Backward compatibility is explicitly out of scope for now** (product decision): `spec.tools` under `deny_unknown_fields` means an artifact pushed from this build fails to deserialize on an older lns-service. `apiVersion` is the field that would carry a spec floor if that changes — it is already parsed and validated, so an older service would report `unexpected apiVersion` rather than a serde field error.
- Hardening follow-up: the provisioner guest currently has the unsupervised NAT egress every unsupervised guest has; tightening it to the known egress set (`*.jdx.dev`, `tuf-repo-cdn.sigstore.dev`, backend hosts) via a synthetic policy is future work. The pinned host set is documented in `mise.toml` rather than a Rust allowlist, since there is no supervisor in that guest to enforce one.
- Remaining isolation gap: a compromised tool's install code runs as root in the provisioner guest alongside its siblings (node/bun/deno execute the installed tool as a post-install check), sharing one `/staging` bind. Duplicate `LNS_TOOL` markers are now a hard fault, which closes marker forgery — a tool can add lines but cannot suppress a sibling's genuine marker, so any forgery attempt trips the check. A tool that leaves a background process could still overwrite a sibling's staged tar; fully closing that needs one provisioner guest per tool.
- Review fixes folded in: the libc flavor is now memoized by the image's layer-digest set (a warm tooled run no longer re-gunzips every layer body) and `scan_layer` stops at the musl loader; `MISE_PARANOID=0` dropped as a no-op that read as selecting the weaker value of a security setting (mise defaults it to false — enabling it is possible hardening, but needs a live-guest run to validate); `StagedTool::requested` dropped as write-only.
- Review fixes folded in: the guest capture now returns both streams separately (`CapturedStreams`) instead of one merged buffer. The guest writes stdout and stderr down independent fds at arbitrary chunk boundaries, so a merged buffer splices a partial stderr line (a progress renderer's `\r`, say) in front of the next stdout line — which would make a successful install read as "the driver reported no result" and downgrade a real `LNS_FAIL` to a generic engine fault. Markers are now parsed from stdout alone, the failure cause comes from stderr (falling back to stdout for a bare `sh` failure that never reaches the engine), and the `lns stats` probe reads `.stdout` so diagnostics can't corrupt its `/proc/stat` payload. Also dropped three items with no production callers (`baseline_egress_hosts`, `ToolRef::is_latest`, `EnsuredTools.record`); the record assertions now read the persisted store, which is the property that actually matters.
- The two suspected upstream mise issues were re-validated and downgraded (see #173 "Upstream findings"): the `.exe` shim traces to claude-code's own manifest, and the source-compile fallback is 404-only by design with an off-switch this PR pins — nothing filed.

- Review fix folded in: `is_exact_version` now counts numeric release components instead of requiring digit-led dot parts, so the vendor pins the resolver itself emits (`temurin-21.0.5+11.0.LTS`) short-circuit the index on re-push and stop the false dry-run "may differ" note; the run path now tries every bounded request as its own cache key (a fuzzy one simply misses), so no version shape needs `resolved.json` to find its cached tree.

- Follow-up to the vendor-pin fix: push now verifies an already-exact pin against the index on a best-effort basis — a listed pin publishes silently, an unlisted one publishes with a warning naming the entry (the index has no veto, since it also drops old valid versions), and an unreachable index stays silent so offline re-push holds. Restores the push-time typo detection the short-circuit removed, without re-coupling publish availability to the index.

- The resolver contract is now grounded in reality: `crates/lns-artifact/src/tools/index_snapshots/` holds real index bodies for a shape-diverse tool set (java's vendor pins, go/jq's finished two-component releases, python/ruby's `-dev` heads), and a contract test proves every published line is a safe version whose resolution is a fixed point. The real data falsified the earlier "every resolver answer is exact" claim (jq ships `1.6` as a finished release), so the durable invariant is now stability — a published pin re-resolves to itself — with offline re-push narrowed to exact-classified pins. `bump-mise bump` refreshes the snapshots alongside `registry.snapshot` (validated before landing, same transactional write), so upstream shape drift turns the contract test red at the engine bump instead of at a user's push; runbook updated.

- **P0 regression caught and fixed by a new Layer-1 scenario:** the consent commit (7624f31e) made `sandbox::real::dispatch` take `Stdin::lock` while `run_from_matches` already held it, self-deadlocking every service-routed sandbox verb (`ps`, `stop`, `inspect`, `pull`, `rm`, `tag`, …) — Layer 2 never sees it because it enters below `dispatch` with a fake input. `dispatch` now borrows the caller's `RunCtx.input`. The full non-microvm e2e suite is green again (69/69; two `run_pull_confirm` assertions updated to the consent flow's new copy, and the pull steps now pass `--yes` and refuse pre-provision warnings).
- New `@microvm` scenario closing the cross-crate coherence window: publish a sandbox pinning `jq@1.7.1`, pull (pre-provisions in a real provisioner guest), delete `resolved.json`, cut the network — the run must start from the cached tree addressed by the published string alone and provision nothing. Verified live on macOS Vz (7/7 steps). `jq@1.6` was tried first as the fuzzy-shaped pin, but 1.6 ships no linux/arm64 build — that nuance stays pinned at Layer 3; the harness's `LNS_E2E_FEATURE` filter now also matches scenario names.

- Review fixes folded in (@jansav, two P0-shaped findings):
  - **`lns sandbox pull` could never be accepted interactively** (83e8283). `owns_terminal` was a per-*command* flag, so the `sandbox` parent claimed the terminal for all fourteen of its verbs on behalf of the four that stream over tokio stdio — the other ten were handed `io::empty()` as stdin, so the documented canonical spelling printed its disclosure and then read EOF. Only `lns pull` or `--yes` worked. Ownership is now decided per parsed invocation (`fn(&ArgMatches) -> bool`), which also fixes `push`/`rm`/`prune`. Pinned by three Layer-3 tests, including the invariant the bug broke: a shortcut and its `sandbox` spelling must get the same stdin. Neither existing layer could have caught it — Layer 2 enters below the dispatcher with a fake input, and Layer 1 can't distinguish the two builds without a real tty (both fail closed on `isatty` with a piped stdin), so "a human can type `y`" stays covered only by the manual `expect` smoke.
  - **`prune` deleted the tool resolution record** (a3476c8). `clear_runtime_cache` swept `<cache>/tools` wholesale, taking `resolved.json` with the cached trees — so reclaiming disk silently re-opened every pinned spec to a newer upstream version and dropped the `@latest` last-known-good the offline fallback depends on. The sweep now names the cache half (`tools/trees`, lifted into `tools::cache::TREES_DIR`); the reclaim total is unchanged. Pinned at Layer 3 and as a Layer-2 scenario phrased as the promise. The rig's old fixture seeded `tools/cached`, which didn't reflect the real layout — part of why this was invisible.


crates/lns-cli/src/sandbox/real.rs
```diff
@@ -277,7 +376,8 @@ impl SandboxService for RealSandboxService {
     }
 }
 
-pub async fn dispatch(args: super::SandboxArgs) -> Result<i32> {
+// The caller already holds the process-wide stdin lock (run_from_matches), so this must borrow it — a second Stdin::lock on the same thread deadlocks every dispatched verb.
+pub async fn dispatch(args: super::SandboxArgs, input: &mut dyn std::io::BufRead) -> Result<i32> {
```
jansav, line 399: **`lns sandbox pull` can never be accepted interactively.**

`dispatch` computes `term.stdin_is_tty` from the real tty but reads the answer from `ctx.input`. The `sandbox` parent `CommandSpec` sets `owns_terminal: true` (`crates/lns-cli/src/sandbox/mod.rs:246`), so `run_from_matches` hands this down `std::io::empty()` (`crates/lns-cli/src/lib.rs:38`).

**Why:** on a terminal, `lns sandbox pull ghcr.io/team/hermes:1.4.0` for a tools-declaring sandbox prints the disclosure and `Continue? [y/N]: `, immediately reads EOF, and exits non-zero with `declined; nothing was installed, mounted, or started`. The user never gets to answer. Only the shortcut spelling `lns pull` works (`owns_terminal: false`), and `--yes` is the sole escape — while `cli-reference.md` documents `lns sandbox pull [--yes]` as the canonical form. The e2e suite exercises just the shortcut, and only with `--yes`, which is how this got past a live run.

Options: derive `interactive` from whether the input is a real stdin rather than from `isatty`; hand the terminal-owning path a real stdin lock for the verbs that don't drive a tty; or move `pull` out from under the terminal-owning parent.

crates/lns-service/src/image_store/mod.rs
```diff
@@ -367,13 +367,72 @@ pub async fn prune_with<F: Fs, C: Caches>(
...
+    }
     Ok(PruneReport {
         removed,
         reclaimed_bytes,
     })
 }
 
+async fn clear_runtime_cache<F: RuntimeCacheFs>(fs: &F, cache_root: &Path) -> Result<u64> {
+    let mut reclaimed = 0;
+    for name in ["composefs", "tools", "content"] {
```
jansav: **`prune` deletes the tool record, not just the cache.**

Clearing `<cache>/tools` wholesale removes `tools/resolved.json` (`crates/lns-service/src/tools/record.rs:60`) — the machine's memory of what each bounded request first resolved to. That file is a record, not a cache; the code distinguishes the two everywhere else.

**Why:** `docs/running-workloads.md` promises `node@22` "resolves **once per machine** and stays there … even after upstream publishes 22.12.0. That is what makes a warm run reproducible." After `lns sandbox prune --force` — a disk-reclaim command — the next run of that same sandbox re-resolves and installs 22.12.0, so a toolchain version changes as a side effect of freeing disk. It also drops the `@latest` last-known-good that the documented offline-start fallback depends on.

Excluding `resolved.json` from the sweep keeps both the reclaim and the promise; otherwise the pinning docs need to say prune resets it.

jansav (commented):
**Problem:** Sandbox authors currently smuggle toolchains in through the base image or reinstall them over the network on every cold start. `spec.tools: [node@22]` declares them instead: provisioned once per machine by a pinned mise engine in a disposable guest, content-addressed, injected read-only, and prepended to the workload `PATH`.

**Approach:** Acquisition is treated as audited *system* egress with the same trust shape as pulling `spec.image` — declaration is consent, no policy route — while what a tool does at runtime stays in the policy cage. A resolved version is a validated newtype checked at every boundary it enters from (index, driver output, `resolved.json`, `manifest.json`); the provisioner guest is stopped before its staged tars are named final; cache reuse is gated on the requester declaring every tool that shared the tree's guest.

**Product alignment:** Fits "no system dependencies" and "ephemeral by default" — the tool cache is a host-side input, not guest state. The one genuinely new trust surface (third-party install code running as root with unrestricted egress) is disclosed in the consent card, the audit chain, and the docs rather than glossed.

The engineering here is unusually careful — the path/symlink/hardlink guards, the two-pass plan that keeps warm runs out of the install lock, and the stdout/stderr split are all correct for the reasons their comments claim. Two things need fixing.

**Verdict:** Needs fixes.

### Critical

**1. `lns sandbox pull` can never be accepted interactively** — see inline on `sandbox/real.rs`.

### Should fix

**2. `prune` deletes the tool record, silently unpinning every bounded version** — see inline on `image_store/mod.rs`.

crates/lns-cli/src/sandbox/real.rs
chenhunghan: Confirmed and fixed in 83e8283. You were right about the mechanism: `owns_terminal` was a per-*command* flag, so the `sandbox` parent claimed the terminal for all fourteen of its verbs on behalf of the four that actually stream over tokio stdio (`run`, `exec`, `logs`, `attach`). The other ten got `io::empty()` as stdin.

I took the first of your options, generalized: `owns_terminal` is now `fn(&ArgMatches) -> bool`, decided per parsed invocation rather than per command, and the sandbox parent answers `true` only for those four verbs. That fixes `pull` and also `push`/`rm`/`prune` — anything under the parent that ever needs to prompt now gets the dispatcher's real stdin lock.

Three pins in `command.rs`, all of which were red before the fix:
- `a_verb_that_prompts_never_owns_the_terminal_under_the_sandbox_parent` — `sandbox pull/push/prune/rm/ls` must not own the terminal.
- `a_shortcut_and_its_sandbox_spelling_agree_on_who_owns_the_terminal` — the invariant your finding actually broke: `lns <verb>` is documented as an exact shortcut for `lns sandbox <verb>`, so both spellings must get the same stdin. This is the one that would have caught it.
- `every_invocation_that_streams_over_tokio_stdio_owns_the_terminal` — replaces the old flag-shaped test, now argv-shaped for both spellings.

On why the existing suites missed it: the Layer 2 harness calls `run_with_writers` directly with its own `Cursor`, so it can't see the dispatcher's ctx wiring at all. Layer 1 can't distinguish it either without a real tty — with a piped stdin both the buggy and fixed builds fail closed on the `isatty` check with the same `--yes` message. So the honest coverage boundary is: the wiring decision is pinned at Layer 3, and "a human at a tty can type `y`" remains covered only by the manual `expect` smoke. I did not add an e2e for it.

`make lint && make complexity && make coverage` are green on the full workspace.

crates/lns-service/src/image_store/mod.rs
chenhunghan: Confirmed and fixed in a3476c8. `resolved.json` and the cached trees were already in separate subpaths — `<cache>/tools/resolved.json` vs `<cache>/tools/trees` — so the sweep just needed to name the cache half instead of the parent. It now clears `tools/trees`, and the record survives a `prune --force`. No exclusion filter needed, and the reclaim number is unchanged (the record isn't reclaimable space).

I also lifted the `trees` literal into `tools::cache::TREES_DIR` with a one-line note that `resolved.json` sits beside it and outlives a prune, so the next person moving that path has to think about the record.

Two pins, both red before the fix:
- Layer 3 `clearing_the_tool_cache_keeps_the_resolution_record` — the record survives, the tree does not, and the reclaim total excludes the record.
- Layer 2 `Pruning reclaims the tool cache but keeps what the machine resolved` in `image_management.feature` — the user-visible half, phrased as the promise rather than the path.

I mutation-proved both by reverting the sweep path: the Layer 3 test panics and the Layer 2 scenario fails on the two new steps.

One note on the fixture: the existing rig seeded the tool cache at `tools/cached`, which no longer falls inside the swept subtree, so `Pruning reclaims the provisioned tool cache` went red on the fix. I moved the rig's seed under `tools/trees` — the previous fixture didn't reflect the real layout, which is part of why this was invisible.

Also tightened the docs bullet in `running-workloads.md` to state it outright: prune reclaims the cached trees *without touching what the machine already resolved*, so the next run re-provisions the same versions. Your reading of the promise is the one the code now keeps.

crates/lns-cli/src/sandbox/real.rs
jansav: Resolved by 83e8283, and fixed at the right level — `owns_terminal` is now decided per parsed invocation instead of per command, so the `sandbox` parent claims the terminal only for `run`/`exec`/`logs`/`attach` and the ten prompting verbs get the dispatcher's real `stdin.lock()`. Every other spec moved onto the `never_owns_terminal`/`always_owns_terminal` helpers, so nothing was left on a stale literal.

Verified: forcing `verb_owns_terminal` back to `true` fails both new tests, the second naming `lns sandbox pull` exactly. The shortcut-vs-canonical agreement test is the one that would have caught this originally.

crates/lns-service/src/image_store/mod.rs
jansav: Resolved by a3476c8. The sweep now names `tools/trees` through a `TREES_DIR` constant that `RealToolCache` builds its own path from, so the cache half and the swept half cannot drift; `resolved.json` sits beside it and survives, and reclaimed bytes correctly exclude it. Covered at both layers plus the doc line ("the next run re-provisions what it needs — to the same versions").

Verified: reverting the path to `tools` fails the unit test on the "must not unpin what this machine already resolved" assertion.

jansav (approved):
Both findings addressed — verified against `83e8283`.

**1. `lns sandbox pull` consent (83e8283)** — fixed at the right level rather than patched at the call site: `owns_terminal` is now `fn(&ArgMatches) -> bool`, evaluated on the parsed invocation, so the `sandbox` parent claims the terminal only for `run`/`exec`/`logs`/`attach` and the ten prompting verbs get the dispatcher's real `stdin.lock()`. Every other spec moved onto the `never_owns_terminal`/`always_owns_terminal` helpers, so no site was left on a stale literal. The invariant is pinned three ways, including the shortcut-vs-canonical agreement test — the one that would have caught this originally.

**2. Prune keeping the tool record (a3476c8)** — the sweep names `tools/trees` through a `TREES_DIR` constant that `RealToolCache` builds its own path from, so the swept half and the cache half cannot drift; `resolved.json` survives and reclaimed bytes correctly exclude it. Covered at both layers, and the docs now say the next run re-provisions *to the same versions*.

Both fixes checked red-first: reverting `verb_owns_terminal` to `true` fails the two new CLI tests (the second naming `lns sandbox pull`), and reverting the sweep path to `tools` fails the record test on its "must not unpin what this machine already resolved" assertion. Local runs: lns-cli command specs 35/35, lns-service Layer 2 178/178; CI green on lint, test, and coverage for this head.

**Verdict:** Ship it. ✅

PR #177, opened 2026-07-28
feat(cli): machine-readable output on every list and status verb

jansav:
## What ships

`--format <table|json>` on nine verbs: `ps`, `sandbox ls`, `volume ls`, `policy list`, `connector list`, `connector grants`, `config list`, `config get`, `service status`. `lns audit` gains `--format <table|jsonl>` — a timeline is an event stream, so its machine-readable form stays one event per line, and `--json` survives as a hidden alias so existing scripts keep working byte-for-byte.

The contract: bare pretty-printed array (single object for status/inspect), camelCase keys, every key present with `null` rather than omitted, raw integers so nothing needs un-humanizing, and **exit codes unchanged** — `config get` on an unset key still exits 1, now emitting `[]`.

## Design

A shared `crates/lns-cli/src/output.rs` seam: each verb declares one row struct implementing `TableRow` + `Serialize`, and both renderings fall out of it. `render_table` moved here from `audit::table`, and three hand-rolled column-width implementations were deleted.

Rows are deliberately *views*, not the IPC or policy-file structs — serializing `RouteRule` or `VolumeInfo` directly would weld the public scripting contract to the wire protocol and the policy file format. The one deliberate exception is `PsRow.status`, which serializes `lns_ipc::RunStatus` itself precisely so `ps` and `inspect` agree on the shape (see below).

`--format` is a flattened per-verb clap arg rather than a global flag, so it never appears in `run`/`exec --help` where it would be silently ignored. List verbs and `audit` use separate value enums, so clap itself generates the right per-command help and the right invalid-value errors instead of a hand-written runtime rejection.

## Behavior changes worth calling out

- **`lns volume inspect` keys are renamed** snake_case → camelCase (`size_bytes` → `sizeBytes`, `disk_bytes` → `diskBytes`, `in_use_by` → `inUseBy`; `name` and `created` unchanged), for consistency with `lns inspect <run>`, which already shipped camelCase. This verb takes no `--format` flag, so the experimental-until-v1.0 disclaimer does not cover it — it is called out in its own "Changed in this release" docs subsection and in the commit body.

  **No compatibility shim, and that is a deliberate decision.** The obvious objection is that a script reading the old names gets `null` rather than an error — a silent wrong answer, the worst failure mode — and that this PR takes the opposite care with `lns audit --json`, which is kept byte-for-byte and pinned by a test. The asymmetry is intentional: `--json` is a *spelling* of a surface that stays, so aliasing it costs one hidden flag; `volume inspect`'s keys are the surface itself, and preserving them means either emitting every field twice or leaving one verb on `VolumeInfo`'s serialization while all eight others are camelCase. Both carry a second shape forward indefinitely. Since `lns` is pre-1.0 and nothing consumes this output today, there is no data or script to migrate, so paying that permanent cost buys nobody anything. The docs note and the commit body exist for the reader who wonders why the shape moved, not for a migration that has no audience. Future reviewer: if this surface ever does acquire consumers, this reasoning expires — the rename would then need the same alias treatment `--json` got.
- **`ps` and `inspect` now report the same `status` shape.** `ps` initially emitted a bare `"running"` string while `inspect` emits `RunStatus`, whose serde tag makes it `{"state": "running"}` — one `jq` status check could not have worked against both. `ps` was aligned to `inspect` rather than the reverse, since `inspect`'s shape already shipped and carries the exit code in `Exited { code }`.
- **`policy list` and `connector list` tables gain a header row**, replacing ad-hoc `allow  pattern  (desc)` and `id  (bundled)  kind` lines — a consequence of sharing the table renderer with the other list verbs.
- **`sandbox ls --format json` reports more than its table**: digest, size, layer count, pull time, and holder, all of which the service already computes.
- **`connector grants` keeps its tab-separated table byte-for-byte.** Its column set varies with `--all`, which a const `HEADERS` cannot express, and existing scenarios pin those exact fields — so only its JSON path goes through the seam.
- **`service status` now writes through the injected writer** instead of `println!`. It was the last list verb on real stdout, which is why its tests could previously assert only the request it sent, never the output it printed.
- `config unset` deliberately does *not* take `--format` — it prints a confirmation, so the flag would promise something it cannot honor.

## Testing

Ten commits, one per implementation step, each written test-first and reviewed before commit. Test count went from 274 → 294 scenarios and 758 → 764 unit tests. Two things the JSON deliberately omits are documented: empty-list prose becomes `[]`, and `config list`'s legacy-key warnings are dropped so stdout stays parseable.

Gate: `make lint`, `make complexity`, `make coverage` all green — lns-cli at 32 files/100%, **no new IGNORES entries**, the new `output.rs` at 100% from day one. `make e2e` 63/63; it caught a Layer 1 assertion the local gate doesn't run, now squashed into the commit that made the rename so no commit lands on main with a red e2e feature.


crates/lns-cli/src/config/mod.rs
```diff
@@ -223,21 +255,21 @@ fn unset(key: ConfigKey, path: &Path, writer: &mut impl Write) -> Result<i32> {
...
-            values_of(&cfg, key)
-                .into_iter()
-                .map(move |v| format!("{} = {v}", key.name()))
-        })
+        .flat_map(|&key| rows_for(key, &values_of(&cfg, key)))
         .collect();
-    if lines.is_empty() {
+    if output.format == crate::output::Format::Json {
+        crate::output::emit_object(&rows, writer)?;
+        return Ok(0);
```
chenhunghan, line 271: **`config list --format json` drops the legacy-key warning entirely.**

The JSON path returns before `legacy_entries`, so a config carrying dead `run.env` / `run.volume` / `run.publish` keys produces clean JSON and no signal at all — exactly the situation where a user is most likely to be confused about why their env isn't being applied.

Stdout does have to stay parseable, but `crate::log::warn!` already writes to stderr (`log.rs` wires every level there), so routing the warning through that keeps it in both formats with stdout still clean. It also gets the `⚠` line off stdout in table mode, which is where CLAUDE.md says warnings belong. Two lines, and the docs' "two things the table says that the JSON deliberately does not" becomes one.

crates/lns-cli/src/output.rs
```diff
@@ -0,0 +1,300 @@
...
+    for row in rows {
+        write_row(out, row, &widths)?;
+    }
+    Ok(())
+}
+
+fn write_row(out: &mut dyn Write, cells: &[String], widths: &[usize]) -> std::io::Result<()> {
+    let last = cells.len().saturating_sub(1);
+    for (i, cell) in cells.iter().enumerate() {
+        if i == last {
```
chenhunghan, line 93: **`policy list` now emits trailing whitespace on every rule without a description.**

`write_row` skips padding only for the *last* cell, so an empty last cell leaves the preceding column's separator dangling. Against the built binary:

```
VERDICT  PATTERN          DESCRIPTION$
allow    api.example.com  $
deny     evil.example     known bad$
```

Description-less rules are the common case, so most `lns policy list` rows regress from clean to trailing-space. `the_last_column_is_not_padded` (line 284) asserts exactly this invariant, but only with a non-empty last cell — so it reads as coverage of a property the code doesn't hold. Skip the separator when the remaining cells are all empty, and extend that test with an empty trailing cell.

docs/cli-reference.md
```diff
@@ -13,6 +13,48 @@ information at the terminal.
...
+- Two things the table says that the JSON deliberately does not: the empty-list
+  sentences (`No rules in …`) become `[]`, and `lns config list`'s warnings about
+  ignored legacy keys are omitted, so stdout stays parseable. Use table output to see
+  those.
+
+> **The JSON shape is experimental until v1.0.** Field names and shapes may change in a
+> minor release, so pin your `lns` version in scripts that depend on them. Table output
+> is for humans and carries no stability promise at all.
+
+### Changed in this release
```
chenhunghan: **A release-scoped section in a permanent reference doc.**

`### Changed in this release` is accurate for 0.18 and stale the moment 0.19 ships, and nobody has a reason to come back and delete it. The `volume inspect` key rename belongs in `CHANGELOG.md`; this section either drops or becomes a plain undated statement of the current key names.

chenhunghan (commented):
**Problem:** Only `lns audit` and `lns inspect` emitted machine-readable output, so scripting anything else meant parsing padded human tables. This closes the v1.0 roadmap item by putting `--format` on the nine list/status verbs.

**Approach:** One `crates/lns-cli/src/output.rs` seam — a `Format` enum, a flattened `OutputArgs`, a `TableRow` trait — so each verb declares a row struct once and both renderings fall out of it; three hand-rolled column-width implementations go away. Rows are deliberately views, not the IPC/policy structs, so the scripting contract isn't welded to the wire protocol. `--format` is per-verb rather than global so it never appears where it'd be silently ignored, and `lns audit` gets its own `table|jsonl` enum with `--json` kept as a hidden byte-for-byte alias.

**Product alignment:** Scripting & automation roadmap item. JSON is disclaimed experimental until v1.0, exit codes are unchanged, and nothing about the policy/approval model moves.

Verified locally in a clean worktree: 764 unit tests + 294 scenarios green, and I exercised the flag surface against a built `lns` — `--format` lands on exactly the nine verbs claimed, and not on `config unset`, `volume inspect`, `run`, or `exec`.

**Verdict:** Needs fixes. Three should-fix items, nothing critical.

crates/lns-cli/src/config/mod.rs
jansav: Fixed — the warning now goes through `crate::log::warn!` *before* the format branch, so it fires in both formats and lands on stderr. As you noted, that also gets the `⚠` line off stdout in table mode, which is where CLAUDE.md wants warnings.

Three tests had asserted it on stdout. Two Layer 3 tests moved to a `capture_warn` helper and I added `the_legacy_warning_never_reaches_the_parseable_stdout`, which pins that JSON stdout still parses *and* table stdout no longer carries the message. The Layer 2 scenario in `sandbox_adjacent.feature` was reframed to assert stdout does not contain it (the behaviours harness only captures stdout), with the content itself now pinned at Layer 3.

The docs bullet collapsed from two omissions to one, as you predicted. (commit e4d0627)

crates/lns-cli/src/output.rs
jansav: Fixed — `write_row` now finds the last *non-empty* cell with `rposition` and stops there, so a description-less rule ends at the pattern.

Your repro was exact; against the rebuilt binary it is now:

```
VERDICT  PATTERN          DESCRIPTION$
allow    api.example.com$
deny     evil.example     known bad$
```

You were right that `the_last_column_is_not_padded` was pinning a property the code did not hold. I added `an_empty_last_cell_leaves_no_separator_dangling` next to it and watched it fail with your exact output first. A reviewer flagged that `take(last + 1)` is itself a truncation risk the new test did not cover, so there is also `an_empty_middle_cell_still_renders_what_follows_it` pinning `["a", "", "c"]` → `"A  B  C\na     c\n"`. (commit 6bbfa10)

docs/cli-reference.md
jansav: Fixed — section deleted. You are right that nobody would have come back to remove it.

The rename now reaches `CHANGELOG.md` via release-please instead. Worth flagging that this needed more than moving the text: the commit footer said `BREAKING:`, which release-please does not recognise, so the rename would have been in neither place. It is now a proper `BREAKING CHANGE:` footer on the commit that made the rename (5a9a131), so the next release picks it up automatically.

The permanent reference doc just states the current key names, undated. (commit bf1c9e1)

chenhunghan (approved):

PR #179, opened 2026-07-29
fix: confine a root run-as and every exec session to the workload's identity

chenhunghan:
## Problem 1 — a root run-as keeps every capability

`lns run --sandbox-user root . -- sh -c '/.lens/nft flush ruleset'` succeeds. The workload tears down the nftables cage that redirects egress to the policy proxy, so network policy stops applying: no approval cards, no audit records, no gating.

The chain:

1. `resolve_run_as` (`crates/lns-service/src/vm/mod.rs:207`) accepts `root` verbatim — no uid-0 handling anywhere in lns-cli / lns-service / lns-init.
2. `lns-init` resolves the name against the image's `/etc/passwd` (`crates/lns-init/src/mount.rs:265`) and exports `LENS_RUN_UID=0` / `LENS_RUN_GID=0`.
3. `resolve_agent_creds` turns that into `Some(SandboxCredentials{uid:0,gid:0})`.
4. In lens-sandbox-core, `apply_privilege_drop` (`child_spawner.rs:85`) takes the `Some(creds)` arm and skips the `else if spec.is_root` cap drop:

```rust
if let Some(creds) = &spec.creds { creds.apply(cmd); }      // taken, uid 0
else if spec.is_root { privilege::apply_cap_drop(cmd); }    // the cap drop, skipped
```

`creds.apply` then runs `setgroups([]) + setgid(0) + setuid(0)` — all no-ops from root — followed only by `NO_NEW_PRIVS`. Its comment states the premise that breaks here: *"setuid(non-root) already zeroed effective/permitted/inheritable"*. That kernel side effect only fires for non-zero uids, so `CAP_NET_ADMIN` survives.

Two consequences, not one:

- **Network policy bypass** — `CAP_NET_ADMIN` in the guest's single netns flushes the cage.
- **Secret exposure** — `CAP_SYS_PTRACE` defeats `harden_supervisor`'s `PR_SET_DUMPABLE(0)` (non-dumpable blocks same-uid ptrace, not `CAP_SYS_PTRACE`), reaching the supervisor's relay token and MITM CA key. That is squarely against "real secrets stay outside the workload".

Same trigger without any flag: an image whose `USER` is `root` or `0` takes the identical path.

## Fix

Never hand core a uid-0 credential — but keep it for everything else. `crates/lns-supervisor/src/run_as.rs` decides the setuid target; the dispatcher keeps the full credentials for `HOME`/`USER` (`build_agent_env`) and the agent cwd (`resolve_cwd`), while only `ChildSpec.creds` and the exec manager get the filtered value. A root run-as therefore reaches core's `apply_cap_drop` while keeping the environment the flag exists to provide.

Deciding at the spawn boundary rather than at credential resolution covers both spawn paths — the piped one at `child_spawner.rs:88` and the hand-rolled PTY closure at `pty.rs:126`, so `-it` is fixed too — and `lns exec` sessions, which take the same credentials through `ExecManager`.

Net effect for a root workload: keeps uid 0, `HOME`, `USER`, cwd, and `KEEP_CAP_MASK` (`CHOWN`, `DAC_OVERRIDE`, `FOWNER`, `FSETID`, `KILL`, `SETUID`, `SETGID`) so package installs and chowns still work; loses `NET_ADMIN` and `SYS_PTRACE`. `NO_NEW_PRIVS` and pdeathsig are unchanged — `apply_cap_drop` installs both.

Guarding at the consumer keeps this to one repo: no core change and no pin bump, so core stays at `4cb1d974` from #178.

### One accepted behaviour change

`-u root:<group>` no longer honours the group. The caps-dropped path never calls `setgid`, so the workload inherits the supervisor's gid 0 where it previously got the requested gid. lns-init still resolves and exports it. Honouring a group without a setuid needs a core change; the combination looks rare enough to defer, and the alternative is leaving `NET_ADMIN` in place.

## Problem 2 — `lns exec` was never confined at all

Found while pinning the exec half of problem 1, and worse than it. `handle_exec` (`crates/lns-service/src/ipc/adapter.rs:621`) connects to `lns_session::BROKER_PORT`, so an exec is forked by the **session broker**, not the supervisor — it never met the privilege drop at all. `exec_child` (`crates/lns-session-broker/src/session/real.rs`) went from `close_range`/env straight to `execvp`.

Probed on `main`, exec'ing into a **default** sandbox with no `-u` anywhere:

```
Uid:      0  0  0  0            # the workload itself runs as 65534
CapEff:   000001ffffffffff
```

So `lns exec <run> -- /.lens/nft flush ruleset` took the cage down on any sandbox. Pre-existing and independent of problem 1 — the supervisor's `ExecManager` is simply not on this path.

## Fix 2 — carry the intent on the wire

`ClientFrame::OpenSession` gains `confine: bool`. When set, the broker applies the identity lns-init resolved (`LENS_RUN_UID`/`LENS_RUN_GID`): `setuid` for a non-root workload, `KEEP_CAP_MASK` for a root one, `NO_NEW_PRIVS` either way. `lns exec` sets it; the primary session leaves it false because its workload *is* the supervisor, which needs root and `CAP_NET_ADMIN` to install the cage in the first place.

Two design points worth stating:

- The decision resolves **before** the fork, so the post-fork child never takes the environment lock. It lives as a pure `confinement()` in `session.rs` (100% covered); only the syscalls sit in the platform leaf that is already in IGNORES.
- `KEEP_CAP_MASK` is duplicated in the broker rather than depending on lens-sandbox-core — the broker is a static-musl PID-1 sibling and pulling in core's async surface for one constant isn't worth it. The constant carries a comment naming core as the source of truth.

Not changed: the unsupervised primary session (no policy → no supervisor) still runs its workload unconfined, and the internal stats probe stays unconfined since its argv is service-controlled and unreachable from the workload. Both are noted as follow-ups rather than silently altered.

## Problem 3 — the relay token rode into every exec

`exec_child` applied the session env *on top of* the inherited environment, so `LENS_SANDBOX_TOKEN` and `LENS_SANDBOX_WS_URL` reached every exec'd command through its own `environ` — including the one fix 2 had just dropped to uid 65534 with no capabilities.

That token is the whole supervisor↔workload trust boundary inside the guest: vsock exposes no peer identity, so the host cannot tell the supervisor from the workload on port 1024, and the bearer token (per run, constant-time compared at `relay/mod.rs:306`) is the only discriminator. A workload holding it can approve its own egress, request credential injection, and write audit entries — defeating the policy engine without touching nftables. Removing `CAP_SYS_PTRACE` in fix 1 was done specifically to keep that token out of a workload's reach.

The cmdline leg was already closed: `lns-init` bind-mounts a sanitized `/.lens/.cmdline` over `/proc/cmdline` (`mount.rs:29,609-618`), verified in a booted guest. `environ` was the one remaining door.

## Fix 3 — scrub the inherited internal vars

A confined session unsets inherited `LENS_SANDBOX_*` before applying its own env, filtering on the same `INTERNAL_VAR_PREFIX` core scrubs before spawning the agent (`privilege.rs:11`). `clearenv()` would have taken `PATH`/`HOME` with it; `LENS_RUN_UID`/`GID` survive the prefix, matching what the agent already sees. Keys are selected before the fork, beside the confinement decision, so the child never walks the environment.

Gated on `Confinement` for the same reason as fix 2: `Inherit` keeps them, because a supervised primary session *is* the supervisor.

Also corrected here: `confine` for the primary session is now `session.is_none()` rather than a hardcoded `false`. With no resolved policy path `start_if_policy` returns `None` and the broker execs the *user's* command as the primary session, so the constant silently ignored `--sandbox-user` and dropped nothing. Deriving it from the value that already decides `argv` keeps the two from drifting. That path has no `@microvm` coverage — every e2e run resolves a policy — so it is structurally tied rather than test-pinned.

## Tests

Red first, at both seams. The capability decision:

```
a_root_run_as_is_not_a_setuid_target ... FAILED
  left: Setuid { uid: 0, gid: 0 }
 right: RootWithCapsDropped
```

Then the identity regression the first pass introduced:

```
a_root_run_as_keeps_its_identity_but_not_its_setuid ... FAILED
  at dispatcher/agent.rs:399: spec.creds.is_none()
```

**Layer 3** — `run_as.rs` (100%): root uid is not a setuid target regardless of gid; non-root uid is, even in group 0; unresolvable ids have no target. `dispatcher/agent.rs` (100%): a root run-as yields no `ChildSpec.creds` but keeps `HOME`, `USER` and a home cwd; a non-root run-as still setuids.

**Layer 1** — `microvm_privilege.feature` (`@microvm`): a root run-as is honoured as an identity; it keeps `HOME`/`USER`; it cannot flush the ruleset; and its `CapEff` is `00000000000000fb`. Verified on a real guest, 8 scenarios / 34 steps green via `LNS_E2E_FEATURE=unprivileged make e2e-microvm`.

The capability assertion replaced an earlier `cat /proc/1/mem` probe that could not fail: the read returns `EIO` from the unmapped offset 0 regardless of capabilities, and PID 1 is the session broker rather than the supervisor holding the placeholder map. Reading `CapEff` pins the real invariant — `0xfb` is `KEEP_CAP_MASK` exactly, so `NET_ADMIN` (bit 12) and `SYS_PTRACE` (bit 19) are both provably absent where the pre-fix workload carried the full `000001ffffffffff`.

Observed in the guest under `--sandbox-user root`, for the record:

```
Uid:         0
CapPrm:      00000000000000fb
CapEff:      00000000000000fb
CapBnd:      000001ffffffffff
CapAmb:      0000000000000000
NoNewPrivs:  1
```

### Why the existing suite did not catch this

Worth recording, because the first pass shipped a regression past a green gate:

- `resolve_agent_creds` lives in `main.rs`, which is in the coverage IGNORES as a guest-only bootstrap. The function whose semantics changed had no test at any layer.
- The gate measures lines, not behaviour. `build_agent_env`'s `if let Some(creds)` arms were already covered from both directions by existing tests, so changing which arm production takes at runtime moved no hit count. Coverage stayed at 100% while the behaviour broke.
- The tests that would have caught it are `@microvm`, which the required suite does not run.

The new Layer 3 tests close the first two by asserting across the seam instead of on either side of it.

### Fix 2's tests

**Layer 3** — `confinement()`: an unconfined session inherits, a confined non-root session setuids, a confined root session caps instead, and missing ids still cap. `build_session_params_asks_the_broker_to_confine_the_exec` pins the service side. `OpenSession` round-trips the flag.

**Layer 1** — two `@microvm` scenarios through a new `-u`-carrying detached-start step: an exec into a default sandbox reports `uid=65534`, and an exec into a root sandbox reports `CapEff: …fb`.

**Mutation-checked.** Flipping `confine` back to `false` fails exactly those two scenarios and nothing else.

### Fix 3's tests

**Layer 3** — `keys_to_scrub()`: a confined session scrubs exactly the two relay vars while keeping `PATH`/`HOME`/`LENS_RUN_UID`, a setuid session scrubs them too, an unconfined session keeps them.

**Layer 1** — `an exec does not inherit the supervisor's relay credentials` asserts an exec's own `env` carries neither var.

**Mutation-checked.** Disabling the prefix filter fails exactly that scenario.

### Verification

Full `@microvm` suite: **70/73**. The 3 failures are pre-existing and unrelated — 2 stale credential specs (#181, broken on `main` since #170) and a local-registry `503`. All pre-existing exec scenarios in `microvm_streaming.feature` still pass, so confining an exec to uid 65534 broke nothing.

## Follow-up, not in this PR

- `CapBnd` stays at the full `000001ffffffffff` because `KEEP_CAP_MASK` excludes `CAP_SETPCAP`, so the bounding set cannot be narrowed from the child. Not exploitable on its own — `CapInh` and `CapAmb` are zero and `NoNewPrivs` is 1, so nothing can promote a bounding-set bit into permitted — but it is the same gap as the item below.
- The unsupervised primary session is now confined, but nothing in the harness can boot a run without a resolved policy, so that path stays unverified end to end.
- The core restructure — registering `apply_cap_drop` *before* `creds.apply` — additionally drops the **bounding set** on non-root runs, which `creds.apply` admits it cannot do today (`privilege.rs:294`, `CAP_SETPCAP` is already gone by then). Worth a `lens-sandbox-core` PR so every consumer gets it instead of each one guarding uid 0.
- `host_known_workload_gid` (`vm/mod.rs:260`) returns `Some` only for the exact default `sandbox`/65534, so any `-u` override serves Linux binds 1:1 with no virtiofsd `--uid-map` (`orchestrate.rs:58`). Separate bug.

### DIFF
--- a/crates/e2e-tests/features/microvm_privilege.feature
+++ b/crates/e2e-tests/features/microvm_privilege.feature
@@ -61,6 +61,15 @@ Feature: the workload runs unprivileged inside a locked-down guest
     Then the exit code is 0
     And the output contains "00000000000000fb"
 
+  Scenario: an exec does not inherit the supervisor's relay credentials
+    Given the Lens Sandbox service is running
+    When the user starts a detached microVM command "/bin/sh -c '/.lens/guest-tools/bin/busybox sleep 60'"
+    Then the exit code is 0
+    When the user execs "/bin/sh -c '/.lens/guest-tools/bin/busybox env'" in that run
+    Then the exit code is 0
+    And the output does not contain "LENS_SANDBOX_TOKEN"
+    And the output does not contain "LENS_SANDBOX_WS_URL"
+
   Scenario: /run is a fresh tmpfs, not the workload's persistent root
     Given the Lens Sandbox service is running
     When the user runs a microVM command "/bin/sh -c '/.lens/guest-tools/bin/busybox grep /run /proc/mounts'"
@@ -72,4 +81,5 @@ Feature: the workload runs unprivileged inside a locked-down guest
     When the user runs a microVM command "/bin/sh -c '/.lens/guest-tools/bin/busybox true && echo tooling-ok-$((7*6))'"
     Then the exit code is 0
     And the output contains "tooling-ok-42"
+
 
--- a/crates/e2e-tests/features/microvm_privilege.feature
+++ b/crates/e2e-tests/features/microvm_privilege.feature
@@ -22,9 +22,9 @@ Feature: the workload runs unprivileged inside a locked-down guest
 
   Scenario: a root run-as is honoured as an identity
     Given the Lens Sandbox service is running
-    When the user runs a microVM command "/bin/sh -c '/.lens/guest-tools/bin/busybox id -u'" as user "root"
+    When the user runs a microVM command "/bin/sh -c 'echo uid=$(/.lens/guest-tools/bin/busybox id -u)'" as user "root"
     Then the exit code is 0
-    And the output contains "0"
+    And the output contains "uid=0"
 
   Scenario: a root run-as keeps HOME and USER
     Given the Lens Sandbox service is running
@@ -56,3 +56,4 @@ Feature: the workload runs unprivileged inside a locked-down guest
     When the user runs a microVM command "/bin/sh -c '/.lens/guest-tools/bin/busybox true && echo tooling-ok-$((7*6))'"
     Then the exit code is 0
     And the output contains "tooling-ok-42"
+
--- a/crates/e2e-tests/features/microvm_privilege.feature
+++ b/crates/e2e-tests/features/microvm_privilege.feature
@@ -45,6 +45,22 @@ Feature: the workload runs unprivileged inside a locked-down guest
     Then the exit code is 0
     And the output contains "00000000000000fb"
 
+  Scenario: an exec lands on the workload's identity, not the broker's root
+    Given the Lens Sandbox service is running
+    When the user starts a detached microVM command "/bin/sh -c '/.lens/guest-tools/bin/busybox sleep 60'"
+    Then the exit code is 0
+    When the user execs "/bin/sh -c 'echo uid=$(/.lens/guest-tools/bin/busybox id -u)'" in that run
+    Then the exit code is 0
+    And the output contains "uid=65534"
+
+  Scenario: an exec into a root sandbox is capped like the workload it joins
+    Given the Lens Sandbox service is running
+    When the user starts a detached microVM command "/bin/sh -c '/.lens/guest-tools/bin/busybox sleep 60'" as user "root"
+    Then the exit code is 0
+    When the user execs "/bin/sh -c '/.lens/guest-tools/bin/busybox grep CapEff /proc/self/status'" in that run
+    Then the exit code is 0
+    And the output contains "00000000000000fb"
+
   Scenario: /run is a fresh tmpfs, not the workload's persistent root
     Given the Lens Sandbox service is running
     When the user runs a microVM command "/bin/sh -c '/.lens/guest-tools/bin/busybox grep /run /proc/mounts'"
--- a/crates/e2e-tests/features/microvm_privilege.feature
+++ b/crates/e2e-tests/features/microvm_privilege.feature
@@ -39,10 +39,11 @@ Feature: the workload runs unprivileged inside a locked-down guest
     Then the exit code is non-zero
     And the output contains "Operation not permitted"
 
-  Scenario: a root workload cannot read the supervisor's memory
+  Scenario: a root workload keeps only the identity-management capabilities
     Given the Lens Sandbox service is running
-    When the user runs a microVM command "/bin/sh -c '/.lens/guest-tools/bin/busybox cat /proc/1/mem'" as user "root"
-    Then the exit code is non-zero
+    When the user runs a microVM command "/bin/sh -c '/.lens/guest-tools/bin/busybox grep CapEff /proc/self/status'" as user "root"
+    Then the exit code is 0
+    And the output contains "00000000000000fb"
 
   Scenario: /run is a fresh tmpfs, not the workload's persistent root
     Given the Lens Sandbox service is running
--- a/crates/e2e-tests/features/microvm_privilege.feature
+++ b/crates/e2e-tests/features/microvm_privilege.feature
@@ -26,6 +26,13 @@ Feature: the workload runs unprivileged inside a locked-down guest
     Then the exit code is 0
     And the output contains "0"
 
+  Scenario: a root run-as keeps HOME and USER
+    Given the Lens Sandbox service is running
+    When the user runs a microVM command "/bin/sh -c 'echo home=[$HOME] user=[$USER]'" as user "root"
+    Then the exit code is 0
+    And the output does not contain "home=[]"
+    And the output does not contain "user=[]"
+
   Scenario: a root workload cannot tear down the network cage
     Given the Lens Sandbox service is running
     When the user runs a microVM command "/bin/sh -c '/.lens/nft flush ruleset'" as user "root"
--- a/crates/e2e-tests/features/microvm_privilege.feature
+++ b/crates/e2e-tests/features/microvm_privilege.feature
@@ -20,6 +20,23 @@ Feature: the workload runs unprivileged inside a locked-down guest
     Then the exit code is 0
     And the output contains "who=sandbox"
 
+  Scenario: a root run-as is honoured as an identity
+    Given the Lens Sandbox service is running
+    When the user runs a microVM command "/bin/sh -c '/.lens/guest-tools/bin/busybox id -u'" as user "root"
+    Then the exit code is 0
+    And the output contains "0"
+
+  Scenario: a root workload cannot tear down the network cage
+    Given the Lens Sandbox service is running
+    When the user runs a microVM command "/bin/sh -c '/.lens/nft flush ruleset'" as user "root"
+    Then the exit code is non-zero
+    And the output contains "Operation not permitted"
+
+  Scenario: a root workload cannot read the supervisor's memory
+    Given the Lens Sandbox service is running
+    When the user runs a microVM command "/bin/sh -c '/.lens/guest-tools/bin/busybox cat /proc/1/mem'" as user "root"
+    Then the exit code is non-zero
+
   Scenario: /run is a fresh tmpfs, not the workload's persistent root
     Given the Lens Sandbox service is running
     When the user runs a microVM command "/bin/sh -c '/.lens/guest-tools/bin/busybox grep /run /proc/mounts'"
--- a/crates/lns-service/src/run/orchestrator.rs
+++ b/crates/lns-service/src/run/orchestrator.rs
@@ -395,7 +395,7 @@ async fn orchestrate(
         tty: args.tty,
         stdin: args.stdin,
         initial_winsize,
-        confine: false,
+        confine: session.is_none(),
     };
 
     let frame_tx_for_session = frame_tx.clone();
--- a/crates/lns-service/src/run/orchestrator.rs
+++ b/crates/lns-service/src/run/orchestrator.rs
@@ -395,6 +395,7 @@ async fn orchestrate(
         tty: args.tty,
         stdin: args.stdin,
         initial_winsize,
+        confine: false,
     };
 
     let frame_tx_for_session = frame_tx.clone();
--- a/crates/lns-session-broker/src/session/real.rs
+++ b/crates/lns-session-broker/src/session/real.rs
@@ -582,6 +582,7 @@ mod tests {
             tty: false,
             stdin: false,
             winsize: None,
+            confine: false,
         };
         let bytes = encode_frame(&frame).expect("encode OpenSession");
         // SAFETY: bytes is borrowed for the duration of the call.
--- a/crates/lns-session-broker/src/session/real.rs
+++ b/crates/lns-session-broker/src/session/real.rs
@@ -14,8 +14,9 @@ const PTY_DRAIN_GRACE: Duration = Duration::from_millis(500);
 const HOST_DRAIN_GRACE: Duration = Duration::from_secs(2);
 
 use super::{
-    LoopAction, SessionError, SessionOutcome, SharedFd, WorkloadSpec, close, dispatch_frame,
-    read_client_frame, signal_target, validate_argv, validate_open_session,
+    Confinement, LoopAction, SessionError, SessionOutcome, SharedFd, WorkloadSpec, close,
+    confinement, dispatch_frame, read_client_frame, signal_target, validate_argv,
+    validate_open_session,
 };
 use crate::forker::{Fork, Forker};
 use crate::pty;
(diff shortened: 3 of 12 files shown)


crates/e2e-tests/features/microvm_privilege.feature
```diff
@@ -20,6 +20,31 @@ Feature: the workload runs unprivileged inside a locked-down guest
     Then the exit code is 0
     And the output contains "who=sandbox"
 
+  Scenario: a root run-as is honoured as an identity
+    Given the Lens Sandbox service is running
+    When the user runs a microVM command "/bin/sh -c '/.lens/guest-tools/bin/busybox id -u'" as user "root"
+    Then the exit code is 0
+    And the output contains "0"
```
jansav: **`the output contains "0"` cannot fail.**

The assertion runs against the whole `lns run` invocation's stdout+stderr, which includes `started run <hex>` and progress/timing lines — a bare `"0"` matches unrelated output and doesn't distinguish uid 0 from any other uid. The rest of this feature already uses a distinguishing prefix for exactly this reason (`uid=65534`, `who=sandbox`, `tooling-ok-42`).

**Why:** this is the scenario that pins "a root run-as is still honoured as an identity" — the half of the fix the second commit had to add back after the first pass regressed it. As written it would stay green if `HOME`/`USER`/uid handling broke again. Use `echo uid=$(… id -u)` and assert `uid=0`.

crates/lns-supervisor/src/dispatcher/agent.rs
```diff
@@ -61,8 +61,11 @@ impl AgentDispatcher {
         reaper: Arc<OrphanReaper>,
         runner: Arc<dyn AgentRunner>,
     ) -> Self {
-        let exec_manager =
-            ExecManager::new(sandbox_creds.clone(), config.core.is_root, reaper.guard());
+        let exec_manager = ExecManager::new(
+            crate::run_as::setuid_creds(sandbox_creds.as_ref()),
```
jansav, line 65: **The `lns exec` half of the fix is unpinned at every layer.**

`ExecManager::new` gets the filtered credentials, but nothing asserts it: `agent_child_spec` has a Layer 3 test, `ExecManager` has none, and the new `@microvm` scenarios only drive `lns run`. Reverting this line to `sandbox_creds.clone()` leaves the whole gate green while `lns exec` into a root sandbox regains `CAP_NET_ADMIN`.

**Why:** the PR's own "why the existing suite did not catch this" section names this exact failure mode — a security-relevant argument with no test across the seam. `ExecManager` doesn't expose its credentials for assertion, so the practical pin is a `@microvm` scenario: start a detached run as root, `exec` `grep CapEff /proc/self/status`, assert `…fb`. The exec-into-that-run steps already exist (`microvm_streaming.feature:12`); only a `-u`-carrying detached-start step is missing.

jansav (commented):
**Problem:** `--sandbox-user root` (or an image with `USER root`) resolved to uid-0 credentials, which took core's `creds.apply` branch instead of `apply_cap_drop`. `setuid(0)` is a no-op that doesn't zero the capability sets, so the workload kept `CAP_NET_ADMIN` (flush the nftables cage → no network policy) and `CAP_SYS_PTRACE` (defeat `PR_SET_DUMPABLE(0)` → reach the relay token and MITM CA key).

**Approach:** Filter uid 0 out of the credentials at the two spawn boundaries (`agent_child_spec` and `ExecManager::new`) while keeping the full credentials for `HOME`/`USER` and the cwd. `ChildSpec.creds: None` + `is_root: true` routes to `apply_cap_drop`, which covers both the piped and PTY paths as well as `lns exec`. No core change, no pin bump.

**Product alignment:** Directly restores "policy you run into, not write" and "real secrets stay outside the workload" for root workloads.

I checked the parts that matter: the cage is mark-based (not uid-based), so `KEEP_CAP_MASK` = `0xfb` genuinely removes the escape (`NET_ADMIN`/`NET_RAW`) and `SYS_PTRACE`; `NoNewPrivs=1` + `CapInh`/`CapAmb`=0 makes the untouched bounding set inert, so the follow-up note is accurate; every `ChildSpec` in the crate flows through `agent_child_spec`; and the unfiltered credentials that still reach `ProxyServer`/`temp_files` are used for chowning injected credential files, where uid 0 is the right answer. Filtering on the resolved uid rather than the name also closes `-u 0` and an image `/etc/passwd` that aliases a name to uid 0.

**Verdict:** Needs fixes — two test-strength issues, both on the seams this PR is specifically about.

crates/e2e-tests/features/microvm_privilege.feature
chenhunghan: Fixed — the scenario now echoes the uid behind a prefix and asserts `uid=0`, matching the convention the rest of the feature uses. You were right that a bare `"0"` matched the run id, the timing lines and the boot line, so it could not fail. (commit 87f651e)

Worth noting this was the second vacuous assertion in the same feature: `cannot read the supervisor's memory` asserted only a non-zero exit, which `cat /proc/1/mem` returns from the unmapped offset 0 regardless of capabilities — and PID 1 is the session broker, not the supervisor. That one is now `grep CapEff /proc/self/status` asserting `00000000000000fb`, which is `KEEP_CAP_MASK` exactly, so it pins the absence of `NET_ADMIN` and `SYS_PTRACE` together and would have shown `000001ffffffffff` before the fix. (commit d04919d)

crates/lns-supervisor/src/dispatcher/agent.rs
chenhunghan: You were right to distrust this line, and following it up found something worse than an untested argument: **the `ExecManager` filter has no effect at all, because `lns exec` never reaches the supervisor.** `handle_exec` (`crates/lns-service/src/ipc/adapter.rs:621`) connects to `lns_session::BROKER_PORT`, so an exec is forked by the session broker, and `exec_child` (`session/real.rs`) went straight from `close_range`/env to `execvp` with no setuid and no capability drop.

So the state on `main` was: an exec into **any** sandbox — no `-u` needed — ran as root with `CapEff: 000001ffffffffff`. I probed it before changing anything:

```
Uid:      0  0  0  0            # workload runs as 65534
CapEff:   000001ffffffffff
```

That means `lns exec <run> -- /.lens/nft flush ruleset` took the cage down on a default sandbox. Pre-existing, unrelated to the root-run-as bug this PR opened for.

Fixed properly (commit f311e6d): `OpenSession` gained a `confine` flag. When set, the broker applies the identity lns-init resolved — `setuid` for a non-root workload, `KEEP_CAP_MASK` for a root one — before `execvp`. The primary session leaves it false, because its workload *is* the supervisor and it needs root plus `CAP_NET_ADMIN` to install the cage. The decision is resolved before the fork (the child never touches the env lock) and lives as a pure `confinement()` in `session.rs` at 100%; the syscalls stay in the platform leaf.

Pinned at both layers you asked for:

- **Layer 3** — `confinement()`: unconfined inherits, confined non-root setuids, confined root caps instead, missing ids cap. Plus `build_session_params_asks_the_broker_to_confine_the_exec` in lns-service.
- **Layer 1** — two `@microvm` scenarios via the new `-u`-carrying detached-start step: an exec into a default sandbox reports `uid=65534`, and an exec into a root sandbox reports `CapEff: …fb`.

And the mutation check you implied: flipping `confine` back to `false` fails exactly those two scenarios and nothing else. Full `@microvm` suite is 69/72 — the 3 failures are lensapp/lens-sandbox#181 (stale credential specs, pre-existing) and a local-registry 503.

crates/lns-session-broker/src/session/real.rs
```diff
@@ -412,6 +504,10 @@ fn exec_child(spec: &WorkloadSpec) -> ! {
             unsafe { libc::putenv(raw) };
         }
     }
+    if let Err(e) = apply_confinement(&spec.confinement) {
```
jansav, line 516: **The exec'd command inherits the relay token from the broker's environment.**

`exec_child` `putenv`s the session env *on top of* the inherited environment and never clears it. `LENS_SANDBOX_TOKEN=…` and `LENS_SANDBOX_WS_URL=…` ride the kernel cmdline (`vm/mod.rs:310`), which the kernel hands to init as envp, which `fexecve(fd, argv, environ)` (`lns-init/src/mount/real.rs:141`) passes to the broker. So every `lns exec` command — now correctly dropped to uid 65534 with no capabilities — still gets the relay token handed to it in its own `environ`.

**Why:** this PR removes `CAP_SYS_PTRACE` specifically so a root workload can't read the relay token out of the supervisor, and cites that as one of its two consequences. Meanwhile an unprivileged `lns exec` reads it out of `/proc/self/environ` for free and can talk to the host relay as if it were the supervisor. The supervised primary session is safe only because the supervisor scrubs `LENS_SANDBOX_*` before spawning the agent (`scrub_internal_vars`, pinned at `dispatcher/agent.rs:884`) — the broker's direct-exec paths have no equivalent.

Pre-existing, not introduced here, so scope it as you see fit — but it's the same function this commit rewrites and the same threat model, and the fix mirrors what core already does: `clearenv()` (or filter `LENS_SANDBOX_*`) before the `putenv` loop.

crates/lns-service/src/run/orchestrator.rs
```diff
@@ -395,6 +395,7 @@ async fn orchestrate(
         tty: args.tty,
         stdin: args.stdin,
         initial_winsize,
+        confine: false,
```
jansav: **`confine: false` is wrong when the run is unsupervised.**

The field's doc comment says "false for the primary session, whose workload is the supervisor" — but that's only true when `session.is_some()`. `build_workload_argv(…, supervised)` returns the supervisor only in that case; with no resolved policy path (`start_if_policy` → `None`, pinned at `supervisor/mod.rs:307`) the broker execs the *user's* command as the primary session, so `confine: false` means `--sandbox-user` is silently ignored and no capabilities are dropped.

**Why:** the new flag makes it look like the primary session's privileges are a decided question when they aren't, and the honest value is free — `session.is_some()` is already computed a few lines up for `argv`, so `confine: !supervised` closes it. There's no cage in unsupervised mode, so this isn't a network-policy bypass, but "`-u sandbox` runs as root" is a promise broken in the one place the code now has the machinery to keep it.

jansav (commented):
Re-reviewed at `541d090`. Both earlier findings are addressed, and the second one turned up something bigger — nice catch on the broker.

`uid=0` now carries a distinguishing prefix, matching the feature's own idiom. On the exec side my aim was slightly off: `lns exec` never reaches the supervisor's `ExecManager` at all — it opens a second broker session, and `exec_child` did `fork` + `execvp` with no privilege drop whatsoever, so `lns exec` into *any* sandbox (including the default `sandbox` user) ran as root with the full capability set. The fix looks right: the `confine` flag on `ClientFrame::OpenSession`, a pure `confinement()` decision in `session.rs` with five Layer 3 tests, the syscalls confined to the linux-only `real.rs`, and two `@microvm` scenarios (exec lands on uid 65534; exec into a root sandbox is capped to `…fb`, which also pins the broker's `KEEP_CAP_MASK` against core's). The setuid ordering (setgroups → setgid → setuid → `NO_NEW_PRIVS`) is correct, `child_exit(126)` fails closed, and putting the decision in the testable module while leaving only syscalls in the platform file is the right split.

Two findings on the new code — one critical (pre-existing, but on the seam this commit rewrites), one small.

**Verdict:** Needs fixes.

crates/lns-session-broker/src/session/real.rs
chenhunghan: Fixed — a confined session now unsets the inherited `LENS_SANDBOX_*` before the `putenv` loop. (commit f2178c8)

You were right about the channel and the threat model. Two refinements from digging into it:

**The cmdline leg is already closed.** The token does travel on the kernel cmdline, but `lns-init` writes a sanitized copy to `/.lens/.cmdline` and bind-mounts it over `/proc/cmdline` (`mount.rs:29,609-618`). I confirmed in a booted guest that `grep -c LENS_SANDBOX_TOKEN /proc/cmdline` finds nothing even as root. So `environ` was the *only* remaining channel — which makes your point sharper: the codebase already decided to defend this secret and missed one door.

**Why it matters, spelled out for the PR record:** vsock exposes no peer identity — no `SO_PEERCRED` equivalent — so the host cannot distinguish the supervisor from the workload on port 1024. The bearer token (`relay/mod.rs:306`, constant-time, per run) is the entire discriminator. A workload holding it could approve its own egress, request credential injection, and write audit entries — defeating the policy engine without touching nftables at all.

`clearenv()` would have taken `PATH`/`HOME` with it and broken exec'd commands, so this filters on `LENS_SANDBOX_` — the same `INTERNAL_VAR_PREFIX` core scrubs (`privilege.rs:11`), named as the source of truth in a comment. `LENS_RUN_UID`/`GID` survive the prefix, matching what the agent already sees. The keys are selected before the fork, beside the confinement decision, so the child never walks the environment.

Gated on `Confinement`: `Inherit` keeps them, since a supervised primary session *is* the supervisor.

**Layer 3** — three tests: a confined session scrubs exactly the two relay vars and keeps `PATH`/`HOME`/`LENS_RUN_UID`, a setuid session scrubs them too, an unconfined one keeps them. **Layer 1** — `an exec does not inherit the supervisor's relay credentials` asserts an exec's `env` carries neither var. Mutation-checked: disabling the filter fails exactly that scenario.

crates/lns-service/src/run/orchestrator.rs
chenhunghan: Fixed — `confine: session.is_none()`, derived from the same value that already decides `argv` and `composed` two lines up, so the two cannot drift. (commit 3bbf72d)

You're right that I had this backwards: I documented the unsupervised gap in the PR body as a deliberate deferral, but the reason was that unsupervised runs have no e2e coverage — and that's an argument for noting the gap, not for shipping a constant the field's own doc comment contradicts. `start_if_policy` returning `None` (`supervisor/mod.rs:91-100`) genuinely puts the user's command in the primary session, so `-u sandbox` was being ignored in the one place the machinery now exists to honour it.

The doc comment on the field is updated to say only a *supervised* primary session leaves it false, and why.

Still worth stating plainly: this path has no `@microvm` coverage, because every e2e run resolves a policy. The value is now structurally tied to supervision rather than asserted by a test, which is the best available pin without a way to drive an unsupervised boot from the harness.

jansav (approved):

PR #180, opened 2026-07-29
refactor(policy)!: adopt egress.http as the canonical route table

jansav:
## What

Migrates our policy representation — in memory, on disk, and on the wire — from `network.allowedRoutes` to `network.egress.http`, which is what the newly-pinned `lens-sandbox-core` rev `4cb1d97` treats as canonical.

Zero new user-facing capability. This is the prerequisite for raw-TCP egress.

## Why

In core's `client.rs::handle_policy` → `parse_egress_tables`: **if `network.egress` is a JSON object, `allowedRoutes` is never read at all** — even when `egress.http` is absent. `allowedRoutes` is read only when there is no `egress` key. So a future `egress.tcp` cannot coexist with `allowedRoutes`; every HTTP route has to be in `egress.http` in the same frame.

Core parses `egress.http` and `allowedRoutes` with the *same* function (`parse_proxy_routes`), so this migration is behaviour-preserving for routing.

## How

- `lns_policy::NetworkPolicy` replaces `allowed_routes: Vec<RouteRule>` with a **non-optional** `egress: Egress`. Non-optional makes `egress: null` — which core reads as "no egress block", silently falling back to `allowedRoutes` — unrepresentable, and turns every touch point into a compile error rather than a silent write to a dead table.
- New `pub struct Egress { pub http: Vec<RouteRule> }`: camelCase, `deny_unknown_fields`, always serializes `http`. **No `tcp` field yet** — deliberate, so a hand-authored `egress.tcp:` is rejected for free until the real thing lands.
- Back-compat on load via `#[serde(try_from = "…")]`: `allowedRoutes` alone folds into `egress.http`; `egress` alone is used as-is; **both present is a hard error** (`policy sets both 'allowedRoutes' and 'egress'; keep one`). Core would silently ignore the `allowedRoutes` half here, and for a local user file silent rule-dropping is worse than an error. No existing file can have both — `egress:` was a `deny_unknown_fields` parse error until now — so the strict path breaks nobody.
- Serialization always emits `egress`, never `allowedRoutes`.
- The deny-first + ceiling merge in `artifact/policy.rs` is now a generic `merge_rule_table<R>`, so a second table can be folded the same way later. This PR still merges exactly one table and its behaviour is pinned unchanged — all 13 pre-existing merge/guardrail tests pass without modification.
- Scaffolds (`DEFAULT_POLICY_YAML`, `lns sandbox init`) emit `egress:\n  http: []`.

## Compatibility — one-way door

`allowedRoutes:` is accepted on load as a deprecated alias, so existing policy files keep working. But **the first save rewrites them**: an approval you accept, `lns policy allow`, or `lns connector connect` writes the file back in the `egress.http:` shape, which a *previous* lns release cannot parse (`deny_unknown_fields`).

This affects two cases:

- **Downgrades** — rolling back to an older lns after any policy write.
- **Team repos** sharing a committed `lns-policy.yaml` across mixed lns versions.

Accepted tradeoff: no dialect-preserving save path. Keeping one would mean two on-disk shapes and two wire shapes alive indefinitely, and the flip would still happen the moment a TCP rule landed — just at a more surprising time. Documented in `docs/policy.md`.

## Tests

- **Layer 3 (lns-policy)** — legacy `allowedRoutes:` loads into `egress.http`; new `egress: {http: [...]}` loads; both keys present errors with the exact message; round-trip save emits `egress` and no `allowedRoutes`; an empty policy still writes `http: []`; an unknown key under `egress` (`tcp:`) is a parse error.
- **Layer 3 (lns-service)** — the `PolicyMessage` wire JSON carries `egress` as an object whose only key is `http`, no `allowedRoutes`, and still `defaultTransport`.
- **Layer 2 (lns-cli Gherkin)** — `lns policy allow/deny/list/remove` behave identically against a new-shape file, plus a new scenario pinning that `lns policy list` still reads a legacy-shape one.

## Gate

`make lint`, `make complexity`, `make coverage` all green; no new entries in the `scripts/coverage-floor.sh` IGNORES table.

### DIFF
--- a/crates/e2e-tests/tests/steps/microvm.rs
+++ b/crates/e2e-tests/tests/steps/microvm.rs
@@ -169,7 +169,7 @@ fn microvm_project(world: &mut E2eWorld) -> std::path::PathBuf {
     std::fs::write(root.join("lns.yaml"), definition).expect("write project lns.yaml");
     std::fs::write(
         root.join("lns-policy.yaml"),
-        "network:\n  allowedRoutes: []\n  defaultVerdict: ask\n  defaultTransport: direct\n",
+        "network:\n  egress:\n    http: []\n  defaultVerdict: ask\n  defaultTransport: direct\n",
     )
     .expect("write the project policy");
     root
@@ -1110,7 +1110,7 @@ fn output_lists_that_run(world: &mut E2eWorld) -> Result<(), String> {
 fn policy_ask_with_direct_route(world: &mut E2eWorld) {
     write_policy(
         world,
-        "network:\n  defaultVerdict: ask\n  defaultTransport: direct\n  allowedRoutes:\n    - match: api.example.test\n      verdict: allow\n      transport: direct\n",
+        "network:\n  defaultVerdict: ask\n  defaultTransport: direct\n  egress:\n    http:\n      - match: api.example.test\n        verdict: allow\n        transport: direct\n",
     );
 }
 
--- a/crates/lns-artifact/src/sandbox.rs
+++ b/crates/lns-artifact/src/sandbox.rs
@@ -479,7 +479,7 @@ mod tests {
     #[test]
     fn parse_reads_the_whole_flat_definition() {
         let json = def_json(
-            r#"{"image":"ghcr.io/team/base:1","command":"agent --serve","workdir":"/workspace","env":{"MODE":"research"},"resources":{"cpu":2,"memory":"1Gi"},"policy":{"defaultVerdict":"deny","allowedRoutes":[{"match":"api.example.test","verdict":"allow"}]},"connectors":["some-provider"],"credentials":[{"name":"some-provider","env":"SOME_TOKEN"}],"volumes":[{"type":"bind","source":".","target":"/workspace"},{"type":"volume","source":"home","target":"/root/.home","readOnly":true}],"ports":[{"container":8080}]}"#,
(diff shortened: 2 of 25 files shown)


chenhunghan (approved):
**Problem:** `lens-sandbox-core` at the pinned rev `4cb1d97` treats `network.egress.http` as the canonical HTTP route table — its `parse_egress_tables` stops reading `allowedRoutes` entirely once `egress` is a JSON object, even when `egress.http` is absent. Routes therefore have to move into `egress.http` before raw-TCP egress can land.

**Approach:** `NetworkPolicy.allowed_routes` becomes a non-optional `egress: Egress { http }` — camelCase, `deny_unknown_fields`, always serialized — so `egress: null` and a silent write to a dead table are both unrepresentable. `allowedRoutes` is removed outright rather than aliased, so a file still naming it fails the load with the key in the error instead of parsing to zero rules. The deny-first ceiling merge is extracted so a second table can fold the same way later.

**Verification:** I read `parse_egress_tables` at the pinned rev — the stated premise holds, including that an `egress` block supersedes `allowedRoutes` even with `http` absent, and that core parses both tables with the same `parse_proxy_routes`, making this behaviour-preserving for routing. The migration is complete: no `allowedRoutes` writer survives anywhere in the tree, `egress` is always emitted alongside `defaultTransport` on the wire, and the guest supervisor is embedded in `lns-service` so there is no cross-version wire skew to worry about. Gate is green.

**Verdict:** Ship it. ✅

One note, not a change request: the PR description's **How** and **Compatibility** sections describe a `try_from` back-compat shim, a "both keys present" error message, and tests that aren't in the diff. The commit message describes what actually landed and is accurate, and since this repo is rebase-merge-only that is what reaches `main` and the changelog — but the body is worth correcting so it doesn't read as promising back-compat that isn't there.

PR #182, opened 2026-07-29
feat(policy): per-binary scoping for network routes

jansav:
## What

Exposes `binaries` on network route rules — scoping a rule to the specific guest executables allowed to use it.

```bash
lns policy allow git.example.test --binary /usr/bin/git --binary /usr/bin/ssh
```

```yaml
network:
  defaultVerdict: ask
  egress:
    http:
      - match: git.example.test
        verdict: allow
        binaries:
          - /usr/bin/git
```

## Why

The pinned `lens-sandbox-core` rev `4cb1d97` supports per-binary scoping, but we never emitted the field — and because our `RouteRule` carries `deny_unknown_fields`, a user who hand-wrote `binaries:` in `lns-policy.yaml` got a parse error rather than the feature.

## Semantics — this fails closed, and that is the point

Straight from core, and now documented in `docs/policy.md`:

- A caller matches if **its own executable path or any ancestor's, up to the guest's init**, is on the list. Paths are compared against the kernel-resolved `/proc/<pid>/exe`, so they must be absolute and canonical — not a symlink or a `PATH` shim.
- The filter scopes the whole rule **regardless of verdict** and **fails closed**: a rule whose host matches but whose binary filter excludes the caller **denies** that caller rather than falling through to the default verdict.
- Once a binary-scoped rule claims a host, a later **unrestricted** allow or ask for the same host does **not** re-open it for the excluded caller — the excluded caller is denied, not prompted.
- It gates DNS too, not just the connection.

This is the one policy feature that denies without asking, which is what drives the ordering and refusal behaviour below.

## Ordering — a rule that never fires is worse than no rule

The gate stops at the first matching rule, so a scoped rule appended behind a rule that already covers the destination does nothing at all: the developer reads their file as restricting egress to one binary while it stays open to every caller. `lns policy allow` and `lns policy deny` now place a new rule ahead of any existing rule that covers the same destination, and say where it landed.

Coverage is decided the way core decides it — `NetworkPolicy::first_shadowing_rule`, in a new `lns-policy::matching` module checked line-by-line against core's `routing.rs`:

- first-match order and the scheme filter;
- a portless rule covers the same hosts pinned to a port, not the reverse;
- address ranges compared numerically via `ipnet` (the crate and version core itself uses): range containment, `cidr:port`, bracketed IPv6, and `::ffff:` mapped literals canonicalized as core canonicalizes the request address;
- a wildcard covers every pattern confined to its suffix, so a scoped rule behind `*` or `*.example.test` is detected;
- caller coverage follows core's own no-reopen guard, including that a later rule scoped to a subset still matches and a later deny is never suppressed.

Where the answer belongs to DNS rather than to the file — a hostname rule against an address range, one mid-segment wildcard against another — the rule is left where the developer put it and `docs/policy.md` says so. Guessing there would silently reorder a file the developer wrote deliberately.

Two placements would widen egress rather than fix order, and both are refused with nothing written:

- an `allow` behind a covering `deny` — ahead of that deny it opens every destination the deny covers;
- an unrestricted `allow` behind a binary-scoped rule — ahead of that rule it opens the destination to every caller in the sandbox, the exact opposite of what the scoped rule says. The error names the covering rule and its binaries, and suggests `lns policy remove <pattern>` only when that command would not also delete the rule being added.

A `deny` behind a deny, or a rule the file already holds, reports and changes nothing.

## Design decisions

**`--binary` doesn't exist on `deny`.** Because the filter fails closed, every caller reaching the host is denied either way — the listed binaries by verdict, the rest by the filter — so a scoped deny is behaviourally identical to a plain deny. Rather than accept and ignore it, the flag is only defined on `allow`, so `lns policy deny host --binary /usr/bin/git` exits 2 with clap's unknown-argument error. A hand-written scoped deny still *loads* (core accepts it); we only refuse to author one.

**Validation rejects what the guest could never match.** Core hard-rejects an empty `binaries: []` (matches no caller, turning an allow into an unconditional deny for the host) and any relative path (can never equal an absolute `/proc/<pid>/exe`); both would fail the **entire policy** closed inside the guest, so we reject them at load and authoring time with a message that says why. On top of those we refuse two more shapes that can never equal a kernel-resolved path either — one climbing through `..`, and one naming no binary — while accepting the redundant separators that *do* compare equal as paths (`/usr/bin/git/`, `//usr/bin/git`, `/usr/./bin/git`), so validation never rejects a scope that would have worked.

**Absent is not empty.** Core reads `binaries: []` as "matches no caller", so a rule open to everyone omits the key entirely (`skip_serializing_if`), and there's a test pinning that.

**Approvals and connector routes stay unscoped.** `RouteRule::allow_host`/`deny_host` (the approval write-back) and `ConnectorRoute::to_route_rule` set `binaries: None`. Catalog routes are connector-scoped by design.

**One host matcher.** lns-service's private `host_matches_pattern` copy is gone; the offer card and the connector-overlap check now use the same port of core's `domain_matches` this scoping needs. The two had drifted on which pattern shapes count as wildcards, and a second matcher deciding host coverage next to a fail-closed filter is not something to keep.

## Artifact merge

An allow now clears a deny-by-default ceiling when some ceiling rule *permits* it, rather than only on exact equality. So an artifact's binary-scoped ceiling can no longer be widened to every caller by an unscoped overlay allow, while a user narrowing an open ceiling to fewer callers survives the merge — and lands ahead of the ceiling's own allow, without which a first-match gate would make the narrowing void.

## Visibility

`lns policy list` gains a `BINARIES` column, and `--format json` reports the same under `binaries` (`null` when unscoped). This is required rather than cosmetic: `lns policy remove <pattern>` deletes **every** rule carrying the pattern, scoped or not, so a developer has to be able to see the scoping before removing anything. The clap help and the docs now both say "every rule".

## Compatibility

A file using `binaries` needs an lns new enough to understand it. Releases since **0.16** refuse to load it rather than silently dropping the scoping; anything older ignores the key and strips the scoping the next time it writes the file — so don't edit a scoped policy with one. (Verified: `deny_unknown_fields` on `RouteRule` first shipped in `lns-v0.16.0`.)

## Tests

- **Layer 3 (lns-policy)** — a scoped rule round-trips through YAML under core's wire name; an unscoped rule omits the key; `binaries: []`, a relative path, a `..` segment and a path naming no binary are each refused with a message explaining why, while path-equal separators are accepted; two rules differing only in `binaries` are not deduped by `add_rule` (they are genuinely different grants).
- **Layer 3 (lns-policy::matching)** — the shadow model against every pattern shape the gate accepts: wildcard subsumption and its near misses, host/port coverage in both directions, bare vs bracketed IPv6, CIDR containment and `cidr:port`, mapped literals, scheme-bound rules, and caller coverage including the subset and no-reopen cases.
- **Layer 3 (lns-service)** — merge guards on the artifact-ceiling escalation surface: an unrestricted overlay allow cannot clear a ceiling whose only allow for that host is binary-scoped; a scoped overlay allow cannot punch through a ceiling that doesn't carry it; a narrowing that a ceiling already permits survives *and* lands ahead of it; a ceiling's description is not treated as part of its grant. Each carries a positive assertion so none can pass on an empty merge.
- **Layer 2 (lns-cli Gherkin)** — scoping an allow records it; repeating the flag records both; a relative path and a `..` path each error and write nothing; `--binary` on `deny` exits 2; `list` shows the scoping in both the table and `--format json`; and the placement behaviour end to end — narrowing behind an open rule, behind a wildcard, behind a catch-all, a deny placed ahead of a scoped allow, a narrowing that must not claim the rule it passed is spent, and the two refusals (behind a deny, and re-opening a scoped destination) each leaving the file untouched.

## Gate

`make lint` and `make complexity` green; every file this PR touches is at 100% under `make coverage`, with no new entries in the `scripts/coverage-floor.sh` IGNORES table.

## Follow-up

Not in this PR: `lns policy allow/deny` still accepts destination patterns core's `parse_matcher` rejects (an unbracketed IPv6 literal, a `host:443:extra` typo). Inside the guest one such rule answers with `force_deny()` over the whole table — fail-closed, but total and invisible from the host. Fixing it means validating pattern shape for every rule on both subcommands plus `Policy::load_or_default`, which is not specific to binary scoping.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3331,6 +3331,7 @@ dependencies = [
 name = "lns-policy"
 version = "0.0.0"
 dependencies = [
+ "ipnet",
  "libc",
  "serde",
  "serde_json",
@@ -3428,6 +3429,7 @@ dependencies = [
  "async-trait",
  "lens-sandbox-core",
  "libc",
+ "lns-policy",
  "rustls",
  "serde_json",
  "serial_test",
--- a/crates/lns-artifact/src/sandbox.rs
+++ b/crates/lns-artifact/src/sandbox.rs
@@ -190,6 +190,10 @@ pub fn parse(config_json: &[u8]) -> Result<Definition> {
         .policy
         .validate_local_transport()
         .context("sandbox policy")?;
+    doc.spec
+        .policy
+        .validate_binary_scopes()
+        .context("sandbox policy")?;
     if let Some(workdir) = &doc.spec.workdir {
         spec::validate_mount_path(workdir).context("workdir")?;
     }
@@ -530,6 +534,42 @@ mod tests {
         assert!(
             format!("{err:#}").contains("upstream transport isn't supported in the local sandbox"),
             "got: {err:#}"
+        );
+    }
+
+    #[test]
+    fn parse_rejects_a_relative_binary_scope() {
+        let err = parse(&def_json(
+            r#"{"image":"ghcr.io/team/base:1","policy":{"egress":{"http":[{"match":"git.example.test","verdict":"allow","binaries":["git"]}]}}}"#,
+        ))
+        .unwrap_err();
+        assert!(
+            format!("{err:#}").contains("is not an absolute path"),
+            "a scope core would reject must fail here, not inside the guest: {err:#}"
+        );
+    }
+
+    #[test]
+    fn parse_rejects_an_empty_binary_scope() {
+        let err = parse(&def_json(
+            r#"{"image":"ghcr.io/team/base:1","policy":{"egress":{"http":[{"match":"git.example.test","verdict":"allow","binaries":[]}]}}}"#,
+        ))
+        .unwrap_err();
+        assert!(
+            format!("{err:#}").contains("matches no caller"),
+            "got: {err:#}"
+        );
+    }
+
+    #[test]
+    fn parse_accepts_an_absolute_binary_scope() {
+        let def = parse(&def_json(
+            r#"{"image":"ghcr.io/team/base:1","policy":{"egress":{"http":[{"match":"git.example.test","verdict":"allow","binaries":["/usr/bin/git"]}]}}}"#,
+        ))
+        .unwrap();
+        assert_eq!(
+            def.spec.policy.egress.http[0].binaries,
+            Some(vec!["/usr/bin/git".to_string()])
         );
     }
 
--- a/crates/lns-cli/src/policy/mod.rs
+++ b/crates/lns-cli/src/policy/mod.rs
@@ -17,13 +17,26 @@ pub struct PolicyArgs {
 #[derive(clap::Subcommand)]
 pub enum PolicyCommand {
     #[command(about = "Add an allow rule for a destination pattern.")]
-    Allow(PolicyRuleArgs),
+    Allow(PolicyAllowArgs),
     #[command(about = "Add a deny rule for a destination pattern.")]
     Deny(PolicyRuleArgs),
     #[command(about = "List the rules in the policy file.")]
     List(PolicyScopeArgs),
-    #[command(about = "Remove the rule matching a destination pattern.")]
+    #[command(about = "Remove every rule matching a destination pattern.")]
     Remove(PolicyRemoveArgs),
+}
+
+/// Only `allow` takes `--binary`: on a deny the listed callers are blocked by verdict and the rest fail closed, so all the scoping buys is that a later rule scoped to one of those others can still let it through — a distinction too fine to hand a flag.
+#[derive(clap::Args)]
+pub struct PolicyAllowArgs {
+    #[command(flatten)]
+    pub rule: PolicyRuleArgs,
+    #[arg(
+        long,
+        value_name = "PATH",
+        help = "Restrict the rule to callers running this guest binary, given as an absolute path (e.g. /usr/bin/git); repeatable. Every other caller is denied the destination rather than asked."
+    )]
+    pub binary: Vec<String>,
 }
 
 #[derive(clap::Args)]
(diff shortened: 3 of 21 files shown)


chenhunghan (approved):
**Problem:** Route rules never emitted core's `binaries` field, so per-binary scoping was unreachable — and `deny_unknown_fields` turned a hand-written `binaries:` into a parse error. Since the filter fails closed, a scoped rule appended behind a rule that already covers the destination reads as a restriction while changing nothing, so `lns policy` becomes order-aware.

**Approach:** Wire `binaries` through `RouteRule` with load- and authoring-time validation, port core's `domain_matches` into a new `lns-policy::matching` module (pinned to core by a parity test in the one crate depending on both), place a new rule ahead of any rule that shadows it, refuse the placements that would widen egress, and relax the artifact-ceiling check from exact equality to "some ceiling rule permits it".

**Product alignment:** Fits "policy you run into, not write" — the CLI does the ordering reasoning instead of leaving a dead rule in the file. Retiring lns-service's second host matcher next to a fail-closed filter is the right call.

I checked the port against core `4cb1d97`: `domain_matches`, `parse_binaries`, `caller_admits_rule`'s no-reopen guard, the ancestor depth limit of 8, and `PathBuf` component comparison making `/usr/bin/git/` equal. The docs match the gate's behaviour, and the artifact-merge guards hold — an unscoped overlay allow cannot widen a scoped ceiling, and a scoped overlay allow cannot punch through one. `cargo test -p lns-policy` and `-p lns-cli` pass in a clean worktree.

**Verdict: Ship it.** ✅

Two non-blocking notes, neither worth holding the PR for:

**The shadow model is port- and scheme-aware; core's DNS gate is not.** `caller_covers` is reached only after `port_covers`/`scheme_covers`, but `qname_matches_rule` matches a QNAME against a `HostPort` rule's host part alone and never consults `scheme`, then applies the same binary filter and no-reopen guard. So `allow db.example.test:5432 --binary /usr/bin/psql` followed by `allow db.example.test:6379` is appended with no warning, and in the guest a non-psql caller's lookup of that name returns `dns-binary-not-allowed` — the second rule never fires for it even though the TCP gate would admit it. Narrow: it needs two port-pinned rules on one name in the http table (the shape that's actually common belongs to `egress.tcp`, which `lns policy` doesn't write), or a hand-written `scheme:`. If it ever comes up, deciding the scoped-existing/unscoped-new case on host coverage alone turns it into the re-open refusal that already exists.

**Refusing an `allow` behind a covering `deny` also refuses the narrower carve-out.** `destination_covers` guarantees the new pattern is a subset of the deny's, so `deny "*.vendor.example"` then `allow api.vendor.example` fails with nothing written. Fail-closed is a defensible default and clearly deliberate here; noting only that the refused placement is narrower than the remedy the error suggests.

PR #183, opened 2026-07-29
feat(policy)!: raw TCP egress, and destinations decided by rules rather than a default

jansav:
## What

Two things, in three commits.

**Raw TCP egress** (`e0a8da8`) — a second egress table for destinations Lens Sandbox cannot read:

```yaml
network:
  egress:
    http: []
    tcp:
      - match: db.internal:5432
        verdict: allow
        binaries: [/usr/bin/psql]
        description: app database
```

```bash
lns policy allow-tcp db.internal:5432
lns policy deny-tcp  10.0.0.0/8:22
```

**Destinations are decided by rules, not a default verdict** (`e489cae`, breaking) — `defaultVerdict` and `defaultTransport` leave both policy file surfaces, and a rule's `verdict` is `allow` or `deny`. Being asked about a destination is the absence of a rule, so there is no third verdict and no `ask-tcp` verb.

Plus a core pin bump (`4d5cf1c`) for lens-sandbox-core#38, and the docs (`0926464`).

## Why

Not everything a workload connects to speaks HTTP. Postgres, Redis and SSH cannot be matched against `egress.http` rules at all, so before this the only options were to deny them or turn the sandbox off.

Driving it by hand against a real Postgres is what produced the second commit. Two things were wrong:

- **You had to declare a rule to be asked about one.** An unreadable connection was dropped silently, so allowing a database meant hand-writing an `egress.tcp` rule with `verdict: ask` for a destination nothing had told you about. Core now raises the approval card on its own; the card writes the rule.
- **"Always allow" denied the request that raised it.** Answering an approval *is* a policy write, and the reload it triggered voided the consent for the held connection. Remembering a decision read as denying it.

Once a card arrives without being declared, `verdict: ask` has nothing left to do — hence the removal, and with it the `defaultVerdict` that was the other half of the same idea.

## Migration

A file naming the removed keys is read by value:

- `defaultVerdict: ask` and `defaultTransport: direct` load and are dropped. They restate what the runtime now does unconditionally, and every file created before this carries them.
- `allow` and `deny` decided something, so they are refused with the rule to write instead: `lns policy deny '*'` (or `allow`), then delete the line.

`lns-service` states both values on the wire, so the guest is unaffected.

## Closing a directory

`defaultVerdict: deny` is replaced by a catch-all deny rule, which needs four things said out loud:

- earlier rules still decide what they name — the catch-all only answers for what is left;
- a later `lns policy allow <host>` goes **in front of** the catch-all rather than being refused, or a closed policy would be editable by hand alone (it raises no cards either);
- closing `egress.http` closes raw destinations too, so `egress.tcp` needs no counterpart;
- a near-catch-all — `0.0.0.0/0`, `*.com`, or a `"*"` narrowed by callers or requests — closes nothing.

## Worth reviewing closely

The artifact overlay merge, where a lockdown ceiling stopped being a field and became a rule. Five defects came out of review there, each fixed with a failing test first:

- a layer's catch-all was hoisted **ahead of its own allows**, denying what the ceiling permits;
- rules written **behind** a layer's own catch-all were resurrected — twice, once by the hoist and once by the ceiling check, the second letting an untrusted baseline widen a user's lockdown through a line the author's own gate never reaches;
- a catch-all **allow** vanished from the merge entirely — the shape the migration message prescribes;
- connector routes were spliced **behind** the closer, so a connected connector died silently in a closed project, with no card to say why;
- `lns policy allow <host>` was refused on every closed policy.

Also: `WireNetwork::seeded` derives the wire verdict from the rules, so a closed policy still says "do not ask" on the path the guest reads a default rather than the tables — the one an unclassifiable connection takes. Without it a locked-down artifact would have started prompting for the traffic it exists to refuse.

## Testing

`make lint`, `make complexity`, `cargo test --workspace`, and `make e2e` (69 scenarios) all green. The live path was driven end to end against a real Postgres on the host over vmnet: a card with no policy file present, `always allow` writing the rule, and the query returning rows.

Local `make coverage` reports four single-statement FAILs, three of them in `lns-ipc` / `lns-ocsf` which this branch does not touch — the known macOS source-mapping artifact. The Linux coverage job here is the arbiter.


crates/lns-policy/src/lib.rs
```diff
@@ -25,48 +27,116 @@ pub struct Policy {
...
+
+impl TryFrom<NetworkPolicyRaw> for NetworkPolicy {
+    type Error = String;
+
+    fn try_from(raw: NetworkPolicyRaw) -> Result<Self, Self::Error> {
+        match raw.default_verdict.as_deref() {
+            None | Some("ask") => {}
+            Some(verdict @ ("allow" | "deny")) => {
+                return Err(format!(
+                    "defaultVerdict is no longer part of a policy file, and `{verdict}` decided what a rule now has to say: end `egress.http` with a catch-all instead — `lns policy {verdict} '*'` — then delete the line. {raw}",
```
chenhunghan: The two steps are in an order the user cannot follow. `lns policy deny '*'` loads the policy file first, so it stops with this same error — the command this message prescribes cannot run until the line it tells you to delete afterwards is already gone. Verified against the built binary:

```
$ lns policy deny '*'
Error: loading policy from …/lns-policy.yaml
Caused by: defaultVerdict is no longer part of a policy file, and `deny` decided what a rule now has to say: end `egress.http` with a catch-all instead — `lns policy deny '*'` — then delete the line. …
```

Every locked-down project upgrading to this release hits that loop, from the one message meant to get it out. Swapping the order fixes it:

```rust
"defaultVerdict is no longer part of a policy file, and `{verdict}` decided what a rule \
 now has to say: delete the line, then end `egress.http` with a catch-all instead — \
 `lns policy {verdict} '*'`. {raw}",
```

The `{raw}` tail still reads correctly after the command sentence. Note that `a_default_verdict_that_decided_something_is_refused_with_the_rule_to_write_instead` passes under either order — it asserts the key, the verdict word and the tail phrase, not the sequence — so it needs an ordering assertion to hold the fix.

The other three exits are fine: the unknown-value and `defaultTransport` messages say "delete the line" with no command, and the `verdict: ask` message says "delete the rule". This branch is the only one that names a CLI verb.

crates/lns-cli/src/run/summary.rs
```diff
@@ -286,9 +281,13 @@ fn rules_line(policy: &Policy) -> String {
         .fold((0u32, 0u32), |(a, d), r| match r.verdict {
             Verdict::Allow => (a + 1, d),
             Verdict::Deny => (a, d + 1),
-            Verdict::Ask => (a, d),
         });
-    format!("{} allow, {} deny, anything else asks", allows, denies)
+    let tail = if policy.network.is_closed() {
```
chenhunghan, line 281: `rules_line` folds over `egress.http` only, so the launch summary reports nothing at all for the raw table:

```
$ lns policy allow-tcp '*:443'
$ lns policy allow-tcp '0.0.0.0/0:5432'
# lns-policy.yaml now holds two catch-all raw splices; the Policy block still says
#   rules: 0 allow, 0 deny, anything else asks
```

What makes this worth fixing rather than noting is that no other surface covers a directory's own policy. `guardrail_flags` runs only against a shipped artifact baseline (`artifact/real.rs:294`), and the raw count added to `lns sandbox inspect` (`raw_rule_note`) is a different command. So the launch summary is the only disclosure for the overlay, and it reads zero for what this PR itself calls the widest grant a policy can express.

A count in `rules_line`, or a sibling line in the same shape as `raw_rule_note`, closes it.

chenhunghan (commented):
**Problem:** A workload that speaks Postgres, Redis or SSH cannot be gated at all — `egress.http` rules cannot match traffic the proxy cannot read, so the only choices were to deny it or turn the sandbox off.

**Approach:** A second `egress.tcp` pre-filter table (port-mandatory, spliced unread, treatment-aware approval cards), plus removing both defaults and re-expressing "closed directory" as a catch-all deny the first-match gate actually reaches.

**Product alignment:** Fits "policy you run into, not write" — the raw card discovers the destination instead of requiring a hand-written `verdict: ask`, and the RAW badge and caption are honest about what approving buys.

The overlay merge — the part the description flagged — holds up: the `live_rules` truncation, the closer-last placement, deny-precedence among the reached closers, and the ceiling check all agree with the first-match gate, and the tcp fold is correctly independent of http.

**Verdict:** Needs fixes. Two items, both small.

crates/lns-policy/src/lib.rs
jansav: Fixed in 33bfdf6, in the order you gave — the loop was real, and `lns policy deny '*'` is the only command any of these four exits names, so it was the only one that could trap the reader.

```
defaultVerdict is no longer part of a policy file, and `deny` decided what a rule now has
to say: delete the line, then end `egress.http` with a catch-all instead — `lns policy
deny '*'`. That rule governs raw destinations too, so `egress.tcp` needs no counterpart
```

You were right that the existing test held nothing here: it asserts the key, the verdict word and the tail phrase, all order-blind. Added `the_refusal_says_delete_the_line_before_the_command_that_needs_it_gone`, which finds both substrings and asserts the positions, so the sequence is what fails if it is swapped back.

crates/lns-cli/src/run/summary.rs
jansav: Fixed in 33bfdf6. The argument that settled it was yours: `guardrail_flags` only ever sees a shipped baseline and `raw_rule_note` is `lns sandbox inspect`, so this line is the whole disclosure for the directory's own policy — and it read zero for the widest grant the file can hold.

Took the suffix rather than a count folded into the existing numbers, and it appears only when the raw table holds something:

```
rules: 1 allow, 0 deny, 1 raw allow, 0 raw deny, anything else asks
```

An http-only policy is byte-identical to before, so no existing summary changes and nobody reads two zeros for a table they never used. The empty-table early return now guards both tables, so a raw-only file reports its rules instead of "none defined" — that was the same blindness one step further along. Two tests, both watched fail first.

chenhunghan (approved):

PR #188, opened 2026-08-04
fix: give every workload guest a public trust store and a working npm

jansav:
Two guest-side regressions from a9e33be, which moved `docs/examples/claude-code/lns.yaml` from `node:lts-alpine` to `debian:bookworm-slim` + `spec.tools: [node@22]`. Running that example greets the user with:

```
Warning: Ignoring extra certs from `/etc/ssl/certs/ca-certificates.crt`, load failed: ... No such file or directory
Reshimming mise 22.23.2...
/.lens/tools/node/22.23.2/bin/npm: line 76: mise: command not found
```

## No public trust store in the workload guest

The supervisor hands every workload process `SSL_CERT_FILE` / `NODE_EXTRA_CA_CERTS` / `REQUESTS_CA_BUNDLE` / `CURL_CA_BUNDLE` / `GIT_SSL_CAINFO` naming `/etc/ssl/certs/ca-certificates.crt`, and the MITM proxy CA is appended to that same file — but nothing ever put a store there. We staged the pinned cacert PEM into the disposable provisioner guest only. An image that ships no `ca-certificates` (every slim base) left node warning on each launch, and the proxy CA append failed silently.

- The service stages the pinned store at `/.lens/ca-certificates.crt` on every run (`crates/lns-service/src/ca_bundle.rs`), reusing the `[ca_bundle]` pin and the same cache slot the provisioner already fills, so a warm machine performs no extra fetch.
- The broker installs it at the canonical path **only when the rootfs ships none** (`crates/lns-session-broker/src/trust.rs`), probing with `symlink_metadata`. An image's own store keeps every root it added — that is why the copy is guest-side rather than a runtime-layer overlay at the canonical path.
- Both sides read one constant (`lns_session::STAGED_CA_BUNDLE_PATH`).
- Best-effort throughout: a store the service cannot fetch warns and stages nothing, and a copy the broker cannot make reports and keeps booting.

## `mise: command not found` from npm

mise's core node plugin replaces node's stock `bin/npm` symlink with its own shim (`install_npm_shim` → `assets/node_npm_shim`, `#!/usr/bin/env bash`, runs `mise reshim` after global installs). We cached that tree verbatim, and a workload guest has neither `mise` (the engine lives only in the provisioner guest) nor necessarily bash. Ingest **and** lookup now restore the launcher to the symlink the upstream tarball ships, so trees already cached on users' machines are fixed without a manifest schema bump — which would have forced a network re-provision.

## Tests

Layer 3 throughout, since both changes sit below the IPC surface:

- `ca_bundle.rs` — staging path/mode/source, warm-cache reuse without a fetch, unreachable pin stages nothing, rotated bytes that miss the pin never reach a guest.
- `trust.rs` — seeded when absent, untouched when the image ships one (incl. a dangling symlink, pinned against the real FS in `trust/real.rs`), no-op when nothing was staged, and each failure reported.
- `tools/cache.rs` — shim restored at ingest, a tree cached with the shim restored on lookup, a launcher without the stock tree beside it left alone, an already-stock symlink unchanged.

`make lint`, `make complexity`, `make test` and `make coverage` are green.

## Follow-ups (not in this PR)

- lens-sandbox-core `ca.rs::write_ca_cert` opens the bundle `append(true)` with no `create(true)`. With the store seeded before the supervisor starts it succeeds today; a `create(true)` hardening PR + pin bump keeps a future ordering change from silently killing MITM again.
- An `@microvm` e2e scenario asserting the canonical path exists in a live guest.


crates/lns-service/src/run/orchestrator.rs
```diff
@@ -283,6 +283,10 @@ async fn orchestrate(
     if let Some(ensured) = &ensured_tools {
         fileset_specs.extend(ensured.specs.iter().cloned());
     }
+    fileset_specs.extend(
+        crate::tools::provisioner::real::workload_ca_spec(&cache_dir, crate::tools::host_arch())
```
chenhunghan: **Critical — unbounded network fetch now sits in front of every run.**

`workload_ca_spec` is called unconditionally, and `RealFetcher` uses bare `reqwest::get` with no timeout. On a cold cache behind a captive portal or a blackholed route, `lns run alpine` — a sandbox declaring no tools, on an image that already ships its own store — blocks indefinitely before the VM boots. Before this PR a cold-cache fetch only sat in front of runs that opted into `spec.tools`.

"Best-effort throughout" is the stated contract, but an unbounded fetch isn't best-effort — it's a hang, and it makes the `Err` → warn-and-stage-nothing arm unreachable.

The unconditional staging is the right call (the host can't cheaply know whether the image ships a store), so the fix is the bound, not the condition. Two notes on where to put it:

- Don't put a total timeout on `RealFetcher`. It's shared with the kernel/VMM/mise-engine downloads, and reqwest's request `.timeout()` covers the body stream — a 10s total there would break large transfers on a slow link.
- A `connect_timeout` on a shared client bounds the captive-portal case for every artifact without tripping on a slow-but-progressing transfer; apply the total bound only at this call, e.g. `tokio::time::timeout(..)` around `ensure_pem` inside `ca_bundle::workload_spec`, mapping elapsed onto the existing warn-and-`None` path. That keeps the bound where the best-effort contract is stated rather than in the shared port.

`update_check/real.rs:12` already uses `Duration::from_secs(10)` for exactly this reason, which is generous for a ~230KB PEM.

On timeout, the existing error arm is the correct behaviour — don't fail the run. Hard-erroring over a *public* trust store would break every image that ships its own. `ensure_pinned` is already self-healing, so no negative caching or backoff is needed; the next run retries once the network is back.

crates/lns-session-broker/src/trust.rs
```diff
@@ -0,0 +1,151 @@
...
+}
+
+/// Every workload process is handed `SSL_CERT_FILE` and friends naming the canonical bundle, and the proxy CA is appended to that same file — so a rootfs that ships none gets the staged store, and one that ships its own keeps every root it trusts.
+pub fn seed_trust_store_with(fs: &dyn TrustFs) -> Result<(), String> {
+    if fs.exists(SYSTEM_BUNDLE) {
+        return Ok(());
+    }
+    let pem = match fs.read(STAGED_PEM) {
+        Ok(pem) => pem,
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => return Ok(()),
```
chenhunghan: **The one state that leaves a guest with zero trust roots is silent.**

A `NotFound` on the staged PEM maps to `Ok(())`. So the combination "rootfs ships no store" *and* "nothing was staged" boots with no diagnostic at all — and that's the state where every HTTPS call in the workload fails, because the supervisor's proxy CA append has nothing to append to.

The broker is the only place both facts are known; lns-service can't say it, because it doesn't know what the image's rootfs contains when it stages. This arm should warn naming both facts rather than return `Ok`.

This matters independently of the timeout, and a timeout makes it reachable on a cold cache rather than only on a rotated pin.

crates/lns-session-broker/src/trust.rs
```diff
@@ -0,0 +1,151 @@
+#![cfg_attr(not(target_os = "linux"), allow(dead_code))]
+
+mod real;
+#[cfg(target_os = "linux")]
+pub use real::seed_trust_store;
+
+use lns_session::STAGED_CA_BUNDLE_PATH as STAGED_PEM;
+
+pub const CERTS_DIR: &str = "/etc/ssl/certs";
+pub const SYSTEM_BUNDLE: &str = "/etc/ssl/certs/ca-certificates.crt";
```
chenhunghan: **The canonical bundle path is spelled three times across the two crates that have to agree.**

`STAGED_CA_BUNDLE_PATH` correctly went into `lns-session` because "one spelling, or the guest silently keeps no store at all". The *canonical* path — the one that actually has to match — is duplicated: `trust::SYSTEM_BUNDLE` here, `provisioner::CA_BUNDLE` in lns-service, and a bare literal at `tools/mise.rs:154`.

This is the failure mode the PR is fixing. If the broker seeds one path and the workload env names another, TLS silently falls back to no trust store — no error, the same warning spam the user just reported. It belongs beside `STAGED_CA_BUNDLE_PATH` in `lns-session`, with lns-service's two sites pointing at it.

crates/lns-session-broker/src/trust.rs
```diff
@@ -0,0 +1,151 @@
...
+
+pub trait TrustFs {
+    fn exists(&self, path: &str) -> bool;
+    fn read(&self, path: &str) -> std::io::Result<Vec<u8>>;
+    fn create_dir_all(&self, path: &str) -> std::io::Result<()>;
+    fn write(&self, path: &str, contents: &[u8], mode: u32) -> std::io::Result<()>;
+}
+
+/// Every workload process is handed `SSL_CERT_FILE` and friends naming the canonical bundle, and the proxy CA is appended to that same file — so a rootfs that ships none gets the staged store, and one that ships its own keeps every root it trusts.
```
chenhunghan, line 36: **Nothing asserts the guest-side result.**

Layer 3 pins the copy-when-absent decision against a fake FS and the mode against a tempdir, but no test observes a real guest. The `@microvm` scenario is listed as a follow-up.

Both bugs in this PR shipped precisely because no test looked inside a booted workload. A `make e2e-microvm` scenario asserting the canonical path exists and `npm --version` works is the only thing that stops the next base-image or engine bump from re-breaking this the same silent way.

chenhunghan (commented):
**Problem:** Two guest-side regressions from moving the example to `debian:bookworm-slim` + `spec.tools` — slim images ship no public trust store at the canonical path every workload env var names (so the MITM proxy CA append also failed silently), and mise's `bin/npm` shim got cached verbatim into trees that have neither `mise` nor bash.

**Approach:** Stage the already-pinned cacert PEM at a shared `/.lens/ca-certificates.crt` on every run and have the broker copy it to `/etc/ssl/certs/ca-certificates.crt` only when the rootfs ships nothing there (`symlink_metadata` probe, so an image's own store keeps its roots). Separately, rewrite `bin/npm` back to the stock relative symlink at both ingest and lookup, so already-cached trees heal without a manifest schema bump.

**Product alignment:** Both fixes serve "no system dependencies" — the guest gets a working trust store and npm regardless of what the base image ships. The lookup-side heal avoids forcing a network re-provision, which matters for "ephemeral by default".

**Verdict:** Needs fixes. One critical (an unbounded pre-boot fetch now sits in front of every run), plus three that will cause real maintenance pain.

crates/lns-service/src/run/orchestrator.rs
jansav: Fixed in 011ee4e7, following your split exactly.

The total bound sits at the call site in `ca_bundle::workload_spec` — a 10s `tokio::time::timeout` around `ensure_pem`, with elapsed mapping onto the existing warn-and-`None` arm. The shared `RealFetcher` gets a `connect_timeout` only, on a `OnceLock` client, so the kernel and VMM transfers keep their unbounded body streams.

Two paused-clock tests pin it: a fetcher that never resolves stages nothing rather than hanging, and a warm cache returns without waiting on the budget. The second one matters because the call is unconditional now — the budget must never cost a run that already has the bytes.

Also worth flagging as out of scope here: `guest_tools/real.rs:12` still uses bare `reqwest::get`, so a cold guest-tools cache behind the same blackholed route hangs the same prepare join. It is preexisting and a hard-failure path rather than best-effort, so I left it; happy to file it separately.

crates/lns-session-broker/src/trust.rs
jansav: Fixed in a8920380.

`seed_trust_store_with` now returns a `Seeding` outcome instead of unit, and the broker prints the diagnostic for the `NoRootsAtAll` arm only, naming both facts: no store at the canonical path and none staged. The two outcomes that leave the guest with roots stay silent.

I went with an outcome enum rather than `Err` because the arm is not a failure of seeding — nothing went wrong, there was simply nothing to copy — and the caller already prefixes `Err` with "best-effort trust store setup failed", which would have been the wrong sentence.

Three tests pin it: the report names both paths, the two healthy outcomes report nothing, and each of the healthy arms now asserts its variant so a regression that started warning on a good boot would fail.

crates/lns-session-broker/src/trust.rs
jansav: Fixed in 49a64c70. The path is now `lns_session::SYSTEM_CA_BUNDLE_PATH`, beside `STAGED_CA_BUNDLE_PATH` as you suggested, and all three sites point at it: `trust.rs`, `provisioner::CA_BUNDLE`, and the literal in `mise::provision_env`.

One duplicate survives on purpose: lens-sandbox-core spells it again in `ca_env.rs` and `ca.rs`. That is a separately published crate and cannot depend on `lns-session`, so it stays. It is also the append side of the same contract, so it is worth keeping in view — the guest-side e2e scenario below is what would catch a drift between the two crates.

crates/lns-session-broker/src/trust.rs
jansav: Fixed in fbd64e0c — agreed, this is the finding that stops the next bump from re-breaking it the same silent way.

New `microvm_trust_store.feature` pins both halves of the seeding decision. It boots busybox rather than the default alpine base, because alpine ships `ca-certificates-bundle` and would never reach the seeding branch at all.

`microvm_tools.feature` gets an `npm --version` scenario. The existing coverage only ran `node --version`, which never touches the rewritten launcher — which is exactly why that bug shipped.

Caveat worth stating plainly: these are `@microvm`-tagged, so they are excluded from the required CI jobs and I have not run them against a real guest yet. They need `make e2e-microvm` on a virt-capable host before merge.

jansav:
Thanks — all four findings are addressed and pushed. Reply on each thread; summary here:

| Finding | Commit |
|---|---|
| Unbounded fetch in front of every run | `011ee4e7` |
| Zero-roots state is silent | `a8920380` |
| Canonical path spelled three times | `49a64c70` |
| Nothing asserts the guest-side result | `fbd64e0c` |

Two changes you have not seen, both from a self-review before yours landed:

- `a978fa2a` — the probe you describe as `symlink_metadata` no longer is. A link with nothing behind it was counting as "the image ships a store", so seeding was skipped and the proxy CA append then failed too, leaving the guest with no roots. It uses `metadata` now, and seeding writes through a resolving link to the target the image chose.
- `05d40172` — the CA warm-up moved into the concurrent prepare `tokio::join!`. Note this did **not** address your critical finding on its own: a hang inside the join still stalls the whole prepare. The bound in `011ee4e7` is what fixes it.

Still open before merge: `make e2e-microvm` has not been run against a real guest, so the two new scenarios are unproven.

chenhunghan (approved):

PR #192, opened 2026-08-05
fix(tools): Take a tool's whole install, not the dir the engine names

jansav:
## Summary

`spec.tools` treated a declared tool as the one directory `mise where` reports. That only holds for a backend whose install is self-contained. mise installs rust through **rustup**, which writes the compiler to `$RUSTUP_HOME` and the proxies to `$CARGO_HOME/bin` — so `mise where rust@1.95.0` is merely a symlink to the latter. Every bin dir the engine reported therefore fell outside the tarred tree, the driver's filter dropped all of them, and the launch failed:

```
✗  provisioning rust@1.95.0 failed: mise rust@1.95.0 [1/3] 1.95.0-aarch64-unknown-linux-gnu installed …
   Nothing was cached; the next run retries from a clean state
```

The tree it *would* have cached held rustup's proxies with the toolchain left behind in a guest that is then destroyed, so rust never worked here — only the failure is new (it arrived with #, which replaced a bin-dir guess with the engine's own answer). rust is just the first declared tool of that shape; npm-, GOROOT-, and python-prefix-shaped backends can all break the same way, so the fix is to drop the containment assumption rather than to teach the driver about rustup.

### 1. The install's own prefix is the tree

Each tool now installs under a prefix of its own, with `HOME` and the engine's data dir inside it, and that prefix is what gets tarred. Whatever the backend writes, wherever it writes it, is captured. `mise where` is no longer consulted at all.

### 2. The prefix is built where the workload will read it

An install may bake its own absolute paths (rustup does, into its toolchain state), so the provisioner guest builds the tree at `/.lens/tools/<name>/<resolved>` — the same path the workload guest sees. That requires knowing the resolved version up front, so `mise latest` runs before the install.

### 3. A tool's environment travels with its tree

A tool whose binaries are launchers needs vars to find its real payload, and the engine is the only thing that knows which. `mise env --json <tool>@<resolved>` supplies them:

| Tool | Reported env |
|------|--------------|
| `jq@1`, `node@22`, `gh@2` | *(nothing beyond `PATH`)* |
| `rust@1.95.0` | `RUSTUP_HOME`, `CARGO_HOME`, `RUSTUP_TOOLCHAIN` |

`PATH` is dropped — the bin dirs own it. A value inside the tree travels tree-relative and is resolved against the injection root; a value naming anything else fails the provision, since it would point at a path only the disposable guest had. The vars reach the workload **and** every later `lns exec`, and fill only what the image and `-e` left unset, so a sandbox that sets its own `CARGO_HOME` keeps it.

### 4. Fail-closed handling of what the guest reports

The provisioner guest is disposable but its output is untrusted input:

- Var names must match the POSIX shape, and `PATH`, `HOME`, `LD_*`, `BASH_ENV`, `ENV`, `IFS` are refused — a loader hook would reach every process in the workload, not just the tool's own binaries.
- One `LNS_ENV` marker per tool, as with `LNS_TOOL`: a sibling's install code can forge a marker but not suppress the genuine one, so two is an engine fault rather than something to pick a winner from.
- The marker is emitted with `printf`, not `echo`: dash and busybox expand `\n` inside the engine's JSON, which would split the marker line and put attacker-influenced text on the marker channel.
- An absolute symlink pointing *inside* the installed tree is respelled relative at ingest, so injection keeps creating in-tree links only. Anything else absolute is still refused, including a sibling prefix (`…/1.2.3-evil`) that shares the root's spelling.

### 5. The workload owns the directories of its own tool trees

A tool env var may name a directory the tool *writes*, not only one it reads — the engine puts cargo's `CARGO_HOME` inside the tree. Every injected entry lands uid 0, and a workload whose image sets no `USER` runs as 65534, so `cargo build` on any project with a dependency failed on the registry it had to create.

Classing the variables would mean a list of names per backend (`GOMODCACHE`, npm's cache, …), and a missing name stays silent until someone runs a real build. The mode of the tree is the actual problem, so `lns-init` fixes that: once it has resolved the workload user from the guest's `/etc/passwd`, it hands that user every **directory** under `/.lens/tools`. Files stay root-owned, so no tool binary is edited in place, and the overlay's upper layer takes the writes and drops them at teardown.

| | `CARGO_HOME` | `cargo build` with one dependency |
|---|---|---|
| before | `drwxr-xr-x root root` | `failed to create directory …/registry/cache/…: Permission denied`, exit 101 |
| after | `drwxr-xr-x sandbox sandbox` | builds, exit 0 |

### 6. `lns exec` no longer contradicts its run

`run_workload_env` filtered the tool vars against the composed image+`-e` env, but `lns exec` composes against its own, nearly always empty, env. A definition setting `CARGO_HOME=/opt/cargo` therefore got `/opt/cargo` in the run and the tool tree's path in an exec session — one sandbox, two registries, no warning. `ToolRuntime::as_applied` keeps only the vars that took effect on the workload, and that narrowed set is what an exec session re-applies. A var the definition owns is now *unset* in exec rather than wrong, which matches exec carrying none of the run's other env.

### 7. npm stops being a special case

npm was the one tool with production code of its own, and it is not even a declared tool: it arrives inside node's tree, where the engine replaces node's stock `bin/npm` symlink with a script that re-shims through `mise` under `bash`. `MISE_NODE_NPM_SHIM=0` stops the rewrite, so the tree keeps the symlink the upstream tarball ships — byte-for-byte what the repair reconstructed. The repair, its constants, both call sites, and its tests are gone: 204 lines out, 6 in. The general rule stays — the driver still deletes the engine's shim dir, and a reported bin dir under it still fails the provision.

## Breaking changes

Manifest schema 3 → 4. Tool trees cached under the old containment rule are a miss and re-provision once, rather than being reinterpreted under today's semantics.

### Before

```
✗  provisioning rust@1.95.0 failed: … Nothing was cached; the next run retries from a clean state
```

### After

```
✓ tools     rust@1.95.0 → 1.95.0 (static.rust-lang.org)
$ cargo build   # a project with one dependency, inside the sandbox
CARGO_BUILD_OK
```

## Test instructions

1. Declare `rust@1.95.0` in a `lns.yaml` over a glibc base image and run `rustc --version`. The tool provisions and prints `rustc 1.95.0`; before this change the launch was refused.
2. `lns exec <run> -- rustc --version` in the same run prints the same version — the tool's env reaches an exec session, not just the workload.
3. Set `CARGO_HOME=/opt/cargo` in the definition's `env` and run `sh -c 'echo $CARGO_HOME'`. The sandbox's own value wins over the one the tool reports.
4. Declare `node@22` and run `npm --version`. It prints a version rather than failing — the engine's mise-and-bash shim is replaced by the stock launcher even though node's layout now sits below the prefix.
5. Run the same definition twice. The second run provisions nothing (`✓ tools` reports the cached tree) and starts without network for the tool bodies.
6. Declare a musl base image with `rust@1.95.0`. rustup publishes no musl host toolchain, so this is expected to fail; the failure names the tool and its cause.
7. In a sandbox declaring `rust@1.95.0`, run `cargo build` on a project with a dependency. It succeeds — before this change it failed with `Permission denied` on `$CARGO_HOME/registry`.
8. Inspect a provisioned node tree: `bin/npm` is a symlink to `../lib/node_modules/npm/bin/npm-cli.js`, not a bash script, and `npm --version` runs in the guest.

## Verification

- `make lint`, `make complexity`, `make coverage` (100% per-file on `lns-service` and `lns-artifact`) — green.
- `cargo test -p lns-service`: 2083 unit tests, 188 scenarios — green.
- `make e2e-microvm`: the whole declared-tools feature passes against a real guest — the cargo build above, the npm launcher with the repair deleted, the nested `gh` bin dir, and both pulled-offline scenarios.
- The tool-tree ownership fix was verified in both directions: the cargo scenario goes red with `chown_tool_dirs` disabled and green with it. An earlier version of that scenario was vacuous — it asserted a literal the run summary's own command echo contained — so the marker is now assembled at run time and the step also requires exit code 0 and no `Permission denied`.
- Six failures elsewhere in that suite (2 credential, 2 trust-store, 2 registry-pull) are unrelated to this change by their own error messages (a malformed local grants sidecar; a base image with no `/bin/sh`; registry rate limits), but they are **not** baselined against `origin/main` yet, so treat them as unverified rather than known-good.

### DIFF
--- a/crates/lns-service/src/run/orchestrator.rs
+++ b/crates/lns-service/src/run/orchestrator.rs
@@ -453,6 +453,8 @@ async fn orchestrate(
             }))
             .await;
     }
+    // What an `lns exec` into this run may re-apply: a var the definition already owns kept its own value here, and exec composes against an empty env, so only what took effect travels.
+    let applied_tools = tool_runtime.as_applied(&composed.env);
     let env: Vec<String> = composed.env;
 
     let params = vm::session_client::SessionParams {
@@ -500,7 +502,11 @@ async fn orchestrate(
         "microVM   ({:.2}s)",
         boot_start.elapsed().as_secs_f64()
     );
-    crate::run_registry::set_connector_with_tools(&run_id, connector.clone(), tool_runtime.clone());
+    crate::run_registry::set_connector_with_tools(
+        &run_id,
+        connector.clone(),
+        applied_tools.clone(),
+    );
     let _vm_stop_guard = vm::VmStopGuard::new(connector.clone());
 
     log::progress("Connecting", "session", 0, 0);
--- a/crates/lns-service/src/run/orchestrator.rs
+++ b/crates/lns-service/src/run/orchestrator.rs
@@ -422,6 +422,13 @@ async fn orchestrate(
         args.workdir.as_deref(),
         crate::workload_cwd::image_workdir(image.config.as_ref()).as_deref(),
     );
+    let tool_runtime = ensured_tools
+        .as_ref()
+        .map(|ensured| crate::workload_env::ToolRuntime {
+            bin_paths: ensured.bin_paths.clone(),
+            env: ensured.env.clone(),
+        })
+        .unwrap_or_default();
     let composed = exec_env_strings(
         image.config.as_ref(),
         args.entrypoint.as_deref(),
@@ -434,10 +441,7 @@ async fn orchestrate(
                 .map(|s| s.managed_env_vars.as_slice())
                 .unwrap_or(&[]),
             workdir: workdir.as_deref(),
-            tool_bin_paths: ensured_tools
-                .as_ref()
-                .map(|ensured| ensured.bin_paths.as_slice())
-                .unwrap_or_default(),
+            tools: &tool_runtime,
         },
     );
     for refused in &composed.refused {
@@ -496,14 +500,7 @@ async fn orchestrate(
         "microVM   ({:.2}s)",
         boot_start.elapsed().as_secs_f64()
     );
-    crate::run_registry::set_connector_with_tool_bin_paths(
-        &run_id,
-        connector.clone(),
-        ensured_tools
-            .as_ref()
-            .map(|ensured| ensured.bin_paths.clone())
-            .unwrap_or_default(),
-    );
+    crate::run_registry::set_connector_with_tools(&run_id, connector.clone(), tool_runtime.clone());
     let _vm_stop_guard = vm::VmStopGuard::new(connector.clone());
 
     log::progress("Connecting", "session", 0, 0);
--- a/crates/lns-service/src/tools/provisioner/mod.rs
+++ b/crates/lns-service/src/tools/provisioner/mod.rs
@@ -5,6 +5,7 @@ use std::path::{Path, PathBuf};
 
 use anyhow::{Context, Result, bail};
 
+use super::cache::{ToolEnvValue, ToolEnvVar};
 use super::{
     Libc, ProvisionError, ProvisionTarget, SafeVersion, StagedTar, StagedTool, ToolRef, mise,
 };
@@ -59,8 +60,9 @@ pub(crate) fn provisioner_runtime_specs(
     specs
 }
 
-/// One shell driver per provision: each tool installs under its own engine state so no install resolves a version a sibling left behind, tars its tree into the staging share, and emits one `LNS_TOOL <name> <resolved> <binpaths>` marker whose bin dirs are colon-joined and relative to the tree root; any failure names its tool with `LNS_FAIL` and stops.
+/// One shell driver per provision: each tool resolves its version, installs into a prefix of its own at the path the workload will read it from, tars that whole prefix into the staging share, and emits an `LNS_TOOL <name> <resolved> <binpaths>` marker plus the `LNS_ENV <name> <json>` the engine reports for it; any failure names its tool with `LNS_FAIL` and stops.
 pub(crate) fn render_driver(requests: &[ToolRef]) -> String {
+    let root = crate::tools::cache::TOOLS_ROOT;
     let mut script = String::from(
         "#!/bin/sh\nset -u\nexport PATH=/.lens/tools-engine/bin:$PATH\nmkdir -p /tmp/mise/home\n",
     );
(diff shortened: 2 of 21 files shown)


crates/lns-service/src/tools/provisioner/mod.rs
```diff
@@ -91,6 +103,58 @@ pub(crate) struct DriverResult {
...
+        if var == "PATH" {
+            continue;
+        }
+        if !lns_artifact::tools::is_safe_env_name(&var) {
+            return Err(ProvisionError::Engine(format!(
+                "the provisioner driver reported an unusable env var for {name}: {var:?}"
+            )));
+        }
+        // A sibling prefix (`<root>-evil/bin`) shares the root's spelling without being inside it, so only the root itself or a path below it counts as in-tree.
+        let inside = value
```
chenhunghan, line 132: This test shows if a value is inside the tree or outside the tree. It does not show one more difference. Some variables give a directory that the tool only reads. `RUSTUP_HOME` and `RUSTUP_TOOLCHAIN` are these variables. Other variables give a directory that the tool writes. `CARGO_HOME` is one of these variables.

The driver sets `HOME="$root/home"`. Therefore `CARGO_HOME` becomes `/.lens/tools/rust/1.95.0/home/.cargo`. This path is inside the tree.

The service injects each entry with uid 0 and mode 0755. If the image gives no `USER`, the workload runs as uid 65534. `resolve_run_as` in `vm/mod.rs` does this. The `debian:12-slim` image gives no `USER`.

Then cargo cannot make `.package-cache`. It cannot make the registry index. It cannot make `.global-cache`. Therefore `cargo build` fails with a permission error. This failure occurs for each project that has one dependency or more. A mounted workspace does not correct the failure, because only `target/` is in the workspace.

rust is not a special tool here. Each backend that writes into its own prefix can fail in the same way. Go can report `GOPATH`, `GOMODCACHE`, or `GOCACHE`. npm can report a cache directory or a prefix. Therefore do not correct this problem with a list of variable names. A missing name in such a list gives no error message. The user finds the fault only when the user runs a real build.

Make one general rule instead. The problem is the mode of the tree, and not the variables. Give the workload write permission on the tool tree. Then no variable needs a class, and each future tool works also. The guest overlay already gives copy-on-write. Therefore each write goes to the upper layer and disappears at teardown. This agrees with the principle "ephemeral by default".

The host cannot always do this, because the host knows the uid only for the default `sandbox` user. An image can give a user name. The guest reads `/etc/passwd`, so the guest resolves the name to a uid. Therefore the guest changes the owner or the mode of `/.lens/tools` after it resolves the workload user. Only the directories need write permission. The files stay read-only.

Accept one trade-off. The workload can then replace its own tool binaries. But the change stays in the overlay upper layer of that run. It does not reach the cache, and it does not reach a different run.

One temporary step is available for rust only. Do not send `CARGO_HOME`. `PATH` already has the proxy directory. `RUSTUP_HOME` and `RUSTUP_TOOLCHAIN` find the compiler. Then cargo uses `$HOME/.cargo`, and the workload can write to that directory. But this step does not help the next tool.

Correct the test also. Line 49 of `crates/e2e-tests/features/microvm_tools.feature` runs `rustc --version`. That command passes with the incorrect design also. Use `cargo build` on a project that has one dependency.

crates/lns-service/src/run/orchestrator.rs
```diff
@@ -422,6 +422,13 @@ async fn orchestrate(
         args.workdir.as_deref(),
         crate::workload_cwd::image_workdir(image.config.as_ref()).as_deref(),
     );
+    let tool_runtime = ensured_tools
```
chenhunghan, line 425: The engine gives this set of variables. The code removes nothing from the set. Line 503 sends the same set to `set_connector_with_tools` for `lns exec`.

`compose_guest_tool_env` has a check for a variable that is already set. That check reads the environment of the workload. It does not read `tool_runtime`. `build_session_params` adds the tool variables to `args.env`. `args.env` holds only the `-e` options of the exec command, and it is usually empty. Therefore nothing stops the tool value there.

Example: the definition sets `CARGO_HOME=/opt/cargo`. Then `lns run ... -- cargo build` uses `/opt/cargo`. But `lns exec <run> -- cargo build` uses the tool tree. One sandbox then has two registries. The user gets no warning.

The comment on `compose_guest_tool_env` says that exec agrees with the workload. This is true for `PATH`. It is not true for these variables.

`composed.env` is ready seven lines after line 425. Use it. Keep only the tool variables that are in `composed.env`.

chenhunghan (commented):
**Problem:** `spec.tools` used one rule. A tool keeps all of its files in the directory that `mise where` gives. rustup does not obey the rule. rustup writes the compiler to `$RUSTUP_HOME`. It writes the proxies to `$CARGO_HOME/bin`. Therefore all bin directories were outside the tar file, and the launch failed.

**Approach:** Each tool now installs into its own prefix. The driver builds the prefix at the path that the workload reads later. `HOME` and the engine data directory are inside the prefix. The driver puts the full prefix into the tar file. `mise env --json` gives the variables that a launcher needs. The manifest schema changes from 3 to 4, so old trees do not match.

The tar part of the change is correct. The guards refuse a link that goes out of the tree. They also refuse a sibling prefix. The marker channel fails closed.

Two problems remain. Both problems are in the environment part.

**Verdict:** Needs fixes.

crates/lns-service/src/tools/provisioner/mod.rs
jansav: You are right on every point, including the one about my test.

Fixed the general way you describe, in `lns-init`: once `resolve_workload_ids` has resolved the user from the guest's `/etc/passwd`, `chown_tool_dirs` walks `/.lens/tools` and hands the resolved workload every **directory**. Files stay root-owned, so no tool binary is edited in place, and the writes land in the overlay's upper layer and disappear at teardown. It sits next to the existing `chown_fileset_owned` call and shares one logged-lchown helper with it. No variable is classified, so go's `GOMODCACHE` and npm's cache need nothing added.

Measured on a live guest, uid 65534, `debian`-based rootfs:

| | `CARGO_HOME` | `cargo build` (one dependency) |
|---|---|---|
| before | `drwxr-xr-x root root` | `failed to create directory .../registry/cache/...: Permission denied`, exit 101 |
| after | `drwxr-xr-x sandbox sandbox` | builds, exit 0 |

On the test: my first replacement was worse than `rustc --version`, not better. It asserted the captured stdout contained `CARGO_BUILD_OK`, and the run summary echoes the command it launches — so the assertion matched its own command text and passed with the guest fix disabled. It now emits the marker as `printf 'CARGO_BUILD_%s\n' OK`, and the step also requires exit code 0 and the absence of `Permission denied`. I verified it goes red without the fix before trusting it green with it. The scenario builds a library target with one dependency (`unicode-ident`, no build script) so an rlib needs no linker the slim base image lacks, and it takes its base image from the engine manifest's own gnu rootfs, which every tools run already fetches.

I did not take the temporary step of dropping `CARGO_HOME`, for the reason you give: it does not help the next tool. (commit 1b91fadb)

crates/lns-service/src/run/orchestrator.rs
jansav: Confirmed and fixed. `ToolRuntime::as_applied` keeps only the tool vars that actually took effect on `composed.env`, and the orchestrator registers that narrowed set for `lns exec` — so the case you describe (definition sets `CARGO_HOME=/opt/cargo`, run uses `/opt/cargo`, exec used the tool tree) cannot happen.

One consequence worth naming: a var the definition owns is now **unset** in an exec session rather than carrying the definition's value. That matches what exec already does with the rest of the run's env — it carries none of it — so exec never contradicts the workload, it just knows less. If you would rather exec inherited those values too, that is a larger change to what an exec session's environment is, and I would rather do it deliberately than as a side effect here.

Pinned by a Layer 3 test that composes both sides: with the image setting the var, `as_applied` drops it, and the exec env composed from the result carries neither the tool's value nor the var at all. (commit 1b91fadb)

chenhunghan (approved):

jansav:
Re-ran `make e2e-microvm` on the combined branch (all three commits), which closes the gap I noted when the npm commit was still stacked separately: the whole declared-tools feature passes against a real guest — `npm --version` with the repair code deleted, and `cargo build` with one dependency inside the tool tree's `CARGO_HOME`.

83 scenarios, 77 passed, 6 failed. The six are the same set as before this branch, in the credential, trust-store, and registry-pull features, and their messages name causes outside this diff (a malformed local grants sidecar, a base image with no `/bin/sh`, and registry rate limits). I have not baselined them against `origin/main`, so I am reporting them as unverified rather than pre-existing.

PR #193, opened 2026-08-06
refactor(tools): Stop the engine from shimming npm instead of repairing it

jansav:
## Summary

Stacked on #192 — review that one first.

npm is the only tool with production code of its own, and it is not even a declared tool. It arrives inside node's tree, where mise replaces node's stock `bin/npm` symlink with a script that re-shims through `mise` under `bash`. A workload guest has neither, so `restore_stock_npm` in the tool cache put the upstream symlink back at ingest and again on cache read.

That is repairing damage we can simply decline. The engine has a switch for the rewrite, verified against the pinned engine (mise 2026.7.14):

| Setting | `bin/npm` in the tree |
|---|---|
| default (`node.npm_shim = true`) | a 1831-byte `#!/usr/bin/env bash` script that re-shims through mise |
| `MISE_NODE_NPM_SHIM=0` | `../lib/node_modules/npm/bin/npm-cli.js` — the stock symlink |

The second row is byte-for-byte what the repair reconstructed, so one line in `mise::provision_env()` replaces the whole mechanism: the function, its three path constants, both call sites, its five tests, and the test helpers only it used. 204 lines out, 6 in.

### What stays

The general rule this special case sat beside is untouched: nothing in a tool tree may reference the engine. The driver still deletes the engine's shim dir (whose entries are absolute symlinks to the engine binary), and a reported bin dir under that dir still fails the provision.

### Cache safety

No second schema bump. Lookup already rejects any manifest whose `schema_version` is not 4, #192 introduced 4, and every build that writes 4 repaired at ingest — so the lookup-time repair this removes could not fire for any cache a user can have.

## Test instructions

1. Declare `node@22` and run `npm --version` in a real guest (`make e2e-microvm`, scenario "A launcher the provisioning engine rewrote still runs in the guest"). It prints a version, with no repair code in the path.
2. Inspect a freshly provisioned tree: `bin/npm` is a symlink to `../lib/node_modules/npm/bin/npm-cli.js`, not a bash script.
3. `lns run` a definition declaring `node@22` twice; the second run reuses the cached tree and npm still runs.

## Verification

- `make lint`, `make complexity`, `make coverage` (`lns-service`, 100% per-file) — green.
- `make e2e-microvm`: the declared-tools feature passed on this exact diff, including the npm scenario, before the rebase onto #192's latest commit. That commit only touches `lns-init` tool-tree ownership, so it is orthogonal — but the suite should be re-run once before the stack merges.

### DIFF
--- a/crates/lns-service/src/tools/cache.rs
+++ b/crates/lns-service/src/tools/cache.rs
@@ -46,30 +46,6 @@ pub enum EntryKind {
     Symlink { target: String },
 }
 
-const NPM_LAUNCHER: &str = "bin/npm";
-const NPM_CLI: &str = "lib/node_modules/npm/bin/npm-cli.js";
-const STOCK_NPM_TARGET: &str = "../lib/node_modules/npm/bin/npm-cli.js";
-
-/// The engine replaces node's `bin/npm` symlink with a shim that reshims through `mise` under bash — neither of which a workload guest has — so the tree goes back to the launcher the upstream tarball ships. The tree root is the install prefix, so node's own layout sits at some depth below it and only the tail of the path is fixed.
-fn restore_stock_npm(entries: &mut [ManifestEntry]) {
-    let prefixes: Vec<String> = entries
-        .iter()
-        .filter_map(|entry| {
-            entry
-                .path
-                .strip_suffix(NPM_CLI)
-                .map(|prefix| format!("{prefix}{NPM_LAUNCHER}"))
-        })
-        .collect();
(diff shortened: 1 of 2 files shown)


jansav:
Folded into #192 rather than reviewed separately: the same review round covers it, and it deletes code #192 itself modifies. GitHub marked this merged automatically because its head commit (9c0bccfd) landed on its base branch `tools-issue`. Nothing was merged to `main` — review continues in #192.

PR #227, opened 2026-08-12
fix: resolve a bare inspect, rm, or tag reference against the Lens hub

chenhunghan:
## Problem

```
$ lns inspect hchen/claude:0.0.4
Error: daemon error: inspecting hchen/claude:0.0.4: failed to pull manifest/config
for docker.io/hchen/claude:0.0.4: Not authorized: url https://index.docker.io/v2/hchen/claude/manifests/0.0.4
```

The Lens hub default (`hub.lns.run`) was applied per verb in the CLI, and only
`push`, `pull`, `run`, and `login` applied it. `inspect`, `rm`, and `tag` handed
an unqualified reference straight to the service, where `oci_client::Reference`
resolved it with *its* default registry — Docker Hub.

`rm` and `tag` were worse than cosmetic: `lns pull you/agent:1` stores the cache
record under `hub.lns.run/you/agent:1`, so `lns rm you/agent:1` looked up
`docker.io/you/agent:1` and could never find what the pull had just written.

## Fix

`apply_registry_default` now covers every verb that addresses a sandbox by
reference:

- `tag` qualifies `<SOURCE>` normally; a bare `<TARGET>` then follows the
  qualified source's registry, so a same-repo retag off a non-default registry
  never turns into a refused cross-registry pair (review follow-up, c4caeb2c).
- `inspect` and `rm` take an id-or-ref, so the default is applied only in the
  cached-artifact fallback, once the service has reported no such run. A run id
  or name still reaches `InspectRun` verbatim.

Qualification is no longer per-verb opt-in (review follow-up, 0d1d83c3): the
shared dispatch `run_with_writers` takes a `DispatchEnv { svc, term, registry }`
and applies `apply_registry_default` itself, so no present or future verb can
silently fall back to Docker Hub, and the `run_*` shortcuts are plain
one-liners again. The inspect/rm ride-along `registry` fields are gone — the
deferred fallback reads the env's registry. `push`, which never reaches the
service dispatch, qualifies in `push_local`, its single production funnel.

## Tests

Layer 2 (`crates/lns-cli/tests/behaviours/`), each watched failing first:

- `a bare inspect reference resolves against the Lens hub` — asserts the
  outgoing `InspectImage` carries `hub.lns.run/…`
- `a bare rm reference removes the Lens hub sandbox` — same for `RemoveImage`
- `a bare tag pair re-refs within the Lens hub` — same for both sides of
  `TagImage`
- `a bare tag target follows the source's registry` — `tag ghcr.io/team/hermes:1.4.0
  team/hermes:latest` sends `to = ghcr.io/team/hermes:latest`
- `a run name is inspected as a run, never registry-qualified` — the guard for
  the id-or-ref overload; proven to fail under an eager-qualification mutation

The existing `tag re-refs a cached sandbox` scenario now uses a fully-qualified
pair, so it still pins the pass-through case. The Layer-2 steps no longer prime
the registry by hand — removing that priming before centralizing qualification
was watched red on the bare pull/tag scenarios. A unit test pins the docker.io
interplay (a docker.io default leaves a bare tag pair to the parser).

## Docs

`docs/cli-reference.md` states the resolution rule on the `inspect`, `rm`, and
`tag` rows; `docs/running-workloads.md` names the full set of verbs it applies
to and the one exception (`tag`'s target follows its source).

Gate: `make lint && make complexity && make coverage` green.

## Note on the audit check

`make audit` failed on RUSTSEC-2026-0257 (`webbrowser` ≤1.2.1), published after
main's last build. The lockfile bump (#228) has since been picked onto this
branch (00b2e781), so audit is expected green here.


crates/lns-cli/src/sandbox/mod.rs
```diff
@@ -232,14 +238,23 @@ pub struct SandboxPruneArgs {
...
+    let qualify = |reference: &mut String| {
+        *reference = crate::config::resolve_default_registry(reference, registry);
     };
-    *reference = crate::config::resolve_default_registry(reference, registry);
+    match command {
+        SandboxCommand::Push(args) => qualify(&mut args.reference),
+        SandboxCommand::Pull(args) => qualify(&mut args.reference),
+        SandboxCommand::Tag(args) => {
+            qualify(&mut args.from);
+            qualify(&mut args.to);
```
jansav: **`tag`'s TARGET is qualified with the default registry, not the source's.**

`to` is resolved independently of `from`, but the service refuses any pair whose registry/repository differ (`crates/lns-service/src/image_store/mod.rs:318`). So `lns tag ghcr.io/team/hermes:1.4.0 hermes:latest` becomes `hub.lns.run/hermes:latest` and fails with "cross-repository tagging isn't supported; … requires `lns sandbox push`" — which is not what the user did. `SandboxTagArgs`' TARGET help still promises "New tag in the source sandbox's registry and repository", and this PR had to fully qualify both sides of the existing scenario to keep it passing.

**Why:** a same-repo retag off any non-default registry is unreachable, and the error blames a cross-repository push the user never asked for. Qualify `to` against `from`'s registry/repository (after `from` is qualified), or change the help text to say the registry must be repeated.

crates/lns-cli/src/sandbox/real.rs
```diff
@@ -84,17 +84,20 @@ pub fn run_kill<'a>(matches: &'a clap::ArgMatches, ctx: RunCtx<'a>) -> RunFuture
 pub fn run_rm<'a>(matches: &'a clap::ArgMatches, ctx: RunCtx<'a>) -> RunFuture<'a> {
     Box::pin(async move {
         let args = super::SandboxRmArgs::from_arg_matches(matches)?;
-        dispatch_command(super::SandboxCommand::Rm(args), ctx.input).await
+        let mut command = super::SandboxCommand::Rm(args);
+        qualify_references(&mut command)?;
```
jansav: **Qualification stays per-verb opt-in, which is the bug being fixed.**

Four shortcut functions (`run_rm`, `run_inspect`, `run_pull`, `run_tag`) now each remember to call `qualify_references` before `dispatch_command`. Nothing enforces it — a new reference-addressing verb that forgets the call silently falls back to Docker Hub again, exactly the defect this PR closes. `dispatch_command` is the single funnel all of them pass through, and `apply_registry_default` is a no-op for the run-addressing verbs.

**Why:** moving the call into `dispatch_command` removes the whole class of regression; leaving it distributed means the next verb repeats this PR. The same applies to the `run_with_writers` seam: the Layer-2 steps had to be patched to prime `registry` by hand (`tests/behaviours/steps/sandbox_cli.rs:557`, `steps/declared_tools.rs:105`), so a caller that forgets gets the wrong registry with no signal.

jansav (commented):
**Problem:** Only `push`/`pull`/`run`/`login` applied the Lens hub default, so `inspect`, `rm`, and `tag` handed bare references to the service, where `oci_client` resolved them against Docker Hub — `lns rm you/agent:1` could never find what `lns pull you/agent:1` had just cached.

**Approach:** `apply_registry_default` now covers `tag` (both sides) and rides along on `inspect`/`rm` via a new `#[arg(skip)] registry` field, applied only in the cached-artifact fallback so a run id/name still reaches `InspectRun` verbatim. The top-level shortcuts route through the same `qualify_references` seam.

**Product alignment:** Correct — the hub is the documented default for every reference-addressing verb, and the docs now say so. `cargo test -p lns-cli` is green here (408 scenarios).

**Verdict:** Needs fixes — two design points, both in the "should fix" band.

crates/lns-cli/src/sandbox/mod.rs
chenhunghan: Fixed — a bare TARGET now resolves against the qualified SOURCE's registry instead of the default, so `lns tag ghcr.io/team/hermes:1.4.0 team/hermes:latest` sends `to = ghcr.io/team/hermes:latest` and a same-repo retag off a non-default registry works; the cross-repository refusal now only fires when the repositories genuinely differ. TARGET's help and the docs state the rule, and the docker.io interplay (both sides stay bare for the parser) is pinned as a unit test. Behaviour scenario: "a bare tag target follows the source's registry". (commit c4caeb2c)

crates/lns-cli/src/sandbox/real.rs
chenhunghan: Fixed — qualification now lives in the shared dispatch itself: `run_with_writers` takes a `DispatchEnv { svc, term, registry }` and applies `apply_registry_default` before matching on the verb, so an entry point can't opt out (the compiler demands the registry) and the four `run_*` shortcuts are back to plain one-liners. The inspect/rm ride-along `registry` fields are gone — the deferred qualification reads the env's registry instead — and the Layer-2 steps no longer prime anything by hand. Push, which never reaches the service dispatch, qualifies in `push_local`, its single production funnel. Proven red-first: removing the step priming before the refactor failed the bare pull/tag scenarios; with qualification in `run_with_writers` they pass with no caller involvement. (commit 0d1d83c3)

chenhunghan:
@copilot+claude-fable-5 resolve the merge conflicts in this pull request

chenhunghan:
@copilot+claude-sonnet-5:max resolve the merge conflicts in this pull request

chenhunghan:
Closing as stale: the defect is still real on main (a bare ref handed to the cache verbs is normalized service-side by oci_client's docker.io default while pull records under hub.lns.run — cli-spec §2.3 says a bare REF qualifies against --registry / run.registry / hub.lns.run, never guessed), but this diff predates the artifact/sandbox namespace split (#283) and the shortcut arbitration rework (#313), so the fix is being reimplemented fresh against the current CLI. The tag rule from this PR's review (a bare TARGET follows the qualified SOURCE's registry) carries over.

PR #236, opened 2026-08-14
feat!: let the later source decide where two disagree about a destination

jansav:
> Stacked on #233 (→ #232 → #231). Base is `feat/lowercase-kind`.
>
> First of the steps closing [§8](https://github.com/lensapp/lens-sandbox/blob/main/docs/sandbox-spec.md#8-the-local-mixin). This one closes [§4.2](https://github.com/lensapp/lens-sandbox/blob/main/docs/sandbox-spec.md#42-the-egress-definition)'s ordering rule and the half of §8.1 that is about precedence.

## The divergence

§4.2 is normative and unambiguous:

> **Across sources, the later source decides.** Its entries are placed ahead of earlier ones in the merged table…
> A later source can therefore turn a `deny` into an `allow`. That is deliberate — overriding is what layering is for — and it is why the merged table is disclosed before boot rather than trusted silently: the developer approves the effective egress, not each contributor's opinion of it.

And §8.1: the developer's decisions are "Last in the merge… Nothing they pulled can overrule what they decided."

`merge_effective` did the opposite. It hoisted every layer's denies to the front, then dropped any allow that a deny-by-default layer did not independently permit (`allowed_by_every_ceiling`). A pulled artifact clamped the developer running it.

This behaviour appeared in no guide. It was undocumented code contradicting a normative section — which is how it survived.

## Why now

§4.2 names its own safeguard: the merged table is *disclosed before boot*. That disclosure did not exist until #232 in this stack, which is why this PR comes after it and not before. Deleting a clamp while nothing showed the developer the result would have removed a safety net with nothing underneath.

## What is lost, stated plainly

**A published artifact can no longer enforce an outbound lockdown against a careless local file.** A local `allow *` now outranks an artifact's targeted `deny telemetry.evil`.

§4.2 makes that trade deliberately, and it holds for this product: the artifact's deny is what the workload's author needed, not a boundary over the person running it; a destination nobody decided is still asked about; and enforced org ceilings are Lens Agents' job.

**The developer's own lockdown is not lost** — it is expressed by ordering instead of intersection. A local catch-all deny sits ahead of every pulled entry and decides them. Three tests that asserted the *losing* rule was absent now assert the *winning* rule comes first, which is what a first-match gate actually means.

One second-order fix falls out: an approval the developer had just clicked could previously be dropped by the clamp on the next watcher reload. Under precedence an approval sticks.

## The two tables stay independent, in both directions

A raw rule must name a port (§4.2), so no `tcp` catch-all is expressible and a `deny *` written in `http` was never a raw lockdown. The old code compensated by selecting ceilings from the `http` table and clamping *both* — which the specification does not describe.

Review caught that this flips a case my first accounting missed: a developer's `deny *` no longer suppresses a pulled artifact's raw-TCP allow. Landed as the spec-faithful behaviour and pinned as an explicit decision rather than an accident, beside the test for the reverse direction.

**One residual limitation is the specification's, and is deliberately not settled here:** a developer cannot express "deny all raw TCP" at all, only per-destination raw denies. Closing that needs a decision — a `tcp` catch-all grammar, or a rule that an `http` catch-all deny closes both tables — not a patch.

## Verification

`make lint`, `make complexity`, full-workspace `make coverage` and `make e2e` (69 scenarios) all exit 0.

Net **375 deletions against 214 insertions**: `allowed_by_every_ceiling`, `permits`, `tcp_permits`, `scope_within`, `live_rules`, the deny-first hoist and the catch-all repositioning are all gone.

The twelve tests pinning the old model were split by what each actually asserted — four inverted (the direction the spec drops), three rewritten as ordering (the direction that survives), four deleted (binary-scope clamping, whose mechanism is gone; review confirmed the guest gate carries that property itself, failing closed for an excluded caller).

Mutation-checked: placing the artifact ahead of the developer turns the Layer 2 scenario *a mixin cannot open a directory that denies by default* red.

## A test helper that was lying

`request_denied_by_policy` commented "the guest gate is first-match-wins" and implemented `find(|r| r.match_pattern == host)` — an exact string compare that cannot see a `*` catch-all. The old clamp *deleted* the losing allow, so exact-match happened to give the right answer; once the allow legitimately survived behind the deny, the helper picked the wrong rule.

Both it and its sibling now share `gate_verdict()`, a real first-match through `domain_matches`. It also no longer treats "no rule matches" as denied — being asked is not being refused.

Review found the same class of defect in my own rewrite: `allow_idx < deny_idx` compares `Option<usize>`, and `None < Some(_)` is `true`, so the test passed on the very regression it guarded. Both positions are now bound with `expect` before comparing; making the merge drop the later source turns 22 tests red.

## Review round 2: the safeguard §4.2 names did not exist for an uncomposed run

The clamp's replacement is disclosure. Review found that for a **published sandbox that layers on nothing**, there was no disclosure at all — so the one state where the deleted clamp was load-bearing was also the state §4.2's reader did not cover.

The raw table is what made it visible. A directory's catch-all deny is an `http` rule, and a raw rule must name a port, so nothing a developer can write decides the raw table: a pulled `allow db.vendor.example:5432` fires while the summary — which reads the local file alone — reports the directory denies by default. It does not even print `0 raw allow`; with an empty local raw table it prints no raw signal at all.

The same hole covers `http`. An allow a pulled sandbox ships against a default-ask directory was equally unnamed, so this is a class, not a corner.

Two things were in the way, and both needed fixing:

- `mixin::resolve` returns early for a document with no mixins, so `contributions` was empty and there was nothing to print. It now attributes the document's own egress to `the sandbox`.
- The summary's `Rules:` block was gated on having resolved a mixin. It now prints whenever there are egress contributions; the `Mixins:` line stays gated.

An uncomposed run names no source on those lines — one author was never an attribution question, and `an_uncomposed_run_prints_exactly_what_it_printed_before_attribution_existed` still holds for a run that ships no rules.

The alternative was to re-add a clamp for `egress.tcp` alone. Rejected: it reverses a decision this PR's review already pinned deliberately, it reintroduces a mechanism §4.2 does not describe, it would settle the open "does an http catch-all close both tables" question in code by accident, and it closes one corner while leaving the http side of the same hole open.

## Two names that outlived their model

`set_policy_floor` documented itself as "an always-merged floor … can never drop the sandbox's rules". A floor is a lower bound that cannot be undercut, which is exactly what this PR deletes — the local file overrules it. It is `set_shipped_policy` now, described as the earlier source each reload re-merges under. `ceiling_denies` is `denies_by_default`.

## Seven tests that could no longer fail

The merge is unconditional concatenation now, so "rule X survives the merge" is structurally true for any input. Seven ceiling-era tests asserting only presence are deleted; review confirmed the two that looked like ordering proofs were redundant against `a_later_sources_entries_precede_an_earlier_sources` and `the_raw_table_carries_both_sources_in_precedence_order`. Reversing the layer order in `merge_effective` still turns three surviving tests red, so no coverage left with them.

One kept its assertion and lost its vocabulary: an allow stranded behind a catch-all must not let an earlier source's identical allow through, which is a real property of a first-match gate.

## Not in this PR

The rest of §8: the rule-4 tier in `flatten`, `spec.egress` vs `spec.policy.egress`, top-level `name:` vs `metadata.name`, moving `connectors:` to the grant sidecar, converting the decisions file to mixin grammar, and the `lns-local-mixin.yaml` rename.

### DIFF
--- a/crates/lns-service/src/approval_flow/session.rs
+++ b/crates/lns-service/src/approval_flow/session.rs
@@ -147,7 +147,7 @@ pub struct ApprovalSession {
     connector: OnceLock<Arc<dyn ConnectPort>>,
     connecting: Mutex<HashSet<String>>,
     ledger: OnceLock<Arc<dyn LedgerRecorder>>,
-    policy_floor: OnceLock<Policy>,
+    shipped: OnceLock<Policy>,
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
@@ -180,7 +180,7 @@ impl ApprovalSession {
             connector: OnceLock::new(),
             connecting: Mutex::new(HashSet::new()),
             ledger: OnceLock::new(),
-            policy_floor: OnceLock::new(),
+            shipped: OnceLock::new(),
         }
     }
 
@@ -205,9 +205,9 @@ impl ApprovalSession {
         let _ = self.connector_routes.set(deriver);
     }
 
-    /// Installs a sandbox's shipped policy as an always-merged floor, so a watcher reload of the local overlay can never drop the sandbox's rules; idempotent, the first wins.
-    pub fn set_policy_floor(&self, floor: Policy) {
-        let _ = self.policy_floor.set(floor);
+    /// Installs a sandbox's shipped policy as the earlier source every reload re-merges under, so a watcher reload of the local decisions carries the sandbox's rules along rather than replacing them; idempotent, the first wins.
+    pub fn set_shipped_policy(&self, shipped: Policy) {
+        let _ = self.shipped.set(shipped);
     }
 
     /// Installs the connect port once the credential subsystem exists; idempotent, the first wins.
@@ -536,8 +536,8 @@ impl ApprovalSession {
 
     pub fn apply_external_policy(&self, mut new_policy: Policy) {
         *self.persisted.lock().expect("persisted mutex poisoned") = new_policy.clone();
-        if let Some(floor) = self.policy_floor.get() {
-            new_policy = crate::artifact::policy::merge_effective(Some(floor), &new_policy);
+        if let Some(shipped) = self.shipped.get() {
+            new_policy = crate::artifact::policy::merge_effective(Some(shipped), &new_policy);
         }
         if let Some(derive) = self.connector_routes.get() {
             let routes = derive(&new_policy.connectors);
@@ -1743,7 +1743,7 @@ pub(crate) mod tests {
         let (s, _n, _store, mut rx) = fixture();
         let mut document = Policy::default();
         document.add_rule(RouteRule::deny_host("api.example.test"));
-        s.set_policy_floor(document);
+        s.set_shipped_policy(document);
 
         let mut reloaded = Policy::default();
         reloaded.add_rule(RouteRule::allow_host("api.example.test"));
--- a/crates/lns-service/src/approval_flow/session.rs
+++ b/crates/lns-service/src/approval_flow/session.rs
@@ -1739,31 +1739,28 @@ pub(crate) mod tests {
     }
 
     #[test]
-    fn a_policy_floor_survives_a_reload_keeping_its_deny_deny_dominant() {
+    fn a_reloaded_decision_decides_over_the_document_it_layers_onto() {
         let (s, _n, _store, mut rx) = fixture();
-        let mut floor = Policy::default();
-        floor.add_rule(RouteRule::deny_host("api.example.test"));
-        s.set_policy_floor(floor);
+        let mut document = Policy::default();
+        document.add_rule(RouteRule::deny_host("api.example.test"));
+        s.set_policy_floor(document);
 
-        // A watcher reload of the local overlay that *allows* the host must not drop the floor's deny.
         let mut reloaded = Policy::default();
         reloaded.add_rule(RouteRule::allow_host("api.example.test"));
         s.apply_external_policy(reloaded);
 
         let routes = s.current_policy().network.egress.http;
-        let deny_idx = routes
-            .iter()
-            .position(|r| r.match_pattern == "api.example.test" && r.verdict == Verdict::Deny);
         let allow_idx = routes
             .iter()
-            .position(|r| r.match_pattern == "api.example.test" && r.verdict == Verdict::Allow);
+            .position(|r| r.match_pattern == "api.example.test" && r.verdict == Verdict::Allow)
+            .expect("the approval the developer just made must survive the reload");
+        let deny_idx = routes
+            .iter()
+            .position(|r| r.match_pattern == "api.example.test" && r.verdict == Verdict::Deny)
+            .expect("what the document said is still in the table, behind the decision that overruled it");
         assert!(
-            deny_idx.is_some(),
-            "the floor's deny must survive the reload"
-        );
-        assert!(
-            deny_idx < allow_idx,
-            "the floor's deny must stay ordered before the reloaded allow: {routes:?}"
+            allow_idx < deny_idx,
+            "the developer's decisions file is the last source, so an approval they just made must decide rather than be dropped by the document under it: {routes:?}"
         );
         let _ = policy_frame(&mut rx);
     }
--- a/crates/lns-service/src/artifact/policy.rs
+++ b/crates/lns-service/src/artifact/policy.rs
@@ -1,7 +1,5 @@
-use std::path::Path;
-
-use lns_policy::matching::{destination_covers, split_destination};
-use lns_policy::{NetworkPolicy, Policy, RouteRule, TcpEgressRule, Verdict};
+use lns_policy::matching::split_destination;
+use lns_policy::{NetworkPolicy, Policy, RouteRule, Verdict};
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum GuardrailFlag {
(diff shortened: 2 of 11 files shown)


crates/lns-service/src/artifact/policy.rs
```diff
@@ -567,50 +463,106 @@ mod tests {
...
+
+        assert_eq!(
+            table(&merged).first().copied(),
+            Some(("api.example.test", Verdict::Allow)),
+            "the developer asked for this host uninspected, and they are the later source"
+        );
+    }
+
+    #[test]
+    fn a_developers_http_catch_all_does_not_decide_the_raw_table_either() {
```
chenhunghan, line 515: **The raw-TCP leg loosens with no clamp *and* no disclosure.**

The justification for dropping the clamp is §4.2's own safeguard — "the merged table is disclosed before boot". For the case this PR newly loosens, it isn't.

An artifact's `egress.tcp: allow db.vendor.example:5432` used to be dropped by a local `deny *` (the deleted `a_deny_by_default_layer_clamps_the_other_layers_tcp_allow`). It now survives, resolves through the DNS union (`PortScope::PerPort` → `Allowed`), and connects. Meanwhile `print_run_summary` reads `Policy::load_or_default(&path)` — the local file alone — so the developer is shown `unmatched destinations: denied by the catch-all rule` and `0 raw allow`. The `Rules:` contribution block that would name those entries is gated on `!args.resolved_mixins.is_empty()`, so a plain published-sandbox run prints nothing at all; `guardrail_flags` only fires on broad wildcards/CIDRs, and a targeted raw allow raises no card.

**Why:** a developer with `defaultVerdict: deny` is told their directory is locked down while a pulled artifact's raw port is live — the one state where the removed clamp was load-bearing is also the one the disclosure doesn't cover. Either surface the merged raw entries pre-boot (count them from the effective table, or print the `Rules:` block whenever there are Egress contributions), or hold this leg until §4.2's "deny all raw TCP" question is decided.

crates/lns-service/src/approval_flow/session.rs
```diff
@@ -1739,31 +1739,28 @@ pub(crate) mod tests {
...
         let allow_idx = routes
             .iter()
-            .position(|r| r.match_pattern == "api.example.test" && r.verdict == Verdict::Allow);
-        assert!(
-            deny_idx.is_some(),
-            "the floor's deny must survive the reload"
-        );
+            .position(|r| r.match_pattern == "api.example.test" && r.verdict == Verdict::Allow)
+            .expect("the approval the developer just made must survive the reload");
+        let deny_idx = routes
```
chenhunghan, line 1757: **`policy_floor` / `ceiling_denies` still name the deleted model.**

The test was renamed off "floor"; the production API wasn't. `set_policy_floor` (`session.rs:208`, field at `:150`, use at `:539`) still documents itself as "an always-merged floor, so a watcher reload … can never drop the sandbox's rules". A floor is a lower bound that can't be undercut — the whole point of this change is that the local file overrules it. `ceiling_denies` in `credential_flow/connectors.rs:195` (and its doc at `:1317`) reads the same way; its behaviour is still correct, since `is_closed` reads the first catch-all of a now precedence-ordered table, so only the name is stale.

**Why:** the next person to touch the merge reads "floor"/"ceiling" and reconstructs the model you just deleted.

crates/lns-service/src/artifact/policy.rs
```diff
@@ -935,25 +789,20 @@ mod tests {
...
+            Some(&closed_allowing("api.artifact-only.example")),
+            &closed_allowing("api.user-only.example"),
         );
-        assert!(
-            is_closed(&merged),
-            "a closed layer must leave the merged policy closed: {:?}",
-            merged.network.egress.http
+
+        assert_eq!(
+            table(&merged),
```
chenhunghan, line 643: **Seven ceiling-era tests survived that no longer constrain anything.**

The merge is now unconditional concatenation, so "rule X survives the merge" is a structural truth for any input. These four assert only presence and can no longer fail under any input:

- `a_ceiling_authorizes_an_allow_its_own_wildcard_already_covers` (`:446`)
- `a_ceiling_authorizes_a_raw_splice_inside_a_range_it_already_splices` (`:570`)
- `merge_reads_a_redundant_separator_as_the_same_binary_the_ceiling_named` (`:752`)
- `merge_keeps_an_allow_both_layers_carry_when_both_deny_by_default` (`:772`)

The redundant-separator one specifically pinned `scope_within`'s `Path` comparison, which is deleted — nothing in production compares binary paths anymore.

Three more (`:607`, `:703`, `:727`) only re-prove the layer ordering that `a_later_sources_entries_precede_an_earlier_sources` already pins, while their names and failure messages still describe ceiling authorization.

**Why:** the PR body accounts for twelve tests (four inverted, three rewritten, four deleted) but these were left behind. They pass by construction and advertise a mechanism that no longer exists — the exact thing the "what behavior am I pinning" bar exists to prevent.

chenhunghan (commented):
**Problem:** `merge_effective` implemented the opposite of §4.2 — it hoisted every layer's denies to the front and intersected allows against each deny-by-default layer, so a pulled artifact clamped the developer running it. This replaces that with plain precedence: the local file's entries sit ahead of the artifact's in a first-match table.

**Approach:** Delete the ceiling machinery (`allowed_by_every_ceiling`, `permits`, `tcp_permits`, `scope_within`, `live_rules`, the deny-first hoist, catch-all repositioning) and concatenate sources later-first, deduped. Test helpers that faked the gate with exact-string matching now share a real first-match `gate_verdict()`.

**Product alignment:** Correct per §4.2/§8.1. I checked the pinned `lens-sandbox-core` (`ceb3815`) — `find_matching_route`, `find_matching_tcp_egress` and the DNS gate are genuinely first-match with a no-reopen guard, so ordering does carry the semantics, including the binary-scope property whose four tests were deleted. The approach is sound.

**Verdict:** Needs fixes — three findings, none of them about the direction.

chenhunghan (approved):
Fine for fixing it in follow up PRs

crates/lns-service/src/artifact/policy.rs
jansav: Confirmed, and the hole is wider than the raw leg. Fixed in a8085a21.

Two things stood between the developer and the disclosure §4.2 relies on, and closing one without the other prints nothing:

- `mixin::resolve` returns early for a document with no mixins, so `contributions` was empty. Ungating the block alone would have rendered an empty list. `resolve` now attributes the document's own egress to `the sandbox`, via a new `merge::own_egress` so the key format stays single-sourced with the merge's own `egress_contribution` — if those two ever disagreed, the disclosure would lie about what a composed run shows.
- The `Rules:` block is no longer gated on `!args.resolved_mixins.is_empty()`. It prints whenever there are egress contributions; the `Mixins:` line stays gated.

An uncomposed run lists the rules but carries no `[from …]` suffix, since one author was never an attribution question — `write_block` takes an `attributed` flag, and `an_uncomposed_run_prints_exactly_what_it_printed_before_attribution_existed` still holds for a run that ships no rules.

Two corrections to the finding, neither of which changes the conclusion:

**It was worse than "0 raw allow".** With an empty local raw table `raw_counts` returns the empty string, so the summary printed *no* raw signal at all beside "denied by the catch-all rule".

**It is not only the raw leg.** An `http` allow a pulled sandbox ships against a default-ask directory was equally unnamed. So this is a class, not a corner — which is also why I did not take the second option.

I considered holding the raw leg behind a `tcp`-only clamp and rejected it: it reverses a decision this PR's earlier review pinned deliberately, it reintroduces a mechanism §4.2 does not describe, it would settle the open "does an `http` catch-all close both tables" question in code by accident, and it closes one corner while leaving the `http` half of the same hole open.

Both halves are red-first and mutation-checked — removing the block call turns four summary tests red, replacing `own_egress(&def.spec)` with `Vec::new()` turns the service test red. One coverage miss caught on the way: the unattributed render of the second and later entries in a block.

crates/lns-service/src/approval_flow/session.rs
jansav: Agreed — the doc was worse than the name. "An always-merged floor, so a watcher reload … can never drop the sandbox's rules" describes the guarantee this PR removes. Fixed in a8085a21.

- `policy_floor` → `shipped`, `set_policy_floor` → `set_shipped_policy`, documented as the earlier source each reload re-merges under. A reload carries the sandbox's rules along rather than replacing them; where the two disagree, the local file decides.
- `ceiling_denies` → `denies_by_default` in `credential_flow/connectors.rs`, and its test fixture's doc no longer says a catch-all deny "makes this policy a ceiling".
- `start_merges_a_sandbox_policy_floor_under_the_local_overlay` → `start_merges_a_sandbox_shipped_policy_under_the_local_decisions`.

The remaining `ceiling` hits in the tree are unrelated — byte ceilings in `image/mod.rs` and the resource ceiling in §7 of the spec.

crates/lns-service/src/artifact/policy.rs
jansav: Right on all seven. Deleted in a8085a21.

I checked the three you flagged as ordering re-proofs rather than taking them on faith, and they are redundant against tests that already pin the same thing more directly: `a_later_sources_entries_precede_an_earlier_sources` for the `http` table, `the_raw_table_carries_both_sources_in_precedence_order` for the raw one. Both assert position, which is the property that survives; the three deleted ones asserted it incidentally while their names and messages described ceiling authorization.

`merge_keeps_an_allow_both_layers_carry_when_both_deny_by_default` had a second assertion (`is_closed(&merged)`) that is not pure presence, but it is structurally true under concatenation — the overlay's catch-all is still the first one the gate reaches — and the property it gestures at is pinned properly by `a_developers_lockdown_still_decides_a_destination_the_artifact_allows` and `two_lockdowns_leave_only_what_the_developer_named`.

I kept one and renamed it: `a_dead_allow_in_a_ceiling_does_not_legitimize_another_layers_copy_of_it` → `an_allow_stranded_behind_a_catch_all_does_not_decide_an_earlier_sources_copy_of_it`. Its assertion is real — a first-match gate must reach the catch-all before either copy of the allow — so it kept the assertion and lost the vocabulary.

To check no coverage left with the deletions: reversing the layer order in `merge_effective` still turns three surviving tests red.

jansav:
All three comments are addressed in a8085a21, replies in thread.

Summary of what changed:

- **The disclosure §4.2 relies on now exists for an uncomposed run.** A document that resolves to itself attributes its own egress, and the `Rules:` block prints whether or not the run composed anything. This closes the `http` half of the same hole, not just the raw leg.
- **Two names that outlived the deleted model** — `set_policy_floor` → `set_shipped_policy`, `ceiling_denies` → `denies_by_default`.
- **Seven structurally-unfailable tests deleted**, one kept and renamed.

The stack above this PR (#237, #238, #242, #239, #240) has been rebased onto the new head and force-pushed. Worth knowing: the #237 rebase applied with **zero conflicts and did not compile** — `own_egress` and its fixture still read `spec.policy.egress`, the exact field #237 renames. Fixed inside #237's own commit.

`make lint`, `make complexity`, full-workspace `make coverage`, `make test` and `make e2e` all exit 0 here and on the stack tip.

PR #249, opened 2026-08-17
feat!: pack each path fileset into a layer of the artifact that declares it

chenhunghan:
## The problem

A fileset was a fourth artifact kind. The specification permits only three kinds: `sandbox`, `connector`, and `mixin`.

The old behavior:

- `lns push` published each `filesets[].path` directory as its own **FileSet** artifact.
- That artifact used the retired API group `lens.dev/v1alpha1`. No other code used this group.
- The push removed the author's `path` field and wrote a `ref` field. The specification does not define `ref`.
- Each boot of such a sandbox fetched extra manifests and layers, one set for each fileset.
- A pulled document that kept a `path` entry was refused.

## What the specification says

`docs/sandbox-spec.md` gives these rules:

- **§3.1.11** — A fileset is not a separate artifact. A `path` directory goes into a layer of the same artifact. The files and the entry that mounts them share one digest. To share one directory across sandboxes, publish a mixin that carries it.
- **§6** — The publish transform packs the directory into a layer of this artifact. The entry keeps its `path` and `mountPath`.
- **§7** — A `sandbox` or `mixin` artifact carries one layer for each `filesets[].path` entry. Nothing else is addressable on its own.

This change makes the code follow these rules.

## The change

**Push.** `lns push` packs each `path` directory into a layer of the same artifact. The layer type is `application/vnd.oci.image.layer.v1.tar+gzip`. The layers follow the declaration order. The published entry does not change: `path`, `mountPath`, and `owner` stay. The tar is deterministic (sorted entries, zero mtime, zero uid/gid) and the gzip has no timestamp. Two equal directories give one blob digest, so registries store the data once. One push is one upload. `--dry-run` shows each layer digest and does not use the network. A document with no `path` entries publishes config-only, with the standard OCI empty descriptor.

**Correlation.** The position of the entry gives the position of the layer, in the source document. Entry number i owns layer number i of that document's artifact. This rule is per source document. After a merge, entries from different sources mix in one list. A mixin's first `path` entry maps to the first layer of the mixin's own artifact, not to a position in the merged list. The merge returns a `FilesetOrigin` for each packed entry that survives. A shared directory now works as the specification says: publish a mixin that carries it, and the fileset materializes from the mixin's own digest.

**Pull.** A `path` entry in a pulled document is now the correct, runnable shape. The run fetches its layer into the content store, keyed by digest. The run compares the blob digest with the digest in the manifest. The run mounts the files at `mountPath` with the same `owner` rules as before. `published_fileset_problems` now refuses the opposite case: a `path` entry with no layer behind it. Without this check, the run would read a directory from the consumer's own disk. When the layer count and the `path` entry count of one source do not agree, the run stops and names the artifact to publish again.

## The result

Example: a document declares two `path` filesets. `lns push` now makes one artifact:

```
manifest
├── config: the document, unchanged (path entries stay)
├── layer 1: tar+gzip of ./skills
└── layer 2: tar+gzip of ./prompts
```

Before this change, the push made three artifacts and rewrote the document. One reference now names one complete thing. The digest the consumer approves covers the document and the files together.

## Deleted

- `Kind::FileSet` and both `vnd.lens.fileset` media types.
- The `lens.dev/v1alpha1` API group constant and its parser.
- The top-level `mount` key.
- `parse_fileset`, `build_fileset`, and the `ref` rewrite in `pack_path_filesets`.
- The `ref` field on `FilesetEntry` and `SandboxFileset`.
- The per-boot fetch of separate fileset artifacts.

A fileset source is now exactly one of three: `path`, `inline`, or `hostPath`.

The product is pre-1.0, so there is no compatibility path. A published document that carries `ref` now fails with `unknown field 'ref'`. This is the intended answer: publish it again.

## Not changed

- `inline` filesets.
- Local runs that read `path` from the disk, from the directory of the document.
- The `hostPath` and `optional` fields. Their removal is a separate decision.
- The guest-side mount machinery. It gets a different source, not a rewrite.

## One new IPC field

A local `lns.yaml` can layer on a published mixin that ships a packed fileset. The merged document alone cannot say which artifact carries which mount. The service resolves the document in the preflight, so the service knows the coordinates. `Response::DefinitionResolved` and `RunImageArgs` carry a new list: `packed_filesets`. This is the CLI-to-service wire, not the document format. A published reference does not use the field. The service reads its own manifest at boot for that path.

## Tests

Each test was written first and seen to fail:

- **Layer 2 (lns-cli)**: a push packs a `path` fileset into a layer of the same artifact, and the published document keeps its `path`; `--dry-run` shows the layer digest; `inline` and `hostPath` entries pack nothing.
- **Layer 2 (lns-service)**: a pulled sandbox mounts its `path` fileset from the approved artifact; a layer-count mismatch stops the run; a fileset from a mixin mounts from the mixin's own artifact.
- **Layer 3**: deterministic packing (same tree gives same digest, in any entry order); the per-source correlation, with a displaced entry and an entry that lost its target to a volume; the count check; the digest check on pull; strict-decode refusal of `ref`, top-level `mount`, and `lens.dev/v1alpha1`.
- **Layer 1**: the distribution and pull-confirm features now pin the packed layer, not the removed companion artifact.

## Docs

`docs/running-workloads.md` and `docs/cli-reference.md` describe the new behavior. `docs/sandbox-spec.md` is not changed.


jansav (approved):
Approved. The change follows §3.1.11, §6 and §7: one push, one artifact, one layer for each `path` entry, and the entry published as written. The specification is untouched.

Every gate is green, `make e2e` included. The per-source correlation rule is the right one — a mixin's entry maps to a layer of the mixin's own artifact, never to a position in the merged list. The digest check on pull and the refusal of a `path` entry with no layer behind it close the two ways a mount could go unseeded.

Two follow-ups, neither a blocker:

- The layer is `tar+gzip` and adds `flate2`. Plain tar needs no new dependency, and registries already compress on the wire.
- `packed_filesets` carries resolved blob coordinates. A fail-closed rule in the service — fetch only a digest-pinned origin, read every other label from disk — would keep the trust decision on the side that acts on it.

PR #250, opened 2026-08-17
fix: state the disk caching mode Vz must use, instead of taking its default

### LINKED ISSUE #247: Guest reads a directory block as zeroes and remounts read-only, while the image on disk is intact
## What happened

A long-lived sandbox took `EXT4-fs error` on two independent block devices within one uptime, remounted both read-only, and eventually wedged. Every write in the guest then failed — the first symptom I saw was an agent unable to write `/tmp`, which is not its own mount but part of the overlay root.

The devices are fine. I stopped everything and ran `e2fsck -fn` against the images from the host: **all of them pass with no errors, including both devices the guest declared corrupt.** So the blocks the guest read as zeroes are intact on disk, and the fault is in the read path rather than in the store.

The reported corruption is always a directory block that reads as entirely zero — `rec_len=0, inode=0, offset=0, size=4096` — and it is always detected in `ext4_readdir`, never on a write. That is the shape of a buffer left zero-filled because a read did not populate it and the failure was not propagated.

Four details that seem to matter:

- **Two devices, one uptime.** `vde` (a named volume at `/opt/target`) failed first, then `vda` (the run's overlay upper) about six minutes later. A third volume, `vdc` at `/root/dev`, was clean and writable when I first probed and had gained `emergency_ro` roughly ten minutes later. This spreads while the box is up — it is not a boot or shutdown artifact.
- **Journalling is not the discriminator.** `vda` is journalled (`ordered data mode`) and failed anyway, alongside non-journalled volumes. I initially suspected the non-journalled volumes were the vulnerable ones; that does not survive the evidence.
- **The churn-heaviest directory went first.** `/opt/target/debug/.fingerprint` — thousands of small files, constant `readdir`, and the most traffic into sparse regions of the image.
- **Single writer.** Only `com.apple.Virtualization.VirtualMachine.xpc` held the images open, so this is not concurrent access.

Not a space problem: the guest had `/` at 4%, `/opt/target` at 40%, `/root/dev` at 64%, and the host had 77 GiB free. The images are sparse (10 GiB apparent, 6.4 GiB allocated for the busiest one), which is why I suspect the sparse read path specifically.

I also checked and ruled out macOS reclaiming purgeable space from `~/Library/Caches`, where the volume store lives: no purge or APFS-pressure events in the host log for the window, and no purgeable xattr on the images.

## Steps to reproduce

I have not reduced this to a minimal case — it emerged from ordinary use over ~45 minutes of uptime. What the workload was doing:

1. Run a long-lived sandbox with named volumes attached, one of them holding a Cargo target directory (`CARGO_TARGET_DIR` pointed at a volume mount).
2. Drive sustained small-file churn there — Cargo rebuilds, which rewrite `target/debug/.fingerprint` constantly.
3. Wait. The first `EXT4-fs error` arrived ~15 minutes in, on the volume carrying the target directory.
4. Once the overlay upper takes the same error, every write in the guest fails and the box becomes unusable.

Happy to try a narrower reproduction if it would help — a loop creating and deleting many small files in one directory on a volume seems like the place to start.

## Expected behavior

A read that cannot be satisfied should surface as an I/O error, not as a silently zero-filled buffer. ext4 is behaving correctly given what it was handed; it validated a directory block, found zeroes, and did the safe thing. The bug is that it was handed zeroes for a block that exists on disk.

## Environment

- OS and hardware: macOS 26.6.1, arm64, 14 cores, 48 GiB
- `lns --version`: lns 0.18.0 (built from 42c5f110, 2026-08-14)
- microVM runtime (Vz / KVM): Vz
- Workload (image or command): `docker.io/library/buildpack-deps:bookworm`, running a terminal multiplexer with coding agents inside; four named volumes plus one virtiofs bind; `user: root`; resources 80% of host cores and RAM

## Logs / output

Guest kernel log — the two failures, six minutes apart:

```
[  940.218035] EXT4-fs error (device vde): ext4_readdir:262: inode #114692: block 461312: comm ls:
  path /opt/target/debug/.fingerprint: bad entry in directory: rec_len is smaller than minimal -
  offset=0, inode=0, rec_len=0, size=4096 fake=0
[  940.225247] EXT4-fs (vde): Remounting filesystem read-only
[ 1281.049993] EXT4-fs error (device vda): ext4_readdir:262: inode #368745: block 1475106: comm Bun Pool 4:
  path (unknown): bad entry in directory: rec_len is smaller than minimal -
  offset=0, inode=0, rec_len=0, size=4096 fake=0
[ 1281.057616] EXT4-fs (vda): Remounting filesystem read-only
```

Mount state afterwards — note `emergency_ro` on two volumes, and that the overlay still advertises `rw` while every path on it rejects writes:

```
overlay / overlay rw,relatime,lowerdir=/composefs-meta::/content,upperdir=/mnt/upper/upper,workdir=/mnt/upper/work,uuid=on 0 0
/dev/vdc /root/dev ext4 rw,nosuid,nodev,relatime,errors=remount-ro,emergency_ro 0 0
/dev/vdd /opt/lns ext4 rw,nosuid,nodev,relatime,errors=remount-ro 0 0
/dev/vde /opt/target ext4 rw,nosuid,nodev,relatime,errors=remount-ro,emergency_ro 0 0
/dev/vdf /root/.config/herdr ext4 rw,nosuid,nodev,relatime,errors=remount-ro 0 0
lns-bind-0 /root/.claude virtiofs rw,nosuid,nodev,relatime 0 0
tmpfs /run tmpfs rw,nosuid,nodev,relatime,size=65536k,mode=755 0 0
```

Write probes in the guest, showing the whole overlay root is gone and only the tmpfs survives:

```
/tmp         cannot touch '/tmp/.probe': Read-only file system
/var/tmp     cannot touch '/var/tmp/.probe': Read-only file system
/root        cannot touch '/root/.probe': Read-only file system
/etc         cannot touch '/etc/.probe': Read-only file system
/usr/local   cannot touch '/usr/local/.probe': Read-only file system
/run         WRITABLE
```

Host-side `e2fsck -fn` (e2fsprogs 1.47.4) with the service stopped and nothing holding the images — this is the part that points at the read path:

```
════════ lns-dev-work (vdc — took emergency_ro) ════════
Pass 1: Checking inodes, blocks, and sizes
Pass 2: Checking directory structure
Pass 3: Checking directory connectivity
Pass 4: Checking reference counts
Pass 5: Checking group summary information
lns-vol: 99433/655360 files (0.3% non-contiguous), 1685282/2621440 blocks

════════ lns-dev-target (vde — first to fail) ════════
… Pass 1–5 clean …
lns-vol: 10488/655360 files (2.0% non-contiguous), 1066204/2621440 blocks

════════ upper.img (vda — second to fail) ════════
Warning: skipping journal recovery because doing a read-only filesystem check.
… Pass 1–5 clean …
lns-upper: 14790/655360 files (0.1% non-contiguous), 140291/2621440 blocks
```

No errors on any image. The only findings anywhere were stale free-block and free-inode counts on the two journalled images, which is the expected appearance of a journalled filesystem taken down uncleanly and read without replaying its journal.

Secondary observation, probably a consequence rather than a cause: while the box was in this state the VM process sat at ~300% CPU with nothing being asked of it, which is consistent with processes retry-looping against the read-only root. `lns-service` itself was at 0.0%. One of the two errors names `Bun Pool 4`, a thread from an agent runtime, so the reads that trip this are ordinary workload activity rather than anything exotic.

jansav:
Fixes #247.

## The bug

A long-lived sandbox takes `EXT4-fs error` on its block devices and remounts read-only, while `e2fsck -fn` against the same images on the host passes clean. Two signatures, both read-side:

```
EXT4-fs error (device vde): ext4_lookup:1789: inode #114691: comm cargo:
  deleted inode referenced: 164044
EXT4-fs error (device vdc): ext4_readdir:262: inode #246521: block 1011775:
  comm gix::dirwalk::i: bad entry in directory: rec_len is smaller than minimal -
  offset=0, inode=0, rec_len=0, size=4096 fake=0
```

One is a stale read, the other an all-zero directory block.

## Cause

Every disk was attached with `initWithURL:readOnly:error:`, which leaves `cachingMode` at `.automatic`. On local storage `.automatic` selects **uncached** host I/O, and an uncached read races APFS delayed writeback into a sparse hole: it sees the on-disk state, which is the old content, or zeroes where the hole was never written. Our images are hand-built sparse ext4 (10 GiB apparent, `write_ext4` writes metadata only), so every unwritten data block is an APFS hole.

That single mechanism produces both signatures — zeroes where the page had never been written, stale content where it had.

UTM shipped `cachingMode: .cached` for the same class of symptom in [utmapp/UTM#5919](https://github.com/utmapp/UTM/pull/5919).

## The change

- New `crates/lns-service/src/vm/vz/attachment.rs` — `policy(DiskRole) -> AttachmentPolicy`, returning `.cached`, `.full`, and the per-role `read_only` flag. Host-tested at 100%.
- `ffi.rs` collapses its three attachment sites into one `disk_attachment(url, role)` helper on the 5-arg initializer. The FFI wall keeps the call; the decision moved out of it.
- `synchronizationMode: .full` is **unchanged behaviour** — Apple documents `.full` as the 3-arg initializer's default. It is now explicit and pinned by a test, so guest journal barriers stay backed by `F_FULLFSYNC`.
- Uniform across all three roles. The read-only composefs descriptor is host-written just before boot, so its pages can sit in pending writeback and race the same way.
- The `ffi.rs` coverage IGNORES reason now names where the policy is tested.

macOS-only path; `vm/vz.rs` is `#[cfg(target_os = "macos")]`, so nothing changes for Cloud Hypervisor or CI's Linux coverage run.

## Tests

Written first, and failed for the stated reason before the fix:

```
assertion left != right failed: UpperRw must state a caching mode; .automatic
picks uncached host I/O, which serves stale-or-zero reads from a sparse image
```

Three Layer 3 tests pin caching, durability, and the read-only flag per role. Review caught that the descriptor's mode was not pinned — a regression to `.uncached` would have passed — so the caching test now covers all three roles. Mutation-proved: flipping `policy()` to `.uncached` fails it, `.cached` passes.

`make lint`, `make complexity`, `make coverage` all exit 0. Coverage: `crates/lns-service/src/vm/vz/attachment.rs: 100.00% (12/12)` — no IGNORES entry.

## Validation

Draft because the evidence is upstream precedent plus mechanism, not a local repro.

Two directed synthetic attempts failed to reproduce (details in the [issue comment](https://github.com/lensapp/lens-sandbox/issues/247#issuecomment-5315360946)): forced `readdir` through `drop_caches`, then paced hole-filling with writeback deliberately in flight. Both clean for their full windows. Real `cargo build` reproduces where a file-creation loop does not, most likely because rustc reads its metadata through `mmap` — a different read path — and churns directory blocks far harder via rename and hardlink traffic.

So validation is a soak on the fixed build. The sharpest signal is the overlay upper: `upper.img` is created fresh per run and is therefore 100% holes, which is why one reproduction landed about a minute after a restart. Each `lns run` is one trial in the worst-case window. Note that an already-filled volume is a weak test — its hole surface is spent, so it faults less regardless of this change.

Ready to un-draft once the soak has run.


jansav:
Soak result: no recurrence since switching the devbox to this branch. Marking ready for review.

What that does and does not establish, so the next reader can judge it:

- **The overlay upper is the meaningful evidence.** `upper.img` is created fresh per run and is therefore 100% holes, which is the worst case for this fault — one reproduction landed about a minute after a restart, writing to `/tmp`. Every `lns run` since has been a trial in that window, and none faulted.
- **The named volumes are not evidence.** Their hole surface was already largely consumed before the fix, so they would fault less regardless of this change.
- **This cannot distinguish "fixed" from "much rarer".** The pre-fix failures were 35 minutes and ~1 minute into a session, so a clean stretch of ordinary development is a real signal, but it is not a proof. A fresh-volume soak would sharpen it.

The `.cached` reasoning stands on the mechanism plus [utmapp/UTM#5919](https://github.com/utmapp/UTM/pull/5919); the soak is confirmation, not the argument.

chenhunghan (approved):

PR #251, opened 2026-08-17
feat!: let a fileset travel inside the document that declares it

jansav:
A fileset was its own publishable artifact: `lns push` packed each `spec.filesets[].path` directory into a `kind: fileset` artifact, uploaded it beside the sandbox, and rewrote the entry to a digest-pinned `ref`.

`docs/sandbox-spec.md` decided otherwise, and already said so: §3.1.11 states a fileset is **not a separate artifact**, §6 packs a `path` into a layer of the artifact the document configures, and §7's kind table names only `sandbox`, `connector` and `mixin`. This closes that divergence. Files and the declaration that mounts them now share one digest, so a consumer approves them together. The specification is untouched.

## What changed

- **`lns-artifact`** — deleted `Kind::FileSet`, `spec::FileSet`, `parse_fileset`, `build_fileset`, the `vnd.lens.fileset.*` media types, the `filesets[].ref` field, and the `lens.dev/v1alpha1` config parser that existed only to read a fileset config. `build_artifact(doc, path_filesets)` packs one deterministic tar layer per `path` entry in declaration order and refuses a layer count that disagrees with the document.
- **`lns-cli`** — `push` packs into the one artifact; `Producer::push_prebuilt` is gone. The entry publishes as written, keeping its `path`.
- **`lns-service`** — a pulled document's `path` fileset materializes from the layer of the artifact that shipped it; a local definition's or a directory mixin's is read from disk. Filesets still work through mixins exactly as before.

## Where the provenance went

Removing `ref` means a merged document no longer records *which* document shipped each fileset. The merge now records each surviving `path` entry's origin — source label plus layer index — and it travels beside the merged bytes as `lns_ipc::FilesetOrigin`, for the same reason `contributions` does. This is an IPC field, not a new field in the `lns.yaml` format.

Routing is fail-closed: only a digest-pinned origin is fetched. Every other label — the decisions file's own name, a directory mixin's path, a floating tag, a truncated digest — names a document on this machine and is read from disk. `pull_fileset_layer` additionally refuses an unpinned source before it touches the registry.

## Verification

`make lint`, `make complexity` and `make coverage` pass; every file in the diff is at 100% and no IGNORES entry was added. 407/407 lns-cli and 229/229 lns-service Layer 2 scenarios pass, plus 2189 lns-service and 232 lns-artifact unit tests.

Two caveats for reviewers:

- **`make e2e` has not been run** — it needs real binaries and a live registry. The Layer 1 features and steps are updated and compile, but CI's e2e job is their first real execution.
- The local test host runs as uid 0 with no network, so 32 tests in modules this branch does not touch (`update`, `kernel`, `composefs`, `content_store`, `host_bind`, `supervisor/adapter`, lns-init `mount`, `coverage-strip-ast`) fail there and account for the only coverage-gate FAILs. They assert permission and network failures that do not occur as root. CI should be the arbiter.

## Breaking

`spec.filesets[].ref` is removed and no fileset publishes as an artifact of its own. Re-push any sandbox or mixin that declared a fileset by `ref`.


jansav:
Closing in favour of #249, which makes the same specification change and is the more complete of the two.

- #249 is green on every gate, including `make e2e`. This branch fails it: the two scenarios flagged as unrun in the description do fail, because a push still publishes a repository beside the sandbox for a fileset sandbox and an inline sandbox.
- #249 also removes the top-level `mount` key, verifies the pulled blob digest against the manifest, and refuses a `path` entry that has no layer behind it.

Two ideas from this branch are worth carrying over to #249: plain tar layers instead of a new `flate2` dependency, and the fail-closed rule that fetches only a digest-pinned origin and reads every other label from disk.

PR #257, opened 2026-08-18
feat(cli): web-based login by default for lns login

chenhunghan:
## Summary

Before this change, a login to the Lens hub had two manual steps. The user made a token on the hub website. The user then sent the token with `echo 'lnshub_pat_…' | lns login -u <user> --password-stdin`.

With this change, the command `lns login` with no credential flags starts a browser login. The CLI shows a one-time confirmation code. The CLI opens the browser on the authorize page of the registry. The user approves the login in the browser. The registry then makes a new token for the user account. The CLI sends the token to the service for verification. The CLI keeps the token only after the service accepts it.

The hub has the server side of this flow. This PR adds the CLI side. The wire contract does not change.

### 1. Branch point in `login()` (`crates/lns-cli/src/login/mod.rs`)

- If the user gives no `-u`, `-p`, or `--password-stdin` flag, the CLI uses the browser flow.
- If the user gives one or more of these flags, the CLI uses the flag path. This path does not change.
- The command `lns login -u me` continues to show the error `a password is required…`.
- The commands `lns login --list` and `lns logout` do not change.
- A test with a guard fake makes sure that the flag path never touches the browser flow.

### 2. New module `crates/lns-cli/src/login/web.rs` (100% line coverage)

- The `WebLogin` component controls the flow. It starts the grant, shows the code, opens the browser, and polls the registry.
- If the registry sends `slow_down`, the CLI adds 5 seconds to the poll interval (RFC 8628).
- If the grant is not complete after `expires_in` seconds, the CLI stops and shows an expiry message.
- The `RealDeviceAuthClient` component sends POST requests to `https://{registry}/cli/device/code` and `https://{registry}/cli/device/token`. Each request has a 10-second timeout.
- A 404 response shows that the registry has no browser login. A 503 response gives the reason from the registry, and the CLI shows it.

### 3. Wiring in `crates/lns-cli/src/login/real.rs`

- `RealBrowserOpener` starts `open` on macOS or `xdg-open` on Linux.
- `RealWebLoginFlow` makes the HTTP client for each login and connects it to `run_login`.

### 4. Requests identify the CLI with the standard user agent (`crates/lns-cli/src/login/web.rs`, `crates/lns-ipc/src/user_agent.rs`)

The CDN in front of hub.lns.run answers 403 to a request with no `User-Agent` header. The reqwest client sends no `User-Agent` by default, so every browser login failed at the first request:

```console
$ lns login
Error: starting browser login to hub.lns.run; pass --username/-u and --password-stdin (recommended) or --password

Caused by:
    https://hub.lns.run/cli/device/code answered HTTP 403
```

The device-auth client now sends the same user-agent string as `lns update` and the service update check, with a new `cli-login` method token:

```
lns/0.16.0 (os=Darwin; arch=arm64; kernel=Darwin/24.6.0; shell=zsh; method=cli-login)
```

The `RealUname` + `detect_platform` leaf moved from the update wiring to a shared `crate::platform` module, so login and update identify the host the same way.

## What the user sees

A login that completes:

```console
$ lns login
Your one-time confirmation code is: WXYZ-1234
Opening https://hub.lns.run/cli/authorize in your browser. Enter the code there.
Waiting for you to approve the login in the browser…
To log in with a token instead, pass --username/-u and --password-stdin (recommended) or --password.
You are now logged in to hub.lns.run as <your-username>.
```

A browser that does not open (for example, on a headless machine):

```console
Could not open a browser. Go to https://hub.lns.run/cli/authorize and enter the code.
```

A registry that has no browser login:

```console
$ lns login ghcr.io
ghcr.io does not support browser login; pass --username/-u and --password-stdin (recommended) or --password
```

A login that the user denies in the browser:

```console
login to hub.lns.run was denied in the browser
```

A confirmation code that expires:

```console
the confirmation code expired before the login was approved; run `lns login` again
```

## Docs

- `docs/cli-reference.md`: the plain `lns login` command is now the first example. The flag forms and `--list` stay.
- `docs/running-workloads.md`: one new sentence in the push-auth section.

## Test instructions

1. Start the service. Run `lns login`. Make sure the CLI shows the code and opens the browser. Approve the login in the browser. Make sure the CLI shows `You are now logged in to hub.lns.run as <you>.` and `lns login --list` shows the entry.
2. Run `echo "$TOKEN" | lns login -u <user> --password-stdin ghcr.io`. Make sure the behavior is the same as before this change.
3. Run `lns login -u me hub.lns.run`. Make sure the error is `a password is required…`.
4. Run `lns login ghcr.io` with no flags. Make sure the error tells you that the registry has no browser login and points to `--password-stdin`.
5. Run `lns login` and deny the login in the browser. Make sure the CLI shows the denied error. Make sure the CLI keeps no credential.
6. Run `lns login` and do not approve the login. Wait until the code expires. Make sure the CLI tells you to run `lns login` again.

## Review fixes

The review by @jansav found five problems. Each fix is one commit with a test.

1. **The CLI checks the service before the browser flow starts** (c04b47d6). `RegistryVerifier` has a new `available()` probe. It sends `Request::Ping` over the service socket. If the service is down, `lns login` stops with the `lns service start` message before it shows a code or opens the browser.
2. **The CLI rejects a verification URL that is not `https`** (846166a2). The registry supplies the URL. A URL that does not parse as `https` never reaches `open`/`xdg-open`. The CLI falls back to the "Could not open a browser" message.
3. **The token poll percent-encodes the device code** (10cd0085). The client now uses reqwest's `.form()`, so a code with `+`, `&`, `=`, or `%` reaches the server unchanged.
4. **The poll interval has a floor of one second** (fd183002). A registry that answers `interval: 0` can no longer cause a hot request loop.
5. **The unused `WebLoginOutcome::Failed` variant is deleted** (cabb51b1). No production path constructed it.


crates/lns-cli/src/login/web.rs
```diff
@@ -0,0 +1,721 @@
...
+                200 => Ok(DeviceStart::Started(parse(&bytes, &url)?)),
+                other => bail!("{url} answered HTTP {other}"),
+            }
+        })
+    }
+
+    fn poll<'a>(&'a self, device_code: &'a str) -> LocalBoxFuture<'a, Result<TokenPoll>> {
+        Box::pin(async move {
+            let url = format!("{}/cli/device/token", self.base);
+            let body = format!("device_code={device_code}");
```
jansav: **The form body is not percent-encoded.**

`format!("device_code={device_code}")` writes a registry-supplied value straight into an `x-www-form-urlencoded` body. A code containing `+`, `&`, `=`, or `%` reaches the server as a different string (`+` decodes to a space).

**Why:** The hub's current code alphabet happens to be safe, so this works today and breaks the day the hub changes its token generator — as a poll that never resolves and then reports "unexpected error code". `self.http.post(url).form(&[("device_code", device_code)])` encodes correctly and sets the content-type, replacing both lines.

crates/lns-cli/src/login/web.rs
```diff
@@ -0,0 +1,721 @@
...
+        }
+        writeln!(out, "Waiting for you to approve the login in the browser…")?;
+        writeln!(out, "To log in with a token instead, {FLAG_FALLBACK}.")?;
+        Ok(())
+    }
+
+    async fn wait_for_grant(&self, auth: &DeviceAuthorization) -> Result<WebLoginOutcome> {
+        let expiry = tokio::time::sleep(Duration::from_secs(auth.expires_in));
+        tokio::pin!(expiry);
+        let mut interval = Duration::from_secs(auth.interval);
```
jansav: **The registry controls the poll interval with no floor.**

`interval` is used verbatim. An `interval` of `0` turns `wait_for_grant` into a hot loop that POSTs `/cli/device/token` as fast as the network allows for the whole `expires_in` window.

**Why:** One bad hub deploy turns every waiting CLI into a request flood with no backoff. `Duration::from_secs(auth.interval.max(1))` closes it. The `start_paused` tests can never surface this.

crates/lns-cli/src/login/mod.rs
```diff
@@ -121,15 +146,12 @@ async fn login(
...
     input: &mut dyn std::io::BufRead,
     out: &mut impl Write,
 ) -> Result<i32> {
     let registry = target_registry(args.registry.as_deref(), default_registry)?;
-    let username = args
-        .username
-        .clone()
-        .ok_or_else(|| anyhow!("a username is required; pass --username/-u"))?;
-    let secret = resolve_secret(args, input)?;
+    let (username, secret) = resolve_credentials(args, &registry, web, input, out).await?;
```
jansav, line 157: **The service check happens after the user has already approved in the browser.**

`resolve_credentials` runs the full browser flow, then `verify` discovers the service is down and bails.

**Why:** The user prints a code, approves in the browser, waits — and then gets "start it with `lns service start`" with nothing stored, so the whole flow must be repeated. It also leaves a live token orphaned on the hub. A `Request::Ping` probe before the flow starts costs one round trip.

crates/lns-cli/src/login/real.rs
```diff
@@ -47,6 +48,41 @@ fn configured_default_registry() -> Result<String> {
...
+struct RealBrowserOpener;
+
+impl super::BrowserOpener for RealBrowserOpener {
+    fn open(&self, url: &str) -> bool {
+        let opener = if cfg!(target_os = "macos") {
+            "open"
+        } else {
+            "xdg-open"
+        };
+        std::process::Command::new(opener).arg(url).spawn().is_ok()
```
jansav, line 61: **`open`/`xdg-open` receives an unvalidated registry-supplied string.**

`verification_uri_complete` comes from the registry's JSON and is handed to the platform opener as-is.

**Why:** A hostile or compromised registry host can make `lns login <host>` invoke `open` on an arbitrary string — on macOS that opens local files or launches a registered scheme handler, which pulling an image from the same host cannot do. Rejecting anything that does not parse as an `https` URL before spawning is one line.

crates/lns-cli/src/login/mod.rs
```diff
@@ -68,6 +73,25 @@ pub trait RegistryVerifier {
...
 }
 
+/// The terminal result of driving a browser-based device login against a registry.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum WebLoginOutcome {
+    Completed { username: String, secret: String },
+    Unsupported,
+    Denied,
+    Expired,
+    Failed(String),
```
jansav: **`WebLoginOutcome::Failed` is never constructed in production.**

Only the match arm at `mod.rs:260` and the fake at `mod.rs:805` touch it — `WebLogin` reports every failure as `Err`.

**Why:** A public variant, an error message, and a test that no production path can reach. The test pins the fake, not the product. Delete the variant and its arm.

jansav (commented):
**Problem:** `lns login` needed two manual steps — mint a token on the hub website, then pipe it in. This PR makes a flagless `lns login` run an RFC 8628-style browser device flow against the registry, so the token is minted and stored for you.

**Approach:** A `WebLoginFlow` seam in `login/mod.rs` branches on "no credential flags"; `login/web.rs` holds the orchestrator (`WebLogin` over `DeviceAuthClient` + `BrowserOpener` ports) plus the reqwest client; the issued token still rides the existing verify-then-store path, so a rejected token is never saved. `RealUname`/`detect_platform` moved to a shared `crate::platform` so login and update identify the host identically.

**Product alignment:** Fits "policy you run into, not write" — the secret still never enters the workload, and it stays in the host-side credential store.

**Verdict:** Needs fixes. Five findings inline, all in the "should fix" band — no critical issues, and CI is green on every required job.

crates/lns-cli/src/login/web.rs
chenhunghan: Fixed — `poll` now sends the code via reqwest's `.form(&[("device_code", device_code)])`, which percent-encodes the value and sets the content-type. Added `poll_percent_encodes_a_device_code_with_reserved_characters`, which pins that a code containing `+&=%` reaches the server encoded. (commit 10cd0085)

crates/lns-cli/src/login/web.rs
chenhunghan: Fixed — the interval is now `auth.interval.max(1)` seconds. Added `a_zero_poll_interval_from_the_registry_is_floored_to_one_second`, which pins (under paused time) that the first poll happens at 1s rather than immediately. (commit fd183002)

crates/lns-cli/src/login/mod.rs
chenhunghan: Fixed — `RegistryVerifier` grew an `available()` probe (a `Request::Ping`/`Pong` round trip over the service socket), and a flagless login now bails with the `lns service start` message before the web flow is consulted. Pinned by `a_down_service_stops_a_web_login_before_the_browser_flow_starts`, which uses the panicking `NoWebLogin` guard to prove the ordering. The flag path is unchanged. (commit c04b47d6)

crates/lns-cli/src/login/real.rs
chenhunghan: Fixed — `WebLogin::present` only hands `verification_uri_complete` to the opener when it parses as an `https` URL; anything else falls back to the "Could not open a browser" manual path. The guard lives in web.rs (host-tested; real.rs is production wiring). Added `a_non_https_verification_url_never_reaches_the_browser_opener` with a `file:///etc/passwd` fixture. (commit 846166a2)

crates/lns-cli/src/login/mod.rs
chenhunghan: Fixed — deleted the `Failed` variant, its match arm, and the test that only exercised the fake. (commit cabb51b1)

jansav (approved):

PR #260, opened 2026-08-18
feat(exec): add interactive exec sessions

chenhunghan:
## Summary

`lns exec` could only run pipe-only commands, so shells, REPLs, editors, and terminal applications failed inside an existing run. This PR gives every exec session independently addressable terminal controls while preserving the primary session and concurrent exec sessions.

### 1. Session-addressed IPC

- Adds explicit primary and exec session targets for stdin, stdin close, resize, signal, and detach controls.
- Returns an opaque exec session ID in the streaming handshake.
- Registers exec input channels before acknowledging startup and removes them on exit, detach, disconnect, or run teardown.
- Detects silent client disconnects on the dedicated exec stream so abandoned exec processes are cancelled promptly.

### 2. Explicit terminal modes

`lns exec` remains non-interactive by default:

```text
lns exec RUN command       stdin closed, no PTY
lns exec -i RUN command    stdin forwarded, no PTY
lns exec -t RUN command    PTY allocated, stdin unattached
lns exec -it RUN command   stdin forwarded through a PTY
```

The command separator remains optional. Terminal resize, signals, stdin EOF, and the detach chord target only the exec session that originated them.

### 3. Regression coverage

- Activates CLI and service behavioral scenarios for explicit terminal flags and multi-session isolation.
- Extends the live Expect smoke with two concurrent exec PTYs, Ctrl+C, detach isolation, primary reattach, and asserted cleanup.
- Adds a real-microVM scenario covering stdout, stderr, exit status, and exclusion from primary logs.

## Breaking changes

The local CLI/service IPC contract replaces the run-only control request variants with session-targeted variants and replaces the exec startup `RunStarted` response with `ExecStarted { run_id, session_id }`. The CLI and service are version-locked and must be updated together; no compatibility shim is retained.

## Docs

Updates the CLI reference and running-workloads guide with the explicit exec terminal modes, optional command separator, and exec-only detach behavior.

## Test instructions

1. Start a long-lived named run: `lns run -d --name exec-test <sandbox> sleep 300`. Run `lns exec -it exec-test sh`; expect a live prompt and successful raw terminal input.
2. Open two `lns exec -it exec-test sh` sessions. Enter the detach chord in one; expect that CLI to return 0 while the second exec and primary run remain usable.
3. Run `printf 'hello\n' | lns exec -i exec-test cat`; expect `hello`, clean EOF handling, and exit 0 without a PTY.
4. Run `lns exec exec-test echo hello`; expect non-interactive output and exit 0. Verify `lns logs exec-test` does not contain the exec output.
5. Run `lns exec -it missing-run sh`; expect a clear `no such run` failure and no run creation.
6. Run `expect -f crates/lns-cli/tests/smoke/interactive-shell.exp`; expect the primary and concurrent exec terminal smoke to pass and remove its temporary run.

## Verification

- `CARGO_LOCKED=--locked make lint`
- `make complexity`
- `make coverage-affected`
- `make e2e`
- `LNS_E2E_FEATURE='session streaming verbs' make e2e-microvm`
- `expect -f crates/lns-cli/tests/smoke/interactive-shell.exp`

## Review follow-ups (post-takeover)

- **Silent client disconnect now cancels the guest exec process.** `OpenSession` carries an explicit `dies_with_client` flag (set for every exec, false for the primary); the broker treats EOF on such a session's stream as a hangup (SIGHUP to the child) while the primary session's EOF behavior is unchanged — an abandoned exec no longer outlives its client, and an undelivered detach degrades to the same hangup.
- The service disconnect scenario now drives the production stream driver (`drive_exec_stream`, extracted seam) and asserts the guest task is actually cancelled.
- The CLI exec behaviours now run the real `exec_image` path (`exec_image_on_stream` seam) over an in-memory duplex against a wire-speaking fake service; flag assertions read the request as decoded off the wire. Byte-level stdin forwarding remains pinned by the expect smoke and the microVM scenario.
- Known limitation for a follow-up: EOF cancellation delivers SIGHUP only — an exec'd process that ignores SIGHUP still survives; no escalation to SIGKILL yet.
- Known limitation for a follow-up: `StdinClose` on a PTY-mode exec writes a bare `^D` (0x04) to the PTY, so with a non-empty line buffer the line flushes but no EOF is delivered — `printf 'x' | lns exec -it RUN cat` waits until the session is closed another way. The pipe path (`-i` without `-t`) closes the fd and delivers EOF correctly.

- **Confined exec sessions now adopt the workload user's identity.** `lns-init` exports `LENS_RUN_HOME` from the same guest-passwd read that resolves `LENS_RUN_UID`/`LENS_RUN_GID`; the broker fills `HOME` and `USER` into a confined session's env unless the session already declares them (runc's rule — `/` when the lookup failed) and starts a workdir-less exec in the effective home, mirroring the supervisor's workdir → home rule. Previously an exec landed with the broker's kernel-given `HOME=/` and no `USER`, so `~` expanded to `//` and config-reading tools looked in the wrong place. Pinned by broker/init unit tests and a new `@microvm` streaming scenario.

- **Exec identity hardened after an adversarial review.** uid/gid/home now come from the one passwd line that supplies the run ids (a sibling line the id parse rejected could hand the exec a different identity's home); a home with control bytes is never exported (a NUL would panic `std::env::set_var` in PID 1 — hostile-image boot DoS); image `ENV HOME`/`USER` no longer outrank the run-as identity in an exec (the run records the author-declared identity vars and the host strips the rest, matching the supervisor's `LENS_SANDBOX_WORKLOAD_*` precedence); the identity-home cwd fallback is chdir-only so the broker never runs mkdir as guest root on a passwd-controlled path; and the broker's spec assembly is a pure host-tested function so identity wiring mutations can't hide in the coverage-exempt platform leaf. Known follow-ups (documented, out of scope): supplementary groups, the root-workload proxy-env question, and the USER-by-name vs USER-by-uid oracle divergence with lens-sandbox-core.

### DIFF
--- a/crates/lns-cli/src/service/orchestrator.rs
+++ b/crates/lns-cli/src/service/orchestrator.rs
@@ -578,14 +578,13 @@ pub async fn exec_image(args: ExecArgs) -> Result<i32> {
     };
     let detach_chord = args.detach_keys.0.clone();
 
-    let request = Request::ExecImage(ExecImageArgs {
-        run: target_run,
-        argv: args.cmd,
-        env: Vec::new(),
+    let request = Request::ExecImage(build_exec_request(
+        target_run,
+        args.cmd,
         tty,
         stdin,
         initial_winsize,
-    });
+    ));
     let frame = encode_frame(&request).context("encoding ExecImage request")?;
     stream
         .write_all(&frame)
@@ -610,6 +609,23 @@ pub async fn exec_image(args: ExecArgs) -> Result<i32> {
         StdinForwarding::of(stdin),
     )
     .await
+}
+
+pub fn build_exec_request(
+    run: String,
+    argv: Vec<String>,
+    tty: bool,
+    stdin: bool,
+    initial_winsize: Option<(u16, u16)>,
+) -> ExecImageArgs {
+    ExecImageArgs {
+        run,
+        argv,
+        env: Vec::new(),
+        tty,
+        stdin,
+        initial_winsize,
+    }
 }
 
 fn decode_exec_started(bytes: &[u8]) -> Result<lns_ipc::SessionTarget> {
--- a/crates/lns-cli/src/service/orchestrator.rs
+++ b/crates/lns-cli/src/service/orchestrator.rs
@@ -669,7 +669,7 @@ fn decode_exec_started(bytes: &[u8]) -> Result<lns_ipc::SessionTarget> {
     }
 }
 
-/// Whether host stdin reaches the session's workload.
+/// Whether host stdin reaches the session's workload; a session opened without `-i` still watches for the detach chord but hands the run nothing.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum StdinForwarding {
     ToRun,
@@ -684,14 +684,6 @@ impl StdinForwarding {
             Self::Withheld
         }
     }
-}
-
-fn should_pump_stdin(stdin: StdinForwarding) -> bool {
-    stdin == StdinForwarding::ToRun
-}
-
-fn should_enable_raw_mode(tty: bool, stdin: StdinForwarding) -> bool {
-    tty && should_pump_stdin(stdin)
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
@@ -826,7 +818,7 @@ where
     O: AsyncWriteExt + Unpin,
     E: AsyncWriteExt + Unpin,
 {
-    let _raw_guard = if should_enable_raw_mode(tty, stdin) {
+    let _raw_guard = if tty {
         crate::raw_mode::RawModeGuard::enable_if_tty()
     } else {
         None
@@ -863,18 +855,16 @@ where
                 let s = socket.clone();
                 tokio::spawn(async move { run_winsize_forwarder(s, winsize_target).await })
             });
-            let stdin_task = should_pump_stdin(stdin).then(|| {
-                let pump_chord = detach_chord;
-                let pump_early_exit = early_exit_tx.clone();
-                tokio::spawn(async move {
-                    let r = run_stdin_pump(socket, target, pump_chord, detach, stdin).await;
-                    if matches!(&r, Ok(true)) {
-                        let _ = pump_early_exit.send(());
-                    }
-                    r
-                })
+            let pump_chord = detach_chord;
+            let pump_early_exit = early_exit_tx.clone();
+            let stdin_task = tokio::spawn(async move {
+                let r = run_stdin_pump(socket, target, pump_chord, detach, stdin).await;
+                if matches!(&r, Ok(true)) {
+                    let _ = pump_early_exit.send(());
+                }
+                r
             });
-            (Some(cancel_task), winsize, stdin_task)
+            (Some(cancel_task), winsize, Some(stdin_task))
         }
         None => (None, None, None),
     };
@@ -2458,15 +2448,6 @@ mod tests {
             );
             assert!(matches!(control, PumpControl::Continue));
         }
-    }
-
-    #[test]
-    fn a_session_opened_without_stdin_does_not_read_host_input() {
-        assert!(!should_pump_stdin(StdinForwarding::Withheld));
-        assert!(should_pump_stdin(StdinForwarding::ToRun));
-        assert!(!should_enable_raw_mode(true, StdinForwarding::Withheld));
-        assert!(should_enable_raw_mode(true, StdinForwarding::ToRun));
-        assert!(!should_enable_raw_mode(false, StdinForwarding::ToRun));
     }
 
     #[test]
--- a/crates/lns-cli/src/service/orchestrator.rs
+++ b/crates/lns-cli/src/service/orchestrator.rs
@@ -560,7 +560,7 @@ pub async fn run_image(
 
 pub async fn exec_image(args: ExecArgs) -> Result<i32> {
     if args.cmd.is_empty() {
-        anyhow::bail!("lns exec requires a command after `--`");
+        anyhow::bail!("lns exec requires a command");
     }
 
     let socket = super::socket_path()?;
@@ -569,7 +569,7 @@ pub async fn exec_image(args: ExecArgs) -> Result<i32> {
         .with_context(|| format!("connecting to {}", socket.display()))?;
 
     let target_run = args.run;
-    let tty = args.tty && crate::raw_mode::stdin_is_tty();
+    let tty = args.tty;
     let stdin = args.interactive;
     let initial_winsize = if tty {
         crate::raw_mode::host_winsize()
@@ -595,24 +595,31 @@ pub async fn exec_image(args: ExecArgs) -> Result<i32> {
     let bytes = read_frame_bytes_async(&mut stream)
         .await
         .context("reading RunStarted frame")?;
-    let run_id = match decode_frame(&mut &bytes[..]).context("decoding RunStarted")? {
-        Response::RunStarted { run_id } => run_id,
-        Response::Error { message } => anyhow::bail!("daemon error: {message}"),
-        other => anyhow::bail!("expected RunStarted, got {other:?}"),
-    };
-    crate::log::debug!(run_id = %run_id, "exec session opened");
+    let target = decode_exec_started(&bytes)?;
+    crate::log::debug!(run_id = %target.run_id(), "exec session opened");
 
-    drive_attached_session(
+    drive_targeted_session(
         stream,
         Some(socket),
-        run_id,
+        target,
         tty,
         detach_chord,
-        DetachBehaviour::SignalAndDrain,
+        DetachBehaviour::CloseSession,
+        CancelBehaviour::SignalSession,
         args.quiet,
         StdinForwarding::of(stdin),
     )
     .await
+}
+
+fn decode_exec_started(bytes: &[u8]) -> Result<lns_ipc::SessionTarget> {
+    match decode_frame(&mut &bytes[..]).context("decoding ExecStarted")? {
+        Response::ExecStarted { run_id, session_id } => {
+            Ok(lns_ipc::SessionTarget::Exec { run_id, session_id })
+        }
+        Response::Error { message } => anyhow::bail!("daemon error: {message}"),
+        other => anyhow::bail!("expected ExecStarted, got {other:?}"),
+    }
 }
 
 /// Whether host stdin reaches the session's workload; a session opened without `-i` still watches for the detach chord but hands the run nothing.
@@ -635,8 +642,15 @@ impl StdinForwarding {
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum DetachBehaviour {
     SignalAndDrain,
+    CloseSession,
     LeaveRunning,
     DetachRun,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum CancelBehaviour {
+    CancelRun,
+    SignalSession,
 }
 
 #[allow(clippy::too_many_arguments)] // one session-shape argument per wire field; the writer seam lives in the _with_writers twin
@@ -653,16 +667,46 @@ pub(crate) async fn drive_attached_session<S>(
 where
     S: AsyncReadExt + AsyncWriteExt + Unpin + Send + 'static,
 {
-    let mut stdout = tokio::io::stdout();
-    let mut stderr = tokio::io::stderr();
-    drive_attached_session_with_writers(
+    drive_targeted_session(
         stream,
         aux_socket,
-        run_id,
+        primary_target(run_id),
+        tty,
+        detach_chord,
+        detach,
+        CancelBehaviour::CancelRun,
+        quiet,
+        stdin,
+    )
+    .await
+}
+
+#[allow(clippy::too_many_arguments)]
+async fn drive_targeted_session<S>(
+    stream: S,
+    aux_socket: Option<PathBuf>,
+    target: lns_ipc::SessionTarget,
+    tty: bool,
+    detach_chord: Vec<u8>,
+    detach: DetachBehaviour,
+    cancel: CancelBehaviour,
+    quiet: bool,
+    stdin: StdinForwarding,
+) -> Result<i32>
+where
+    S: AsyncReadExt + AsyncWriteExt + Unpin + Send + 'static,
+{
+    let mut stdout = tokio::io::stdout();
+    let mut stderr = tokio::io::stderr();
+    drive_targeted_session_with_writers(
+        stream,
+        aux_socket,
+        target,
         tty,
         std::io::stdout().is_terminal(),
         detach_chord,
         detach,
+        cancel,
         &mut stdout,
         &mut stderr,
         quiet,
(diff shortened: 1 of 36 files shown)


crates/lns-service/src/ipc/adapter.rs
```diff
@@ -737,20 +749,14 @@ async fn handle_exec(mut stream: UnixStream, args: lns_ipc::ExecImageArgs) -> an
...
-    let outcome =
-        pump_responses(&mut stream, &mut frame_rx, dead_cancel_rx, dead_detach_rx).await?;
+    let outcome = pump_exec_responses(&mut stream, &mut frame_rx).await?;
     if let PumpOutcome::WriteFailed(e) = &outcome {
         log::debug!(error = %e, "exec stream write failed; tearing session down");
     }
 
-    drop(input_keepalive);
+    crate::run_registry::deregister_exec_session(&target_run_id, &session_id);
     session_task.abort();
```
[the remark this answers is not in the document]
chenhunghan: Fixed guest-side, which covers both concerns at once: the broker now treats EOF on a confined session's stream as a hangup — `eof_action` decides it in `session.rs` and both client loops deliver SIGHUP to the exec child before exiting, so `forker.wait` returns and the session tears down instead of waiting on a long-running child. The primary (Inherit) session keeps today's EOF behavior. Because the hangup now comes from the stream vanishing itself, the abort-vs-queued-detach race is moot — a detach that never got delivered degrades to the same SIGHUP via EOF. Pinned by `host_eof_hangs_up_a_confined_sessions_child` / `host_eof_leaves_an_inherited_sessions_workload_running`. (commits b8b907b8, 7a4c5710)

crates/lns-service/tests/behaviours/steps/exec_session_routing.rs
```diff
@@ -0,0 +1,266 @@
...
+
+#[when("the first exec client disconnects unexpectedly")]
+fn first_exec_disconnects(world: &mut BehaviourWorld) {
+    let target = world.exec.first_target.as_ref().expect("first exec target");
+    let lns_ipc::SessionTarget::Exec { run_id, session_id } = target else {
+        unreachable!()
+    };
+    assert!(lns_service::run_registry::deregister_exec_session(
+        run_id, session_id
+    ));
```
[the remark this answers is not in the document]
chenhunghan: Fixed — the host teardown is now an extracted production seam, `lns_service::ipc::drive_exec_stream` (pump → deregister → cancel guest task), and `handle_exec` calls it. The step drops the client half of an in-memory duplex and drives that seam; the scenario asserts the session's guest task was actually cancelled (a drop-probe inside the spawned task must fire) as well as deregistered, so removing either the abort or the deregistration fails the scenario. (commit 7a4c5710)

crates/lns-cli/tests/behaviours/steps/interactive_exec.rs
```diff
@@ -0,0 +1,142 @@
...
+        return;
+    }
+    let args: lns_cli::cli::ExecArgs =
+        lns_cli::command::parse_args(tokens).expect("exec scenario argv should parse");
+    if !world.exec.active {
+        world.result = Some(CliRun {
+            exit_code: 1,
+            output: format!("no such run: {}", args.run),
+        });
+        return;
```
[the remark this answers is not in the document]
chenhunghan: Fixed — exec now has the same seam the run path already had (`exec_image_on_stream`, mirroring `drive_attached_session_with_writers`), and the scenarios run the real CLI path over an in-memory duplex against a wire-speaking fake service: real clap parse, real ExecImage encode, real ExecStarted/Error decode, real session-stream rendering and exit-code propagation. Flag assertions now read the `ExecImageArgs` the fake decoded off the wire instead of a fabricated copy, and the inactive-run scenario asserts the daemon refusal surfaced with no `ExecStarted` handshake. Live keystroke forwarding through a real PTY can't be exercised in-process (the stdin pump reads the real process stdin), so the feature header now states explicitly that it's pinned by the interactive-shell expect smoke and the microVM exec scenario. (commit fe54723a)

chenhunghan:
@copilot+claude-fable-5:medium resolve the merge conflicts in this pull request

crates/lns-cli/src/service/orchestrator.rs
```diff
@@ -668,27 +810,45 @@ where
...
             });
-            let pump_chord = detach_chord;
-            let pump_early_exit = early_exit_tx.clone();
-            let stdin_task = tokio::spawn(async move {
-                let r = run_stdin_pump(socket, run_id, pump_chord, detach, stdin).await;
-                if matches!(&r, Ok(true)) {
-                    let _ = pump_early_exit.send(());
-                }
-                r
+            let stdin_task = should_pump_stdin(stdin).then(|| {
```
jansav: **`lns run --interactive=false` silently loses the detach chord.**

The stdin pump is now spawned only when `stdin == ToRun`. On `origin/main` it was spawned unconditionally, so a non-interactive attached run could still be detached with `Ctrl-P Ctrl-Q`. The doc comment that promised this was deleted, but `--detach-keys`' own help text still promises it, and `a_session_opened_without_stdin_still_honours_the_detach_chord` now pins a code path no caller reaches.

**Why:** `lns run --interactive=false <sandbox>` becomes undetachable — Ctrl-C cancels the run instead. This is unrelated to exec and is not listed under Breaking changes.

crates/lns-cli/tests/behaviours/steps/interactive_exec.rs
```diff
@@ -0,0 +1,194 @@
...
+    let request = request(world)?;
+    if request.stdin && request.tty {
+        Ok(())
+    } else {
+        Err("interactive PTY mode was not requested".into())
+    }
+}
+
+#[then("the user receives a live shell prompt")]
+fn live_shell_prompt(world: &mut BehaviourWorld) -> Result<(), String> {
```
jansav: **Three scenarios claim behaviour these steps do not check.**

`the user receives a live shell prompt`, `raw-mode terminal programs can run`, and `terminal output is displayed live` all delegate to one assertion: the wire request had `tty && stdin`. Three different promises, one flag check.

**Why:** The scenario names read as live-PTY coverage in the coverage-bearing layer. A regression in raw-mode entry, CRLF handling, or output flushing leaves all three green. Either rename them to what they pin, or drop them and let the expect smoke carry the claim.

crates/lns-service/tests/behaviours/steps/exec_session_routing.rs
```diff
@@ -0,0 +1,306 @@
...
+}
+
+#[given("two interactive exec sessions are active")]
+fn two_exec_sessions(world: &mut BehaviourWorld) {
+    register_exec(world, "exec-1", true);
+    register_exec(world, "exec-2", false);
+}
+
+#[when(regex = r#"^the user runs \"(lns exec(?: [^\"]*)?)\"$"#)]
+fn user_runs(world: &mut BehaviourWorld, _command: String) {
```
jansav, line 84: **The `exec works while the primary session is attached` scenario drives no production code.**

This step ignores the command string, calls `register_exec_session` directly, and hard-codes `response = Acknowledged`. Its `Then` asserts `first_target.is_some()` — that the fixture it just built exists.

**Why:** The one scenario covering exec alongside an attached primary cannot fail. Drive it through `handle_request` / the exec handshake like the sibling scenarios do.

crates/lns-cli/src/service/orchestrator.rs
```diff
@@ -668,27 +810,45 @@ where
...
-            let cancel_run_id = run_id.clone();
-            let cancel = tokio::spawn(async move {
+            let cancel_target = target.clone();
+            let cancel_socket = socket.clone();
+            let cancel_task = tokio::spawn(async move {
                 let _ = tokio::signal::ctrl_c().await;
-                cancel_client.cancel_run(cancel_run_id).await;
+                match cancel {
+                    CancelBehaviour::CancelRun => {
+                        if let Ok(client) = real_client() {
```
jansav: **Ctrl-C now swallows a client-construction failure.**

`real_client()?` was propagated before; it is now `if let Ok(client) = real_client()` inside the spawned task.

**Why:** If the client cannot be built, Ctrl-C on an attached run does nothing and reports nothing — the user presses Ctrl-C and the run keeps going with no explanation.

crates/lns-session-broker/src/session.rs
```diff
@@ -104,15 +104,25 @@ pub(crate) fn validate_open_session(frame: &ClientFrame) -> Result<(), SessionEr
...
     WriteStdin(Vec<u8>),
+    CloseStdin,
     Resize(Winsize),
     Signal(i32),
     Detach,
     Stop,
 }
 
+/// A confined session exists only for its host client, so a vanished stream hangs its child up; the inherited primary outlives any one connection.
+pub(crate) fn eof_action(confinement: &Confinement) -> LoopAction {
```
jansav: **Session lifetime is keyed off a security property.**

`eof_action` decides "hang up the child on host EOF" from `Confinement`. It is correct only because the primary session happens to pass `confine: !SUPERVISED`, and `SUPERVISED` is `true`.

**Why:** Flipping `SUPERVISED`, or confining a primary session for any other reason, makes every dropped CLI connection SIGHUP the run's workload. An explicit "this session dies with its client" field on `OpenSession` costs one line and cannot be broken from a distance.

crates/lns-service/src/ipc/mod.rs
```diff
@@ -526,8 +576,73 @@ async fn wait_for_exit(run_id: &str, timeout: std::time::Duration) -> bool {
...
+                target,
+                session_input_from_resize(*rows, *cols),
+                "SessionResize",
+            )
+            .await
+        }
+        Request::SessionSignal { target, signal } => {
+            forward_target_input(target, session_input_from_signal(*signal), "SessionSignal").await
+        }
+        _ => unreachable!("handle_session_control only accepts session control requests"),
```
jansav: **`handle_session_control` re-matches what `handle_request` already matched.**

The five session variants are matched in `handle_request`, collapsed into one call, then matched again — which forces this `unreachable!` and a `#[should_panic]` test to cover it. Dispatching from the original arms removes both.

crates/lns-session-broker/src/session/real.rs
```diff
@@ -363,14 +366,29 @@ fn drive_session_pipes(
...
+            }
+            break;
+        };
         match dispatch_frame(frame) {
             LoopAction::WriteStdin(bytes) => {
                 if !vsock::write_all(master_in, &bytes) {
                     break;
                 }
             }
+            LoopAction::CloseStdin => {
```
jansav, line 404: **Optional:** `StdinClose` on a PTY writes a bare `0x04`. `printf 'x' | lns exec -it RUN cat` sends `^D` with a non-empty line buffer: the line flushes, no EOF is delivered, and the client waits forever. The pipe path (`close(fd)`) is correct; only the PTY path has this. Worth listing next to the SIGHUP-only limitation the description already records.

jansav (commented):
**Problem:** `lns exec` could only run pipe-only commands, because the IPC control verbs addressed a run and not a session, so keystrokes had nowhere to go. This PR adds a session-addressed control plane and explicit `-i`/`-t` exec modes.

**Approach:** The `Request::Run*` control variants become `Request::Session*` carrying a `SessionTarget` enum. The service assigns an opaque exec session id in a new `ExecStarted` handshake and keeps a per-run `exec_sessions` map. The broker gains `StdinClose`, and hangs up a confined session's child when its host stream vanishes. Breaking, with no shim — correct for pre-1.0.

**Product alignment:** Fits the docker-shaped CLI and `docs/running-workloads.md`; the docs land in the same range.

**Verdict:** Needs fixes. Six findings below, plus one optional. The routing design is sound — the fixes are a behavioural regression outside exec's scope, three tests that do not pin what they claim, and two coupling issues.

crates/lns-cli/src/service/orchestrator.rs
chenhunghan: Fixed — restored main's shape: the stdin pump spawns unconditionally again and raw mode is gated on `tty` alone, so a `--interactive=false` attached run detaches with the chord while `run_stdin_pump` still withholds the typed bytes from the run. The deleted doc comment on `StdinForwarding` is restored, and the helper test that pinned the wrong behavior went red first and was replaced by the reachable `plan_feed` chord pin. (commit b1cd34bb)

crates/lns-cli/src/service/orchestrator.rs
chenhunghan: Fixed — the CancelRun client is constructed before the task spawns and its failure propagates with `?` again, so a broken client construction fails the attach up front instead of leaving Ctrl-C silently dead. The SignalSession arm needs no client and is unchanged. No dedicated test: `dirs::data_dir()` can't be made to fail deterministically on a host without a client-factory port — same coverage as before the regression. (commit dd7a07d7)

crates/lns-session-broker/src/session.rs
chenhunghan: Fixed — `OpenSession` now carries an explicit `dies_with_client` flag and `eof_action` keys off it alone; confinement no longer decides session lifetime. The host sets it true for execs (pinned by `build_session_params_declares_the_exec_dies_with_its_client`) and false for the primary and capture sessions, so flipping `SUPERVISED` or confining a primary can no longer SIGHUP a run's workload from a distance. (commit fa75c189)

crates/lns-cli/tests/behaviours/steps/interactive_exec.rs
chenhunghan: Fixed — dropped the three alias steps; the scenario is renamed to "interactive TTY mode requests a PTY with host stdin forwarded" and keeps the single wire-contract assertion it actually makes. The live-prompt / raw-mode / live-output claims stay with the expect smoke and the microVM scenario. (commit ea500113)

crates/lns-service/tests/behaviours/steps/exec_session_routing.rs
chenhunghan: Fixed — the When now drives the production handshake (`run_registry::resolve` + `ipc::register_exec_input`, exposed as the handshake seam) and the Thens route stdin through `handle_request` to both targets: the exec must receive its own bytes with the primary untouched, and the primary must stay attached (live detach channel) and still receive its own input. Mutation-checked: disabling `register_exec_input` fails the scenario. (commit 454738e8)

crates/lns-service/src/ipc/mod.rs
chenhunghan: Fixed — the five session-control variants dispatch directly from `handle_request`'s own arms; `handle_session_control`, its `unreachable!`, and the `#[should_panic]` test are gone. (commit b5def258)

crates/lns-session-broker/src/session/real.rs
chenhunghan: Agreed — recorded next to the SIGHUP-only limitation in the description: PTY-path `StdinClose` writes a bare `0x04`, so a non-empty line buffer flushes without EOF and `printf 'x' | lns exec -it RUN cat` waits; the pipe path closes the fd and is correct. Left as a follow-up alongside the SIGHUP→SIGKILL escalation.

jansav (approved):
Re-reviewed at `454738e8`. All six findings are fixed, and fixed at the root rather than papered over.

1. **Detach chord** (`b1cd34bb`) — the stdin pump and raw-mode-on-tty are unconditional again, the two dead helpers and their test are gone, and the doc comment that promised the behaviour is back.
2. **Lying CLI steps** (`ea500113`) — all three deleted; the scenario is renamed to `interactive TTY mode requests a PTY with host stdin forwarded`, which is what it checks.
3. **Tautological service scenario** (`454738e8`) — now drives `register_exec_input`, routes stdin through `handle_request` to both the exec and the primary, and asserts the primary's detach channel was never consumed. It can fail now.
4. **Swallowed Ctrl-C error** (`dd7a07d7`) — the match is split so `CancelRun` constructs the client with `real_client()?` before spawning.
5. **Lifetime keyed off `Confinement`** (`fa75c189`) — explicit `dies_with_client` on `OpenSession`, threaded to both client loops, `true` only for exec. `eof_action` no longer sees `Confinement`. This is the fix I hoped for.
6. **`unreachable!`** (`b5def258`) — dispatch moved into `handle_request`'s own arms; the helper and its `#[should_panic]` test are gone.

Finding 7 (`0x04` as PTY EOF) stays as agreed, alongside the SIGHUP-only limitation in the description.

One leftover, not worth blocking on: `#[derive(Clone)]` on `Confinement` (`crates/lns-session-broker/src/session.rs:37`) was added only to clone it into the client loops, and nothing clones it any more.

**Verdict:** Ship it. ✅

PR #262, opened 2026-08-18
feat(sandbox): Decide per machine which of its files a pulled sandbox reads

jansav:
## Summary

`filesets[].hostPath` names one file on the machine that *runs* a document, and `optional` says a missing source is skipped rather than fatal. Both ship. Neither appeared in `docs/sandbox-spec.md`, and both structs decode strictly — so a document written against the specification parsed, and one written against the code parsed nowhere else. The register called this a `REMOVE`.

Deleting them is the wrong fix. `hostPath` is the *narrow* way to seed a guest from the machine: one file, snapshotted at launch, never packed, and the guest's writes never reach the host. The alternative an author reaches for is a bind of the whole home directory, which is strictly worse. And `optional` is what lets a published sandbox naming `~/.gitconfig` run at all on a machine that has none.

The real defect was narrower than "these fields should not exist": **they were undeclared, and nothing decided how far a pulled document may reach.** This PR specifies both, then closes the gap that specifying them exposed.

### 1. The specification declares both fields

`§3.1.11` now names three fileset sources instead of two, with the `hostPath` rule table taken from what `validate_host_source` already enforces — anchored at `/` or `~/`, no `..`, no whitespace or quotes, never secret-shaped, and one file so `guestPath` may not end in `/`. `§3.1.10` gains `optional`, refused on a named volume. `§5`, `§6`, and `§1.3` are updated to match.

This is the specification changing deliberately, in its own commit, as CLAUDE.md's transitional mode allows when the code encodes a constraint the specification did not know about.

### 2. A pulled `hostPath` becomes this machine's decision

Specifying `hostPath` surfaces the question it always raised and nobody had answered: a `hostPath` makes what a document reads depend on the machine running it, so a published artifact was reading your files on the strength of its own declaration.

It asks now. The answer is recorded per machine in `~/.lns-host-path-decisions.json`, keyed by the artifact's **repository** and the host path:

```json
{
  "ghcr.io/acme/agent|~/.gitconfig": "allow",
  "ghcr.io/acme/agent|~/.npmrc":     "deny"
}
```

The key shape is the design decision. Stripping the tag and digest means a version bump keeps your answer, while a different sandbox never inherits it.

| Situation | Result |
|---|---|
| First run, you allow | File is read; answer recorded |
| First run, you decline, entry is `optional` | Fileset skipped, run continues |
| First run, you decline, entry is required | Run refused |
| Later run, same repository, new tag | Not asked |
| Different repository, same path | Asked |
| `--yes`, nothing recorded | Allowed, nothing recorded |
| `--yes`, an answer recorded | **The recorded answer wins** |
| Your own `lns.yaml` | Never asked |

The last two are worth calling out. `--yes` answers a question this machine has not been asked; it does not overrule one you already answered. And a document in your own directory is your own consent, so `DocumentOrigin` carries the local-vs-pulled distinction in the type rather than leaving it implicit at the call site.

### 3. `mountPath` becomes `guestPath`

Specifying the fileset block put a second naming problem in plain view. A fileset is **not a mount**: its entries become `RuntimeFileSpec` values that `inject_into_filesystem` writes as inodes into the composefs image tree before the guest boots (`runtime_layer/mod.rs:223`). Nothing is mounted at the path and nothing can be unmounted from it. Meanwhile the field that *is* a mount point — `volumes[].target` — was never called `mountPath`. The names were inverted.

```yaml
filesets:
  - hostPath: ~/.gitconfig
    guestPath: /home/agent/.gitconfig   # was: mountPath
```

`guestPath` says what the field decides: where the files land. `volumes[].target` keeps its name, because a volume really does mount. The rename covers the wire field, every Rust identifier, both IPC structs, the error strings, the guides, and the shipped example — no alias, no shim.

### 4. The refusal holds at the seam that opens the file

A prompt the service ignores is not a control. The denial travels on `RunImageArgs.denied_host_paths`, and `host_fileset_specs` skips a source on that list before it resolves or stats anything — so the file is never opened, not merely left out of the guest. `plan_local` passes `&[]`, which is the "your own directory is your own consent" rule expressed where it is enforced.

## Breaking changes

- `filesets[].mountPath` becomes `filesets[].guestPath`. A document using the old field no longer decodes — strict decoding reports it on the offending line rather than defaulting. `docs/examples/claude-code/lns.yaml` is updated.
- `lns_artifact::spec::validate_mount_path` becomes `validate_guest_path`, and `lns_ipc::SandboxFileset` / `PackedFilesetSource` rename their `mount_path` field, which changes the CLI↔service IPC. Both ship as one unit.
- `FilesetSummary.from_host: bool` becomes `host_path: Option<String>` plus `optional: bool`, with `from_host()` as a method. The disclosure and the decision now read one field instead of two derived from the same thing.
- `lns_service::artifact::fileset::host_fileset_specs` takes a `denied: &[String]` parameter.
- `RunImageArgs` gains `denied_host_paths` (`#[serde(default)]`, so old frames decode).

Pre-1.0, no shims, every caller and test updated in the same commit.

## Docs

- `docs/sandbox-spec.md` — declares `hostPath` and `optional`, decides the pulled-`hostPath` rule, and renames `mountPath` to `guestPath` with a paragraph stating why a fileset is not a mount.
- `docs/running-workloads.md` — describes what a pulled run now asks and where the answer is kept. Written after the behaviour landed, so the guide describes what ships.

## Test instructions

1. **Spec/code agreement.** A document with `hostPath` and `optional` validates: `lns sandbox validate` on an `lns.yaml` whose fileset is `{hostPath: ~/.gitconfig, guestPath: /home/agent/.gitconfig, optional: true}`. Then change `guestPath` back to `mountPath` and confirm it is refused by name, not silently defaulted. Confirm the fields in `§3.1.11`'s table match what the error messages enforce when you break each rule (`~alice/…`, a `..` segment, a `guestPath` ending in `/`, `optional` on an `inline` fileset).
2. **Local runs are never asked.** `lns run` on that local `lns.yaml`. Expect no host-file prompt, and the file present in the guest. Confirm `~/.lns-host-path-decisions.json` is not created.
3. **First pulled run asks.** `lns push` it, then `lns run <ref>` from another directory. Expect a prompt naming the repository and `~/.gitconfig`. Answer `y`; expect the file present in the guest and the answer in `~/.lns-host-path-decisions.json` under `<repo>|~/.gitconfig`.
4. **A version bump does not re-ask.** Push the same sandbox as a new tag and run it. Expect no prompt.
5. **A different sandbox does ask.** Push under a different repository and run it. Expect a prompt, and a second key in the decisions file.
6. **Declining an optional entry skips it.** Delete the decisions file, run the pulled sandbox, answer `n`. Expect the run to continue with the fileset absent, and the service log line `skipping hostPath … — this machine did not grant it to this sandbox`.
7. **Declining a required entry refuses the run.** Same with `optional` removed from the document. Expect a refusal naming the path, and — this is the part that regressed in review — expect the `deny` to be **recorded**, so the refusal message's advice to edit the file is true.
8. **`--yes` does not overrule you.** With a recorded `deny` for an optional entry, run with `--yes`. Expect no prompt and the file still not read.
9. **A corrupt decisions file does not break unrelated runs.** Write `{ not json` into `~/.lns-host-path-decisions.json` and run a pulled sandbox that declares no `hostPath`. Expect it to run normally.


crates/lns-service/src/artifact/real.rs
```diff
@@ -155,6 +157,7 @@ pub(crate) async fn plan_local(
     crate::artifact::fileset::host_fileset_specs(
         &RealSnapshotDir,
         &resolved.host_filesets,
+        &[],
```
chenhunghan: **A pulled mixin's `hostPath` is read with no decision.**

`plan_local` passes `&[]`, and consent runs only when `published.is_some()` (`crates/lns-cli/src/service/orchestrator.rs:400`). But a local `lns.yaml` merges its mixins *before* planning, and `def`/`json` are replaced by the merged document at `orchestrator.rs:296-305`. So `lns run` in your own directory, with `mixins: [ghcr.io/someone/toolkit@sha256:…]` declaring `hostPath: ~/.gitconfig`, reads the file with no prompt and nothing recorded — and under `-q` there is no disclosure line either, since `print_run_summary` is skipped.

Verified against the real merge path:

```
sources: ROOT {"image":"x:1"}
         ghcr.io/someone/toolkit {"filesets":[{"hostPath":"~/.gitconfig","mountPath":"/home/agent/.gitconfig"}]}

filesets=[FilesetEntry { path: None, inline: None, host_path: Some("~/.gitconfig"),
  mount_path: "/home/agent/.gitconfig", owner: Workload, optional: false }]
contribution.source="ghcr.io/someone/toolkit"
```

`fold_mounts` (`lns-artifact/src/merge.rs:348-403`) claims every fileset kind from every source with no filter, and `refuse_what_no_sandbox_could_be` (`mixin.rs:446`) only re-parses the result — nothing refuses or strips a mixin-contributed `hostPath`. `artifact/mod.rs:231-242` then turns it into a `HostFileset` whichever source contributed it.

**Why:** §3.1.11 as this PR writes it says an artifact from a registry MUST NOT read one on the strength of its own declaration. A mixin is an artifact from a registry. "You typed the mixin reference, so it's your consent" does not hold either — `flatten` walks the transitive graph, so a `hostPath` can arrive from a mixin the developer never named. The attribution needed to key the decision already exists and already crosses the wire: `Contribution { block: Mount, key: <mountPath>, source: <label> }`, reaching the CLI as `ResolvedDefinition.contributions`.

crates/lns-policy/src/host_path_decisions.rs
```diff
@@ -0,0 +1,239 @@
+//! Whether a pulled sandbox may read one of this machine's files lives in `~/.lns-host-path-decisions.json`, per machine and never committed — a `hostPath` makes what a document mounts depend on the machine running it, which is a risk one developer accepts on one computer, not a rule a directory keeps.
```
chenhunghan: **This is a verbatim clone of `host_bind_decisions.rs`.**

The store trait, the `HashMap<String, T>` alias, `default_*_path()` with its `LNS_*_PATH` / `HOME` / cwd triad, the `JsonFile…Store` impl, and roughly fifteen tests (kebab-case, round-trip, unknown variant, missing file, malformed JSON, non-`NotFound` I/O, the three env-guard path tests) are copied with only the enum renamed. Both files independently re-prove serde's behaviour.

**Why:** Two per-machine decision files of identical shape now drift independently, and the third will copy this again. One store parameterized by filename and env var reduces this to the `HostPathDecision` enum plus the tests that pin the genuinely new behaviour — the key shape.

crates/lns-cli/src/run/host_path_consent.rs
```diff
@@ -0,0 +1,122 @@
...
+    let declared: Vec<&FilesetSummary> = filesets.iter().filter(|f| f.from_host()).collect();
+    if declared.is_empty() {
+        return Ok(HostPathGrant::default());
+    }
+    let mut recorded = store
+        .load()
+        .context("reading this machine's host file decisions")?;
+    let mut settled = Vec::with_capacity(declared.len());
+    for fileset in declared {
+        let host_path = fileset.host_path.as_deref().unwrap_or_default();
```
chenhunghan: **The invariant lives in a fallback instead of the type.**

`fileset.host_path.as_deref().unwrap_or_default()` appears here and at lines 75 and 93. It is unreachable — `declared` filters on `from_host()` — but an empty string would silently become a `repo|` decision key and a deny entry matching nothing.

**Why:** CLAUDE.md's rule is to encode the invariant in the type. `filter_map` the declared set into `(&FilesetSummary, &str)` once and the fallback disappears from all three sites.

chenhunghan (commented):
**Problem:** `filesets[].hostPath` and `optional` shipped undeclared, and nothing decided how far a pulled document may reach into the files of the machine running it. This declares both in `docs/sandbox-spec.md` and makes a pulled `hostPath` a per-machine decision recorded in `~/.lns-host-path-decisions.json`.

**Approach:** `DocumentOrigin` carries local-vs-pulled in the type; the CLI prompts once per (repository, host path), records the answer immediately, and ships refusals to the service on `RunImageArgs.denied_host_paths`, where `host_fileset_specs` skips a denied source before it resolves or stats anything.

**Product alignment:** Squarely "policy you run into, not write" and "real secrets stay outside the workload". Stripping tag and digest from the decision key is the right call, and `--yes` answering only where nothing is recorded is the correct reading of what that flag means.

**Verdict:** Needs fixes. Three inline, plus one that lands on a file this diff does not touch:

### Should fix — `lns uninstall` leaves the decisions file behind

`crates/lns-cli/src/uninstall/real.rs:79-83`

The `secret_files` list names `credentials`, `registry_auth`, and `host_bind_decisions`; `host_path_decisions` was not added, and `purge_targets` (`uninstall/mod.rs:260-278`) forwards that list verbatim with no glob over `~/.lns-*`.

**Why:** A recorded `allow` is the only thing between `lns run <repo>` and reading `~/.gitconfig` unprompted. Surviving a wipe of everything else means uninstall → reinstall → run inherits an answer from an installation that no longer exists. If these answers should instead persist, then `host_bind_decisions` should not be purged either — but the two files are the same category of state and should agree, and today they do not.

---

Two observations, not filed as findings: the developer answers two prompts about the same file in one run (`confirm_pulled_effects` discloses `host file ~/.gitconfig` and takes a `Continue? [y/N]`, then `decide_host_paths` asks per path), and the only way to revise an answer already given is hand-editing the JSON — there is no `lns` command to list or reset these.

crates/lns-service/src/artifact/real.rs
jansav: Confirmed and fixed — thank you, this was a real hole and my reasoning behind it was wrong. "You typed the mixin reference, so it's your consent" doesn't survive `flatten` walking the transitive graph, as you say.

`FilesetSummary` now carries `declared_by`, filled from the merge attribution you pointed at (`ContributionBlock::Mount`), and the rule lives in one function:

```rust
fn decider<'a>(origin: &'a DocumentOrigin, fileset: &'a FilesetSummary) -> Option<&'a str> {
    if let Some(mixin) = fileset.declared_by.as_deref() {
        return Some(mixin);   // a registry artifact, whatever layered it in
    }
    match origin {
        DocumentOrigin::OwnDirectory => None,
        DocumentOrigin::Pulled { reference } => Some(reference),
    }
}
```

The orchestrator now runs consent for local runs too, and `plan_local` forwards `denied_host_paths` instead of `&[]`, so the refusal holds at the seam that opens the file.

One correction to my first attempt, which you'd have caught: filtering only `ROOT_LABEL` made `lns-local-mixin.yaml` and directory mixins into deciders, so a non-interactive local run whose own decisions file declared a `hostPath` would have bailed "no terminal to decide" — over the developer's own file. The decider filter is now `is_digest_pinned_image`, since the preflight pins every reference it fetches; that separates a registry source from the root document, this directory's decisions file, and a `--mixin ./dir` directory mixin, and it subsumes the `ROOT_LABEL` case.

Four Layer 2 scenarios plus six Layer 3 tests on the mapping. Red first: with the fix reverted, "the prompt names ~/.gitconfig" and "the run is refused naming ~/.gitconfig" both failed; the local-source tests failed with `Some("lns-local-mixin.yaml")` against `None`. (commit 9c7eb9de)

One asymmetry I left, deliberately: a mixin's `hostPath` inside a **pulled** sandbox still keys to the sandbox reference rather than the mixin, because `adopt_pinned_mixins` runs after `args.filesets` is built. It over-asks and never under-asks, so I'd rather align it in its own change than widen this one.

crates/lns-policy/src/host_path_decisions.rs
jansav: Fair, and extracted. New `crates/lns-policy/src/decision_store.rs` holds what both files were re-proving: `DecisionFile<T>`, a generic `DecisionStore<T>`, `JsonDecisionStore<T>`, and one `default_path(override_env, filename)` for the env / `HOME` / cwd triad. Its tests run over a synthetic `SomeAnswer` fixture, so serde's behaviour is proved once.

Each decision module now holds only what is genuinely its own:

- `host_bind_decisions.rs` — `SecretDisposition`, its aliases, its filename, and two tests: the kebab-case wire form and the unknown-variant refusal (`"mask"` must fail loudly rather than read as keep).
- `host_path_decisions.rs` — `HostPathDecision`, its aliases, its filename, and the key-shape tests, which are the genuinely new behaviour you identified.

Both keep a test pinning that they use *different* files, since sharing one would merge two unrelated risk decisions.

The ~15 duplicated tests are gone. `HostBindDecisionStore` and `HostPathDecisionStore` became `dyn` type aliases, so call sites dropped their `dyn` keyword and the fakes implement `DecisionStore<T>` — that reaches `host_bind.rs`, the orchestrator, `uninstall/real.rs`, and three step modules, all mechanical.

Coverage after the change: `decision_store.rs` 100% (57/57), `host_bind_decisions.rs` 100% (15/15), `host_path_decisions.rs` 100% (39/39). (commit 9c7eb9de)

crates/lns-cli/src/run/host_path_consent.rs
jansav: Fixed — the fallback is gone from all three sites, and via `filter_map` as you suggested. The declared set is now built once into a small private type, so the host path arrives as `&str` and there is nothing left to unwrap:

```rust
struct Declared<'a> {
    host_path: &'a str,
    decider: &'a str,
    optional: bool,
    guest_path: &'a str,
}
```

`ask` and `apply` take `&Declared` instead of re-deriving from the summary, so the `repo|` key you described is now unrepresentable rather than merely unreachable. The same restructure carried the mixin-decider fix from the other thread, since both wanted the deciding artifact resolved once per entry instead of per call site. (commit 9c7eb9de)

chenhunghan (approved):

PR #266, opened 2026-08-19
feat: let a document say how big a disk it needs

jansav:
The default disk sizes are not enough, and nothing could say otherwise. One
constant sized every block device at 10 GiB — both the disk a run writes to and
every named volume.

A document now declares both:

```yaml
spec:
  resources:
    disk: 40Gi          # the run's own disk, discarded when the run ends
  volumes:
    - name: build-cache
      target: /root/.cache
      size: 100Gi       # a named volume, which outlives the run
```

The run's disclosure names what it will boot with, before it boots:

```
Resources: 3 vCPU · 6144 MiB · 40 GiB disk
```

## `size` is a floor

A named volume outlives the run that made it, so `size` states the capacity the
volume must have, not the capacity it was created with. A volume that does not
exist is created at it; a smaller one grows to it in place, keeping everything
it holds; a larger one is left alone. Nothing ever shrinks a volume — a shrink
must choose which data to drop, and no document may make that choice on a
machine it cannot see.

The floor is what makes the field usable. `size` defaults to `10Gi`, so every
document declares one whether it says so or not; were a smaller declaration an
error, every default document would refuse every volume that had ever grown.

Raising the number is enough. The next run grows the volume before the workload
starts.

## Growing an ext4 image in place

The descriptor table sits immediately after the superblock, and every group's
bitmaps follow it, so extending that table on a full image would move live data.
The fix has to be made at creation: hold blocks back for it. `e2fsck` refuses
reserved blocks on their own — it reads them as the property of inode 7 — so the
reservation and the resize inode arrive together.

A grow never re-derives geometry from the new size. How many inodes a group
holds, how long the inode table is, where the journal sits, and where the resize
block lives all follow the size the image was *created* at. `Layout::for_existing`
says so at the call site, and the resize block is read back out of inode 7 rather
than calculated. Existing group descriptors are copied from the live table, never
rebuilt: they carry allocation counts nothing else records.

A live volume is far too large to copy, so there is no rename to hide a partial
write behind. The image is marked unclean, extended, written, and only then does
the superblock publish the new size. A grow that stops halfway leaves a valid
smaller volume inside a longer file.

It refuses an image it did not write, one that was not unmounted cleanly, one
whose counts contradict each other, and a grow that would spend the last of its
reserved run.

## Verification

`e2fsck` accepts every shape of grow — inside one group, filling a short tail,
appending groups, and 10 GiB → 100 GiB, which consumes reserved descriptor
blocks — through both the primary and a backup superblock. Our feature set now
matches what `mke2fs` writes for the same image.

The test that proves the most: a volume grown from 32 MiB to 400 MiB returns the
payload written before the grow byte for byte, then accepts 60 MiB more — twice
what it originally held. `e2fsck` alone cannot show that the new blocks are
allocatable; only writing to them can.

## Two corrections along the way

- The specification put the minimum disk at `16Mi`. Nothing can format an image
  that small: group 0 must hold its own metadata and four times the smallest
  journal JBD2 can recover. A document asking for `16Mi` passed validation and
  failed later, at provisioning. The minimum is now `20Mi`, and a test ties the
  validator's floor to what the writer can actually format.
- Neither field takes a share of the host. A share of total cores or RAM names a
  size the host can honour, because a run returns them when it ends. Disk bytes
  stay spent.

## Commits

Eight, each reviewed and landing one concern. The specification changes land
alone, ahead of the code that implements them, per the Transitional mode rules.


crates/lns-service/src/upperfs/writer.rs
```diff
@@ -34,6 +36,225 @@ pub fn write_ext4(plan: &Plan, path: &Path) -> Result<()> {
...
+pub fn grow_ext4(path: &Path, new_size_bytes: u64) -> Result<()> {
+    let mut f = OpenOptions::new()
+        .read(true)
+        .write(true)
+        .open(path)
+        .with_context(|| format!("opening {}", path.display()))?;
+
+    let mut sb_bytes = [0u8; 1024];
+    read_at(&mut f, 1024, &mut sb_bytes).with_context(|| format!("reading {}", path.display()))?;
+    let superblock = Superblock::from_bytes(&sb_bytes)
```
chenhunghan, line 49: **An unclean volume can never be acquired again, even when nothing needs to grow.**

`Superblock::from_bytes` runs its preconditions — including `state == EXT2_VALID_FS` — before `plan_grow` on line 52 gets to say "nothing to do". And `ensure_image` calls this on *every* acquire of an existing volume, not only when the declared size went up.

The guest mounts volumes `ext4` read-write (`lns-init/src/mount.rs:551`), so the kernel clears the valid bit for the duration of every run. Any hard-killed VM (`lns rm`, tray quit, host crash) leaves the volume with `state == 0`, and typically with `INCOMPAT_RECOVER` (0x4) set too — which trips the other guard with an even less actionable message ("image sets incompatible features 0x00000004 this writer cannot rewrite").

**Why:** after one killed sandbox, every subsequent `lns run` touching that volume fails outright, with unchanged sizes and no grow requested. There is no fsck in the product, so the only recovery is `lns volume rm` — data loss. The test comment at `lns-init/src/mount.rs:2646` ("a volume left inconsistent by a killed run") says this state is expected.

Minimum fix: plan the grow first, and only enforce the preconditions when there is something to write.

crates/lns-service/src/upperfs/writer.rs
```diff
@@ -34,6 +36,225 @@ pub fn write_ext4(plan: &Plan, path: &Path) -> Result<()> {
...
+        .with_context(|| format!("reading the superblock of {}", path.display()))?;
+
+    let Some(plan) = crate::upperfs::grow::plan_grow(&superblock, new_size_bytes)? else {
+        return Ok(());
+    };
+
+    mark_unclean(&mut f, &superblock)?;
```
chenhunghan, line 68: **An interrupted grow leaves the volume permanently ungrowable.**

`mark_unclean` is the crash-safety marker — but the writer then refuses any image carrying that exact marker. The grow is otherwise idempotent (`before` is re-derived from the untouched primary superblock, so a retry replans identically), yet the retry can never start.

**Why:** a service restart or SIGKILL mid-grow bricks the volume for `lns` in the one case the marker exists to survive. This is not fixed by the finding above: after the interrupt the size is still short, so a grow *is* needed.

The check itself is right for a genuinely crash-dirty ext4 — an unreplayed journal would overwrite the rewritten bitmaps and descriptors after the grow published the new size. That is precisely why the writer needs a marker of its own that it can tell apart from the guest's, rather than reusing ext4's clean bit for both meanings.

chenhunghan (commented):
**Problem:** One constant sized every block device at 10 GiB — both the run's own disk and every named volume. The document now declares both (`spec.resources.disk`, `volumes[].size`), and an existing volume grows in place to meet a raised floor.

**Approach:** `size` is a floor, never a shrink. `Layout::for_fresh_image` now holds back a reserved GDT run plus a resize inode so the descriptor table can extend without moving live data; `grow.rs` plans a grow from the geometry the image was *created* with, and `grow_ext4` writes unclean → extend → new groups → GDT → backups → superblock-as-commit. The ext4 work is careful and well-evidenced: `e2fsck` through primary and backup superblocks over every grow shape, plus a debugfs write/grow/read-back that proves the added blocks are allocatable.

**Product alignment:** Fits "ephemeral by default" (run disk discarded) and "one directory = one project" (the document owns the size, no flag). The specification changes land ahead of the code in their own commits, per Transitional mode.

**Verdict:** Needs fixes. Two findings, both on the interaction between the grow's preconditions and a volume that a killed run left dirty.

(Separately, and not commented inline: the `coverage` gate is red — `artifact/inspect.rs` 99.60%, `upperfs/format.rs` 99.36%, `upperfs/mod.rs` 98.04%, `upperfs/writer.rs` 99.34%.)

chenhunghan (approved):

crates/lns-service/src/upperfs/writer.rs
jansav: Fixed in 972cad41.

You were right about the consequence: the store calls this on every acquire, so one killed sandbox made the volume unusable with no resize requested, and `lns volume rm` was the only way out.

`Superblock::from_bytes` now only parses — magic, block size, blocks-per-group, inode size, and count consistency. Reading a volume says nothing about whether it may be written to. `grow_ext4` plans first and returns early when there is nothing to do, so the preconditions are enforced only when there is something to write.

The refusal is also actionable now, since there is no fsck to send anyone to: `the volume was not unmounted cleanly, so its journal still holds writes this resize would overwrite; run the sandbox once to let the guest replay it, then ask for the larger size again`. You noted the RECOVER path gave an even less actionable message — a killed run sets both bits, so that was the message it actually hit. `INCOMPAT_RECOVER` now gets its own refusal with the same wording, ahead of the generic feature one.

Pinned by `a_volume_a_killed_run_left_dirty_still_attaches_when_nothing_needs_to_grow`, which fails if the precondition runs before the plan.

crates/lns-service/src/upperfs/writer.rs
jansav: Fixed in 972cad41, and your framing is what the fix follows — the writer needed a marker it can tell apart from the guest's, rather than reusing ext4's clean bit for both meanings.

A `<image>.growing` sidecar is written and fsynced before `mark_unclean` and removed after the commit. When it is present the cleanliness check is skipped, because the cleared bit is this writer's own; the grow re-plans from the primary superblock, untouched until the commit, so the retry replans identically.

Two holes in that, found reviewing my own fix and worth recording here:

1. The marker must excuse the cleared valid bit and **nothing else**. A guest can mount and be killed while the marker sits there, and its unreplayed journal would replay over everything the resize wrote — exactly the corruption you said the check exists to prevent. The feature checks, `INCOMPAT_RECOVER` among them, now run on every resize whether the marker is there or not.
2. The tail retally was read-modify-write. A resize interrupted *after* its descriptor table landed would add the tail's freed blocks a second time on retry. It now counts the group's free blocks from the bitmap instead of adding to the descriptor, so it is idempotent — pinned by byte-comparing a resumed grow against a control that ran uninterrupted.

Also: `lns volume rm` now deletes the marker with the image, so a volume recreated under the same name cannot inherit an excuse it never earned.

Pinned by `a_grow_cut_short_can_be_run_again` and `a_stale_marker_does_not_excuse_a_guest_journal_that_needs_replay`, both of which fail if the respective fix is reverted.

jansav:
Both findings fixed in 972cad41, replies on each thread.

The coverage note from your review is also cleared — that was 3c4b6615. Nine lines, all in code I had added: two `panic!` arms in my own `let ... else` tests that cannot be taken, three `ensure!` message arguments that only evaluate on failure, one closure body that became its own line when rustfmt split a call, and the three early-return arms of the grow writer. CI is green across `coverage`, `test`, `lint`, and `audit`.

Worth recording why these got past me: every test I had written started from a cleanly-written image, so nothing exercised the state a killed run actually leaves behind. Thanks for reasoning about the lifecycle rather than the diff.

PR #272, opened 2026-08-20
feat(service): Reclaim run scratch dirs on exit, rm, prune, and at startup

### LINKED ISSUE #271: Run scratch dirs are never reclaimed after a sandbox run ends
## Problem

`~/Library/Caches/lns/runs/<run-id>/` survives every successful run forever. Each dir holds a sparse 10GiB `upper.img` plus `console.log`. On one dev machine: 30 dirs, 38G actual usage, largest single `upper.img` at 9.4G. This is a leak, not a cache — no command reclaims it.

## Why it happens

- `RunScratchGuard` (`crates/lns-service/src/run/scratch.rs:16`) removes the dir on drop, but only while the guard is armed. `crates/lns-service/src/run/orchestrator.rs:588` calls `keep()` as soon as the broker connects, so any run that boots keeps its dir.
- `lns sandbox rm` and `lns sandbox prune` only drop the in-memory registry entry (`run_registry::remove`/`prune_exited`, `crates/lns-service/src/run_registry.rs:382`, `crates/lns-service/src/ipc/mod.rs:434`). Neither touches the filesystem.
- Ephemeral-by-default says the scratch upper of a finished run has no reason to survive it.

## Expected behavior

1. When a run exits, the service reclaims that run's scratch dir. The audit chain lives in the data dir, so nothing user-visible is lost; `console.log` retention needs a decision (keep the log, drop `upper.img`, or keep the dir until `sandbox rm`).
2. `lns sandbox rm <run>` and `lns sandbox prune` reclaim the scratch dirs of the runs they remove and report the reclaimed bytes, like `volume prune` and `image prune` do.
3. Startup reclaims orphans: a scratch dir with no live run left over from a killed service or a crash is removed.

## Open decisions

- Does an exited run keep `console.log` for `lns sandbox logs`? If yes, delete only `upper.img` on exit and remove the dir on `rm`/`prune`.
- Does `--rm`-style behavior differ from the default? Today there is no distinction.

## Acceptance criteria (Gherkin sketch)

```gherkin
Scenario: A finished run gives back its scratch space
  Given a run has exited
  When the service reclaims the run
  Then the run's upper.img no longer exists

Scenario: Prune reports what it reclaimed
  Given two exited runs with scratch dirs
  When I run `lns sandbox prune`
  Then both scratch dirs are gone
  And the command reports the reclaimed bytes

Scenario: A scratch dir orphaned by a service crash is reclaimed
  Given a scratch dir whose run is not in the registry
  When the service starts
  Then that scratch dir is removed
```

chenhunghan:
## Summary

Closes #271.

`~/Library/Caches/lns/runs/<run-id>/` — a sparse 10GiB `upper.img` plus `console.log` per run — survived every booted run forever: the scratch guard is disarmed at broker connect, `rm`/`--rm` only dropped the in-memory registry entry, and `lns sandbox prune` never touched runs at all. One dev machine had 30 dirs holding 38G. This is a leak, not a cache; *ephemeral by default* says a finished run's scratch upper has no reason to survive it.

### 1. Exit reclaims the upper disk

When a run's session ends, the service deletes that run's `upper.img`. `console.log` stays in the dir for post-mortem diagnostics until the run is removed. Nothing user-visible is lost: `lns sandbox logs` streams from the in-memory `RunLogBuffer`, and the audit chain lives in the data dir (`data_root()/runs/<id>/audit.jsonl`), which this PR never touches.

### 2. rm, --rm, and prune reclaim the whole dir

- `lns sandbox rm <run>` (and `--rm` at exit) now removes the run's scratch dir along with its record.
- `lns sandbox prune` now prunes exited runs too: the CLI sends `PruneRuns` before `PruneImages`, lists `removed run <id>` lines, and reports one combined `reclaimed <bytes>` line. `RunsPruned` gains `reclaimed_bytes` on the wire (pre-1.0, no compatibility shim).
- Reclaimed bytes are actual usage (`st_blocks * 512`), so a sparse `upper.img` reports what it really held — same accounting as the volume store.

### 3. Startup sweeps orphans

The registry is in-memory, so after a service crash or restart every dir under `cache_root()/runs/` is an orphan (including tool-staging scratch a killed provisioning left behind). `lns-service` sweeps them once at startup, inside `run_server` immediately after the exclusive socket bind and before accepting connections — so a second instance that loses the bind (the socket answers Ping) bails without deleting anything, and the winner's sweep can never race a live run or a live tool-provisioning scratch dir.

### 4. Design

All decision logic lives in `run/scratch.rs` behind a new pub `ScratchFs` port (`remove_dir_all` / `remove_file` / `list_dirs` / `allocated_bytes`) with seam functions (`reclaim_upper_with`, `remove_dir_with`, `prune_with`, `sweep_orphans_with`) held at 100% coverage; the coverage-ignored composition points (`orchestrator.rs`, `ipc/adapter.rs`) get one-line calls. Reclaim is best-effort: a missing path is benign, any other failure is a `log::warn!` and never fails the request — the startup sweep is the backstop.

Behaviour is pinned at Layer 3: `scratch.rs` unit tests cover every seam and port error path, `ipc/mod.rs` pins the registry → scratch integration through `handle_request` (`remove_and_prune_reclaim_the_runs_scratch_dirs`, `test_env` HOME redirect), and `ipc/adapter.rs` pins the sweep ordering — an exclusive bind sweeps, a losing second instance leaves scratch dirs untouched.

### Known target-spec tension

`docs/cli-spec.md` words `rm`/`--rm` as removing "the writable layer", which pairs with the target `start` verb (not shipped; #263 closed unmerged). Since nothing shipping can reuse an exited run's `upper.img`, exit-reclaim removes no capability — but when `start` lands, exit-time reclaim of the upper disk must be revisited.

## Docs

`docs/cli-reference.md`: the `rm`, `prune`, and `--rm` rows now describe scratch reclaim and the reclaimed-bytes report.

## Test instructions

1. `lns run alpine -- true`, then look under `~/Library/Caches/lns/runs/<run-id>/`: `upper.img` is gone, `console.log` is present.
2. `lns sandbox rm <run-id>`: the whole scratch dir is gone; `lns sandbox ls -a` no longer lists the run.
3. Two exited runs, then `lns sandbox prune --force`: prints `removed run <id>` for both, plus a `reclaimed <N> MiB` line covering their scratch space; both dirs are gone.
4. `lns run --rm alpine -- true`: the scratch dir is gone the moment the run exits, with no `rm` needed.
5. Kill `lns-service` mid-run (leaving a populated scratch dir), restart it (`lns service start`): the orphaned dir is removed at startup.
6. While a run is live, `lns sandbox prune --force`: the running run's scratch dir survives.
7. With a service already running, launch a second `lns-service`: it exits "service already running" and no scratch dir is touched.


crates/lns-service/src/main.rs
```diff
@@ -45,6 +45,7 @@ fn run_ipc_runtime(
 
     rt.block_on(async {
         spawn_signal_listener(shutdown.clone());
+        lns_service::run::scratch::sweep_orphans();
```
jansav: **Critical — the startup sweep runs before the service knows it is the only instance.**

`sweep_orphans()` runs here, before `ipc::run_server`. The single-instance check (`bind_or_replace_stale` → `is_instance_alive`) lives *inside* `run_server`. So a second `lns-service` deletes every dir under `cache_root()/runs/` — including the running instance's live `upper.img` and any in-flight tool-staging dir — and only then discovers another instance owns the socket and bails.

The `live` filter cannot save this: the second process has its own empty in-memory registry, so every live run looks like an orphan. That filter only ever protects the one case that can't happen (a sweep in a process that already has runs).

Reachable via the tray app launched while a CLI-started service runs, two `lns service start` racing past the ping check, and a healthy-but-slow instance that misses the 2s liveness probe.

**Why:** a user's running sandbox silently loses its writable layer mid-run. The fix is an ordering move — sweep after `bind_or_replace_stale` succeeds (inside `run_server`, or by binding in `main` and passing the listener down), so exclusive ownership is proven before anything is deleted.

crates/lns-service/tests/behaviours/scratch_rig.rs
```diff
@@ -0,0 +1,87 @@
...
+    fn alloc_id(&mut self) -> String {
+        let id = format!("{:032x}", self.next_id);
+        self.next_id += 1;
+        id
+    }
+
+    fn make_scratch_dir(&self, run_id: &str) {
+        let dir = self.run_dir(run_id);
+        std::fs::create_dir_all(&dir).expect("scratch dir");
+        std::fs::write(dir.join("upper.img"), vec![7u8; 4096]).expect("upper.img");
```
jansav: **Should fix — this Layer 2 rig writes to the real filesystem and bypasses the crate's public surface.**

The rig `create_dir_all`s and `write`s real files, and calls the `*_with` seams with `RealScratchFs`. CLAUDE.md's Layer 2 rule is "**Forbidden.** No FS writes… Every cross-crate dep mocked through its `pub` trait surface."

It also drives the seams directly rather than `handle_request`, so the scenarios prune whatever list the rig hands them — `run_registry::prune_exited` and `remove_if_exited` are never exercised. "Prune keeps a running run's scratch dir" therefore proves only that `prune_with` removes the ids it was given, which the Layer 3 test already pins.

**Why:** these six scenarios restate `scratch.rs`'s unit tests at a higher cost and break the layer contract, while the integration they appear to cover (registry → scratch) is left to a single Layer 3 test. Drive them through `handle_request` with an injected `ScratchFs`, or drop them to Layer 3.

crates/lns-service/tests/behaviours/run_scratch_reclaim.feature
```diff
@@ -0,0 +1,39 @@
...
+    And an exited run with a scratch dir
+    When the exited runs are pruned
+    Then only the exited run's scratch dir is gone
+
+  Scenario: A scratch dir orphaned by a service crash is reclaimed
+    Given a scratch dir whose run is not in the registry
+    When the startup sweep runs
+    Then that scratch dir is removed
+
+  Scenario: The startup sweep never touches the audit chain
```
jansav: **Should fix — this scenario cannot fail.**

`sweep_orphans_with` only ever receives `cache_root`; the rig's `data_root` is a sibling tempdir the sweep has no path to. No implementation change to the sweep could turn this scenario red.

**Why:** a green-forever scenario advertises a safety property nobody is checking. Either give the sweep a case where the two roots could collide, or delete it.

jansav (commented):
**Problem:** Run scratch dirs (`cache_root()/runs/<id>/` — a sparse 10 GiB `upper.img` plus `console.log`) were never reclaimed by any path, so they accumulated forever. This PR reclaims them at exit, on `rm`/`--rm`, on `prune`, and sweeps orphans at startup.

**Approach:** All decisions live in `run/scratch.rs` behind a new `ScratchFs` port with `*_with` seams; production wrappers resolve `cache::root()` and delegate. `RunsPruned` gains `reclaimed_bytes`; the CLI sends `PruneRuns` before `PruneImages` and sums the two into one `reclaimed` line.

**Product alignment:** Directly serves *ephemeral by default*. The scope is right — the audit chain in `data_root()` is untouched, and `console.log` survives until removal.

**Verdict:** Needs fixes. One critical ordering bug that can destroy a live run's writable layer, plus two test-layer issues.

crates/lns-service/src/main.rs
chenhunghan: Fixed — the sweep moved out of `main.rs` into `run_server`, immediately after `bind_or_replace_stale` returns, so exclusive socket ownership is proven before anything is deleted. Added two Layer 3 tests in `ipc/adapter.rs`: a red-first one pinning that an exclusive bind does sweep a staged orphan, and one pinning that a second instance losing the bind (against a Pong-answering listener) bails with the orphan intact. (commit c72cff51)

crates/lns-service/tests/behaviours/scratch_rig.rs
chenhunghan: Fixed — took your second option and dropped the scenarios to Layer 3: the feature file, rig, and steps are deleted. The seam behaviours were already pinned by `scratch.rs`'s unit tests, and the registry → scratch integration through `handle_request` stays pinned by `remove_and_prune_reclaim_the_runs_scratch_dirs` in `ipc/mod.rs`. (commit bf8b3402)

crates/lns-service/tests/behaviours/run_scratch_reclaim.feature
chenhunghan: Fixed — deleted along with the feature file (commit bf8b3402). You're right that it could never go red: `sweep_orphans_with` only ever receives the cache root, so the audit-chain property is structural rather than something a scenario can falsify.

chenhunghan:
@copilot+claude-fable-5 resolve the merge conflicts in this pull request

PR #274, opened 2026-08-21
feat: run a document's pre-start scripts before the workload

jansav:
## What

A new `spec.scripts` block: an ordered list of shell scripts the guest runs before the workload entrypoint, allowed in both `kind: sandbox` and `kind: mixin`. Each entry names the slot it runs in (`when: pre-start`, the only slot defined), a `run` body, an optional `user`, and an optional `description`.

```yaml
spec:
  egress:
    http:
      - match: deb.debian.org
        verdict: allow
        description: the mirror the pre-start script installs from
  scripts:
    - when: pre-start
      user: root
      description: the psql this sandbox's prompts assume
      run: |
        apt-get update
        apt-get install -y --no-install-recommends postgresql-client
```

Six commits, each reviewed and approved on its own:

| | |
|---|---|
| `576e5a7d` | the normative decision in `docs/sandbox-spec.md` §3.1.13 (plus §3.1.3, §3.1.8, §3.3, §3.3.2, §5) and the matching `docs/cli-spec.md` changes |
| `f3818d23` | a §5 wording fix, split out because a normative change may not ride an implementation commit |
| `ae9c3e78` | the document format, the merge rule, and staging into the guest |
| `e633b531` | execution in `lns-supervisor` |
| `f76670a1` | the pre-boot disclosure |
| `a09973ea` | the user guides |

## Why

A mixin can declare what a run reaches, mounts and installs, but it cannot *run* anything — so a mixin that needs `apt-get install`, a generated config file, or a seeded cache has nowhere to put that work. `tools` deliberately refuses arbitrary install code (a tool name "is the tool itself, never a provisioner-specific coordinate"), and `command` is a launch block the sandbox owns and a mixin may not carry. The same gap exists in a sandbox.

Five decisions worth surfacing, because each one had an alternative:

- **Scripts run under the run's own egress policy**, after the network cage and the first policy frame. So a document shipping `apt-get install` ships its mirror's `egress` in the same document, and a script is not a way around the policy the consumer approved. Rejected: running pre-lockdown, which would give a pulled mixin an open-network window with the project's binds mounted.
- **Mixin scripts append rather than override.** This is the first block that is neither keyed nor last-wins. Two scripts answer no shared question, so dropping one would not be an override — it would be a missing dependency, discovered as a command not found inside the workload.
- **The 32-entry and 512 KiB ceilings bound one *document*, not a resolved one.** A sandbox with 20 scripts layering a mixin with 20 must not refuse a run neither author could fix. This is why `sandbox::parse_resolved` exists and is used at all five post-resolution re-parse sites.
- **What keeps a per-script `user` from being a way around the forbidden `user` block is disclosure, not narrowness of reach.** A bounded root is still root. Every script is approved before boot, attributed, showing the user it asks for.
- **A failing script exits `125`** — which `docs/cli-spec.md` §5 already defines as a failure before the workload started — rather than propagating the script's own status, which would collide with the reserved `1`/`2`/`125`/`126`/`127` meanings.

## Testing

- **Layer 3** — `lns-artifact` (format, validation, the two-tier ceiling split, append-order merge and per-script attribution), `lns-ipc` (wire round-trip and defaults), `lns-supervisor` (`scripts/`, `scripts/ids`, the config loader, and the dispatcher's ordering), `lns-cli` (`run/summary`, `run/pull_confirm`, `sandbox` inspect rendering, and `read_resolved_definition`).
- **Layer 2** — new `crates/lns-service/tests/behaviours/guest_scripts.feature`, 8 scenarios over staging: declaration order, mixin scripts appending after the sandbox's own, the user travelling beside the body, root-owned and unwritable staged files, and a run declaring none staging nothing. Service behaviours 254/254, CLI behaviours 448/448.
- **Gates** — `make lint` and `make complexity` clean. The per-file coverage gate reports **0 failures** for `lns-artifact`, `lns-ipc`, `lns-session`, `lns-supervisor` and `lns-service`.

## Security impact

This is the largest single increase in what a mixin can do: **a pulled mixin can now run arbitrary code as root inside the guest.** What bounds it:

- **The cage holds.** A root script keeps only `CAP_CHOWN`, `CAP_DAC_OVERRIDE`, `CAP_FOWNER`, `CAP_FSETID`, `CAP_KILL`, `CAP_SETGID`, `CAP_SETUID`. Without `CAP_NET_ADMIN` it cannot take down the nftables cage enforcing `egress`. A root run-as reaches the cap-drop path rather than `setuid(0)`, which would leave those capabilities in place.
- **Egress applies.** Scripts run after `install_network_lockdown()` and the first policy frame, through the same proxy, with the same ask-on-unknown-destination flow.
- **Consent is required.** A pulled sandbox whose only effect is a script still prompts before boot, and `--yes` names scripts among what it accepts.
- **The bytes are pinned.** Scripts are staged into the run's composefs runtime layer, so a changed script changes the descriptor digest. Staged files are `0o555` and root-owned, so a workload cannot rewrite the script a later run of the same sandbox will execute.
- **Identity cannot fall back.** A `user` the guest cannot resolve refuses the run before any script spawns, rather than falling through to root. Group resolution is numeric-first, matching `lns-init`, so an image naming a group for a numeral cannot hand a script a gid the workload would never get.

Residual risk: the pre-boot consent prompt is the only thing standing in front of a stranger's root script. I reused the existing single prompt and `--yes` for consistency with tool installers rather than adding a dedicated flag — worth a second opinion.


chenhunghan (approved):
**Problem:** A mixin (or sandbox) can declare what a run reaches and mounts but cannot run anything, so `apt-get install`, generated config, or cache seeding had nowhere to live. `spec.scripts` adds ordered pre-start shell scripts, allowed in both kinds.

**Approach:** Scripts run inside the guest after the network cage and first policy frame, under the run's own egress; mixin scripts append rather than override; authoring ceilings (32 entries / 512 KiB) bound one document while `parse_resolved` lets a legal merge exceed them; bytes are staged read-only into the run's runtime layer with a manifest the supervisor reads; a failure exits 125 before the workload. Disclosure travels the whole way — merge attribution, launch summary with per-script digest, full bodies in `inspect`, and the pull-confirm prompt.

Verified directly rather than from the description: every author-controlled input path still runs the ceiling-enforcing `parse` (`resolve()`'s root parse, image ingest, push, local target) and only post-resolution re-parse sites were relaxed, so a hand-crafted over-ceiling registry artifact is still refused at preflight; the shared `agent_started` latch flips before the script task spawns, so a mid-script reconnect cannot double-run the scripts; filesets cannot claim `/.lens`, so nothing a document declares can shadow a staged script; an unresolvable `user` refuses the run before the first script spawns.

On the residual risk flagged in the description (single prompt + `--yes` covering a stranger's root script): reusing the existing consent surface is right — a dedicated flag would fragment consent, and the disclosure line already names the body, the user, and the digest. If this ever needs tightening, the lever is making the prompt distinguish "runs code as root" from "mounts a directory", not another flag.

**Verdict:** Ship it. ✅

PR #280, opened 2026-08-24
feat(lns-cli): close the artifact-verb gaps §3.1 names

chenhunghan:
Stage 3 of 9. This PR closes the five artifact-verb gaps that §3.1 names.

> **Stacked on #279.** The base of this PR is `cli-spec-s2-run-flags`. The diff shows stage 3 only. Merge #279 first.

The verbs stay under `lns sandbox` here. Stage 5 (#283) moves them to `lns artifact`.

## What ships

| Gap | What it does now |
|---|---|
| `init` made one kind, with one file name | `lns sandbox init [--kind sandbox\|mixin] [-f FILE]`. You write a mixin in an editor too, so it gets a scaffold. `init` still refuses to overwrite. |
| `validate` answered only "is this valid" | `--kind <KIND>` also holds the document to the kind that you name. The report shows this together with everything else that validation found. |
| `inspect` of a local file supposed a sandbox | It reads the kind. `parse_document` replaces `sandbox::parse`, so a mixin shows as a mixin. Before, a mixin failed as a broken sandbox. |
| `ls` showed a reference and the word `cached` | The table shows `ARTIFACT  KIND  DIGEST  SIZE  HOLDER`. `--kind` filters it. |
| `prune` refused without `--force` | It asks you, as `lns volume prune` already does. `-f` skips the question. |

## What you see

Make a mixin:

```console
$ lns sandbox init --kind mixin
✓ created lns.yaml — your mixin, every field ready to edit

  1. name the capability it layers on — tools, filesets, egress, credentials
  2. try it with `lns run --mixin .`
  3. share it with `lns push`, e.g. `lns push ghcr.io/acme/my-mixin:1.0.0`

$ cat lns.yaml
apiVersion: lns.run/v1
kind: mixin
name: mixin
spec:
  egress:
    http: []
    tcp: []
  credentials: []
  filesets: []
  tools: []
```

`init` protects a file that is already there:

```console
$ lns sandbox init
Error: lns.yaml already exists in this directory; not overwriting it
```

`validate --kind` holds the document to the kind that you name:

```console
$ lns sandbox validate --kind sandbox
lns.yaml is not valid:
  - kind: expected a sandbox, but this is a mixin

$ echo $?
1

$ lns sandbox validate
lns.yaml is valid.
```

`inspect` reads the kind, so a mixin shows as a mixin:

```console
$ lns sandbox inspect
Mixin: mixin
  egress:       0 route(s)
```

`ls` shows what each cached artifact is:

```console
$ lns sandbox ls
ARTIFACT  KIND  DIGEST  SIZE  HOLDER

$ lns sandbox ls --kind sorcery
error: invalid value 'sorcery' for '--kind <KIND>'
  [possible values: image, sandbox]

```

`prune` asks first. With no terminal to ask at, it refuses:

```console
$ lns sandbox prune
Error: this removes every cached sandbox not held by a running one and, when none is
live, the provisioned tool cache; there is no terminal to ask at, so pass --force to confirm
```

## Where the kind comes from

`ImageInfo` gets a `kind` field: `image` or `sandbox`. The store reads the kind from the dependency edge of the record. An artifact record names the base image that it needs, and holds no layers of its own. A plain image record is the opposite. The stored format does not change, so no cache is lost, and no cache becomes unreadable.

The table does not list mixins. A pulled mixin takes no index entry, because it is configuration only, and nothing must reclaim it. When that changes, `CachedKind` gets one more value.

## `--kind connector` is not here

`lns_artifact::Kind` has only `Sandbox` and `Mixin`, so `kind: connector` is not a document that the code can read. `init --kind connector` would write a file that `validate`, `inspect`, `push` and `pull` all refuse:

```console
$ lns sandbox init --kind connector
error: invalid value 'connector' for '--kind <KIND>'
  [possible values: sandbox, mixin]
```

§3.3 waits for the same gap. The scaffold lands with that document format.

## Tests first

Twelve scenarios in `sandbox_author.feature` and `sandbox_manage.feature`. The code landed before the re-run, so each scenario got a mutation proof. Four mutations — a revert of `parse_document`, a drop of the `--kind` filter, a return to the old headers, and a short-circuit of the prompt — each failed the one scenario that claims the behaviour, and no other scenario.

## Docs

`cli-reference.md` — the `init`, `validate`, `ls`, `prune` and `inspect` rows and their synopses. It also says that the JSON now holds more than the table, and not different data. `running-workloads.md` — the three `init` forms. `policy.md` — the project's own mixin is a document that `validate` and `inspect` answer for, which is now true.

### DIFF
--- a/crates/lns-cli/src/sandbox/mod.rs
+++ b/crates/lns-cli/src/sandbox/mod.rs
@@ -532,7 +532,7 @@ where
         SandboxCommand::Logs(args) => logs(svc, args, stdout, stderr).await,
         SandboxCommand::Attach(args) => attach(svc, args, term, stdout, stderr).await,
         SandboxCommand::Rm(args) => rm(svc, args, out).await,
-        SandboxCommand::Prune(args) => prune(svc, args, input, out).await,
+        SandboxCommand::Prune(args) => prune(svc, args, term, input, out).await,
     }
 }
 
@@ -954,11 +954,19 @@ async fn remove_cached<W: std::io::Write>(
 async fn prune<I: std::io::BufRead, W: std::io::Write>(
     svc: &impl SandboxService,
     args: &SandboxPruneArgs,
+    term: TermInfo,
     input: &mut I,
     out: &mut W,
 ) -> Result<i32> {
-    if !args.force && !confirm_prune(input, out)? {
-        return Ok(0);
+    if !args.force {
+        if !term.stdin_is_tty {
+            bail!(
+                "this removes every cached sandbox not held by a running one and, when none is live, the provisioned tool cache; there is no terminal to ask at, so pass --force to confirm"
+            );
+        }
+        if !confirm_prune(input, out)? {
+            return Ok(0);
+        }
     }
     match svc.one_shot(Request::PruneImages).await? {
         Response::ImagesPruned {
@@ -2601,6 +2609,10 @@ mod tests {
         let code = prune(
             &svc,
             &SandboxPruneArgs { force: false },
+            TermInfo {
+                stdin_is_tty: true,
+                stdout_is_terminal: false,
+            },
             &mut std::io::Cursor::new(b"n\n".to_vec()),
             &mut out,
         )
@@ -2621,6 +2633,7 @@ mod tests {
         let code = prune(
             &svc,
             &SandboxPruneArgs { force: true },
+            TermInfo::default(),
             &mut std::io::empty(),
             &mut out,
         )
@@ -2642,6 +2655,7 @@ mod tests {
                 message: "registry poisoned".into(),
             }),
             &SandboxPruneArgs { force: true },
+            TermInfo::default(),
             &mut std::io::empty(),
             &mut Vec::new(),
         )
@@ -2652,6 +2666,7 @@ mod tests {
         let err = prune(
             &CannedService::new(Response::Pong),
             &SandboxPruneArgs { force: true },
+            TermInfo::default(),
             &mut std::io::empty(),
             &mut Vec::new(),
         )
--- a/crates/lns-cli/src/sandbox/mod.rs
+++ b/crates/lns-cli/src/sandbox/mod.rs
@@ -27,8 +27,8 @@ pub struct SandboxArgs {
 
 #[derive(clap::Subcommand)]
 pub enum SandboxCommand {
-    #[command(about = "Scaffold a default ./lns.yaml (kind: sandbox) in this directory.")]
-    Init,
+    #[command(about = "Scaffold a document in this directory; `--kind` chooses which.")]
+    Init(SandboxInitArgs),
     #[command(about = "Validate ./lns.yaml — schema, cross-field, and secret checks, offline.")]
     Validate(SandboxValidateArgs),
     #[command(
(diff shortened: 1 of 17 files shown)


crates/lns-cli/src/sandbox/mod.rs
```diff
@@ -80,12 +80,81 @@ pub struct PsArgs {
...
+    )]
+    pub kind: Option<CachedKindFilter>,
+
     #[command(flatten)]
     pub output: crate::output::OutputArgs,
 }
 
+#[derive(Copy, Clone, Debug, PartialEq, Eq, clap::ValueEnum)]
+#[value(rename_all = "lower")]
+pub enum CachedKindFilter {
```
jakolehm, line 97: `CachedKindFilter` accepts `image|sandbox`, while §3.1 defines artifact kinds as `sandbox|mixin|connector`. This introduces `image` as a CLI value absent from the target and leaves actual artifact kinds unfilterable. Stage 5 would otherwise move an incompatible surface to `lns artifact ls`, deepening the documented divergence.

crates/lns-cli/src/sandbox/mod.rs
```diff
@@ -858,15 +951,22 @@ async fn remove_cached<W: std::io::Write>(
...
     if !args.force {
-        bail!(
-            "this removes every cached sandbox not held by a running one and, when none is live, the provisioned tool cache; pass --force to confirm"
-        );
+        if !term.stdin_is_tty {
+            bail!(
+                "this removes every cached sandbox not held by a running one and, when none is live, the provisioned tool cache; there is no terminal to ask at, so pass --force to confirm"
+            );
+        }
+        if !confirm_prune(input, out)? {
```
jakolehm, line 967: The confirmation occurs before any service request, so candidates are only reported after deletion. §3.1 explicitly says prune ‘Lists them and asks.’ Users should be able to review the artifacts before consenting to their removal.

crates/lns-cli/src/sandbox/mod.rs
```diff
@@ -885,6 +985,25 @@ async fn prune<W: std::io::Write>(
     }
 }
 
+fn confirm_prune<I: std::io::BufRead, W: std::io::Write>(
+    input: &mut I,
+    out: &mut W,
+) -> Result<bool> {
+    write!(
+        out,
+        "This removes every cached sandbox not held by a running one and, when none is live, the provisioned tool cache. Continue? [y/N] "
```
jakolehm, line 994: This writer is stdout (`real.rs:436`), while the answer comes from the process stdin lock (`lib.rs:48`). §§4.1 and 7.2 require prompts on stderr and answers read from the terminal. Redirecting stdout currently hides the prompt while the command waits and contaminates redirected command output.

crates/lns-cli/src/sandbox/mod.rs
chenhunghan: Agreed — and it turned out to run deeper than the clap enum: the service can't classify a mixin (`CachedKind` is `{image, sandbox}` and `image_store::kind_of` infers from the dependency edge), and `pull_mixin_graph` deliberately records no index entry for mixins at all, so a `--kind mixin` value today could never match anything. Aligning the filter with §3.1 needs the store to record the document kind at ingest plus a decision on whether mixins appear in `ls` — tracked in #294 rather than papered over here.

crates/lns-cli/src/sandbox/mod.rs
chenhunghan: Right — this is the "Lists them and asks" gap, and it holds in all three prune homes (artifact, sandbox, volume), not just here. Tracked as #291: fetch the candidates, print them with the namespace's `ls` columns, then ask; empty set short-circuits without a prompt.

crates/lns-cli/src/sandbox/mod.rs
chenhunghan: Fixed — all three prune confirms (artifact, sandbox, volume) now write the prompt and the declining `Aborted.` to stderr, where §4.1 puts them; a new `prompt_streams.feature` pins per namespace that the prompt lands on stderr and never stdout. Shipped as dfad54ab on #293 (stacked, so it lands without rebasing this PR). The other half of §7.2 — answers read from the terminal instead of process stdin — needs a /dev/tty port and a sweep of the remaining prompt sites; tracked in #295.

crates/lns-cli/src/sandbox/mod.rs
chenhunghan: Fixed in #296 (commit 5d7808c9) — all three prunes now list what they would remove before asking, on stderr with the prompt. Artifact candidates come from the service via a new `ListPrunableImages` request (the CLI can't compute the dependency closure from `ls` data); sandbox/volume reuse `ListRuns`/`ListVolumes`. Empty sets short-circuit without prompting (artifact still asks, since it also reclaims the tool cache). Closes #291.

jakolehm (approved):

crates/lns-cli/src/sandbox/mod.rs
chenhunghan: Progress, tracked in #294: #297 (stacked on the series tip) makes mixins index-visible — pulling a mixin now records its pinned graph in the store, so `ls` lists it (`--kind mixin` filters), `rm` refuses one a cached sandbox layers on, and prune's closure protects what a live sandbox reaches. The follow-up PR removes `image` as a `--kind` value (base-image rows leave the user surface; their records stay internal for layer reclamation). `connector` joins the grammar when `kind: connector` documents land. Leaving this thread open until the `image` value is gone.

chenhunghan: Done in #299 — `image` is no longer a kind the cache admits to: `CachedKind` is `{Sandbox, Mixin}`, `--kind image` is refused (exit 2), base images never list, and `rm`/`tag` answer them as unknown while prune reclaims them silently. (commit 9a3fd9a7)

PR #281, opened 2026-08-24
feat(lns-cli)!: give the inspect verbs a --format, and run/exec the exit codes §5 names

chenhunghan:
Stage 4 of 9, first half. This PR gives the inspect verbs a `--format`, and gives `run` and `exec` the exit codes that §5 names.

> **Stacked on #280.** The base of this PR is `cli-spec-s3-artifact-verbs`. The diff shows the first half of stage 4 only. Merge #280 first.

## Output (§4)

`lns sandbox inspect <RUN>` and `lns volume inspect <NAME>` take `--format table|json`.

`table` is the default (§4.3). Both verbs printed JSON before, and they now print a `FIELD  VALUE` summary. This is the visible break: `lns inspect 3 | jq` becomes `lns inspect 3 --format json | jq`.

The JSON does not change. It stays one object, never an array (§4.4). It holds what the table has no room for: the full launch configuration and the resolved policy.

### What you see

```console
$ lns volume inspect some-cache
FIELD     VALUE
NAME      some-cache
CAPACITY  10 GiB
ON DISK   21.9 MiB
CREATED   2026-08-24 20:08:04
IN USE    -

$ lns volume inspect some-cache --format json
{
  "name": "some-cache",
  "sizeBytes": 10737418240,
  "diskBytes": 22970368,
  "created": "2026-08-24T20:08:04Z",
  "inUseBy": null
}
```

The table rounds the sizes for a reader. The JSON keeps the raw byte counts for a script.

```console
$ lns sandbox inspect --help
Print one sandbox's live state and launch configuration.

Usage: lns sandbox inspect [OPTIONS] <RUN>

Options:
      --format <FORMAT>  Output format. json is experimental: its shape may change
                         before v1.0. [default: table] [possible values: table, json]
```

### Why a cached artifact refuses the flag

§3.2 gives `--format` to `lns sandbox inspect`, and that verb answers for a **running** sandbox. Today the same command also shows a cached artifact and a local document. A cached artifact shows itself, and the flag would do nothing there, so that path refuses the flag and says why. A local document routes offline before the flag is read, so it is not refused here; stage 5 (#283) closes that, moving both targets to `lns artifact inspect`, which has no `--format` at all.

## Exit codes (§5)

Before this PR, `lns` used only 0, 1 and 2.

### 125 — every failure of `run` or `exec`

A workload's own status comes back as a code, and never as an error. So an error on the `run` or `exec` path is always a failure of `lns`, and never the workload. §5 reserves 125 for this, "so a failure of `lns` is never mistaken for the workload exiting with the same number".

A service that is not running is one of these failures:

```console
$ lns run alpine
error: Lens Sandbox is not running. Run `lns service start` to start it.

$ echo $?
125
```

Other commands keep the ordinary 1 that §5 gives them:

```console
$ lns sandbox ls
Error: Lens Sandbox is not running. Run `lns service start` to start it.

$ echo $?
1
```

To make this possible, `require_running` no longer calls `std::process::exit(1)`, which no caller could overrule. It returns a `Result`. Two `process::exit` calls leave the codebase with it.

### 126 and 127 — the guest already knew the difference

`exec_child` exited 127 for every failure of `execvp`. It now reads the errno:

| errno | Code | Meaning |
|---|---|---|
| `NotFound` | 127 | The command is not in the sandbox. |
| anything else | 126 | The command is there, and it cannot run. |

The decision is a pure function in `session.rs`, which holds the host-testable half of the session driver. A unit test covers the errno kinds.

## Tests first

- `sandbox_cli.feature` — a run inspect gives a summary to a reader; `--format json` gives the record to a script; a cached artifact refuses the flag.
- `volume_cli.feature` — the table names `CAPACITY` and `ON DISK`; `--format json` holds the raw byte counts.
- `volume/mod.rs` — JSON for one thing is an object, and not a list.
- `session.rs` — a missing command gives 127; `PermissionDenied`, `InvalidData` and `IsADirectory` give 126.

## Docs

`cli-reference.md` gets an **Exit codes** table. It lists the two `inspect` verbs with the other `--format` takers, and it states the default and the refusal. The `lns volume` synopsis also changes.


crates/lns-cli/src/service/orchestrator.rs
```diff
@@ -30,10 +30,26 @@ fn primary_target(run_id: impl Into<String>) -> lns_ipc::SessionTarget {
...
+
 pub fn run_command<'a>(matches: &'a clap::ArgMatches, ctx: RunCtx<'a>) -> RunFuture<'a> {
     Box::pin(async move {
-        let args = RunArgs::from_arg_matches(matches)?;
-        launch_run(args, ctx.debug).await
+        let started = async {
+            let args = RunArgs::from_arg_matches(matches)?;
+            launch_run(args, ctx.debug).await
+        };
+        Ok(as_pre_start_failure(started.await))
```
jakolehm, line 52: Only the top-level `lns run` and `lns exec` wrappers call `as_pre_start_failure`. `lns sandbox run` and `lns sandbox exec` invoke `launch_run`/`exec_image` directly, so identical failures still exit 1 instead of 125. Exact aliases should not produce different exit codes.

crates/lns-cli/src/service/orchestrator.rs
```diff
@@ -30,10 +30,26 @@ fn primary_target(run_id: impl Into<String>) -> lns_ipc::SessionTarget {
     }
 }
 
+/// §5: a `run` or `exec` that fails is `lns` failing, never the workload — the workload's own status comes back as a code, so an error here is always the 125 case and is never mistaken for a workload that exited 1.
+pub const PRE_START_FAILURE: i32 = 125;
+
+fn as_pre_start_failure(result: Result<i32>) -> i32 {
+    match result {
+        Ok(code) => code,
```
jakolehm, line 38: This preserves `RunExit` received before `SessionReady` as though it came from the workload. Service boot and session-setup failures are currently encoded as exit codes, so they remain 1, 126, or 127 instead of becoming 125. Infrastructure failures can therefore still be mistaken for workload statuses, which §5 explicitly prevents.

crates/lns-session-broker/src/session.rs
```diff
@@ -241,6 +241,14 @@ pub(crate) fn validate_argv(argv: &[String]) -> Option<Vec<CString>> {
     (cargs.len() == argv.len()).then_some(cargs)
 }
 
+/// A command the sandbox does not have exits 127 and one it has but cannot execute exits 126, so a caller can tell a typo from a file that is there but unusable.
+pub(crate) fn exec_failure_code(kind: io::ErrorKind) -> i32 {
+    match kind {
```
jakolehm, line 246: `execvp` also returns `ENOENT` when the command exists but its shebang interpreter or ELF loader is missing. Those are ‘found but cannot execute’ cases and should be 126, so the errno alone cannot implement the §5 distinction.

crates/lns-cli/src/sandbox/mod.rs
```diff
@@ -290,6 +290,9 @@ pub struct SandboxInspectArgs {
         help = "Resolve this mixin into the sandbox before rendering it, as `lns run --mixin` would. Repeatable."
     )]
     pub mixins: Vec<String>,
+
+    #[command(flatten)]
```
jakolehm, line 294: This makes `--format` parse for every inspect target, but local documents route through `inspect_local`, which never reads it. `lns inspect . --format json` therefore exits successfully with prose instead of JSON or a refusal, which will silently break scripts expecting the requested format.

crates/lns-cli/src/service/orchestrator.rs
chenhunghan: Fixed upstack in #288 (8e07af5e) — `lns sandbox run` and `lns sandbox exec` now wrap their launch through `as_pre_start_failure` exactly like the top-level spellings, with behaviour scenarios pinning the 125 parity.

crates/lns-cli/src/service/orchestrator.rs
chenhunghan: Fixed in #298 (a4baf95d) — `pre_phase_step` now maps every `RunExit` received before `SessionReady` to 125. `SessionReady` is emitted before the broker session that carries the workload's status, so an earlier exit is by construction infrastructure, never the workload. Test: `a_run_exit_before_session_ready_is_lns_failing_never_the_workload` (red with 42 leaking through, green at 125).

crates/lns-session-broker/src/session.rs
chenhunghan: Fixed in #298 (b5a67697) — after an exec failure the broker child probes whether the command itself resolves (directly for a slash path, through each PATH entry otherwise, empty entry = cwd as execvp treats it); ENOENT only answers 127 when the command is truly absent, so a missing shebang interpreter or ELF loader is 126. The probe is a pure `command_is_present` host-tested with an injected existence check.

crates/lns-cli/src/sandbox/mod.rs
chenhunghan: Fixed upstack in #288 (3f5de87e) — the inspect shortcut now carries both namespaces' flags and refuses `--format` for a document or cached-artifact target with exit 2 (`lns inspect . --format json` no longer exits 0 with prose), pointing at `lns sandbox inspect <RUN> --format json` instead.

jakolehm (approved):

PR #282, opened 2026-08-24
feat!: one directory, ~/.lns, and LNS_HOME to move it

chenhunghan:
Stage 4 of 9, second half. This PR puts everything that `lns` keeps for you in one directory, `~/.lns`, and lets `LNS_HOME` move it.

> **Stacked on #281.** The base of this PR is `cli-spec-s4-output-exits`. The diff shows the second half of stage 4 only. Merge #281 first.

## What changes

Before this PR, `lns` used four places:

```text
<cache_dir>/lns       artifacts, layers, builds, volumes, the kernel
<data_dir>/lns        the audit chains and the connection ledger
<config_dir>/lns      config.yaml
~/.lns-*.yaml|json    connectors, credentials, registry logins,
                      workload grants, host-path and host-bind answers
```

All of it is now `~/.lns/`. `LNS_HOME` moves the root itself. It is not a directory that holds `.lns`.

| Before | Now |
|---|---|
| `<config_dir>/lns/config.yaml` | `~/.lns/config.yaml` |
| `~/.lns-connectors.yaml` | `~/.lns/connectors.yaml` |
| `~/.lns-credentials.json` | `~/.lns/credentials.json` |
| `~/.lns-registry-auth.json` | `~/.lns/registry-auth.json` |
| `~/.lns-workload-grants.json` | `~/.lns/workload-grants.json` |
| `~/.lns-host-path-decisions.json` | `~/.lns/host-path-decisions.json` |
| `~/.lns-host-bind-decisions.json` | `~/.lns/host-bind-decisions.json` |
| `<cache_dir>/lns`, `<data_dir>/lns` | `~/.lns/` |

## What you see

`lns config set` names the file that it wrote, and the file is in the one root:

```console
$ lns config set run.cpus 2
Set run.cpus to 2 in /Users/you/.lns/config.yaml

$ lns config list
run.cpus = 2
```

`LNS_HOME` moves the root, and every path follows it:

```console
$ export LNS_HOME=/tmp/demo/home
$ lns config set run.cpus 2
Set run.cpus to 2 in /tmp/demo/home/config.yaml

$ lns volume create some-cache
some-cache

$ find /tmp/demo/home
/tmp/demo/home
/tmp/demo/home/config.yaml
/tmp/demo/home/install-id
/tmp/demo/home/volumes
/tmp/demo/home/volumes/some-cache.img
```

One variable is now enough to isolate a test or a machine.

## What happens to an install that is already there

Nothing happens to it. From the first run, `lns` reads `~/.lns/` only. What the old locations hold stays on disk, and `lns` does not read it. Pull the artifacts again, log in again, and connect the connectors again.

There is no migration code, and there is no warning. The product is pre-1.0, so this is a deliberate clean break.

## Shape

`lns_ipc::paths` is now the one answer to "where does lns keep this on this machine". The six `default_*_path()` functions leave `lns-policy`. That crate keeps the stores, and it takes the path from its caller. So the policy crate is about policy again. No new crate edge was necessary, because every caller of those functions already depends on `lns-ipc`.

Six per-file environment overrides go with them: `LNS_CONFIG_PATH`, `LNS_CONNECTORS_PATH`, `LNS_CREDENTIALS_PATH`, `LNS_REGISTRY_AUTH_PATH`, `LNS_WORKLOAD_GRANTS_PATH`, and the two decision-store variables. §9 names one variable. The e2e harness now sets `LNS_HOME` where it set three `XDG_*` variables, and twelve unit tests lose their inert XDG guards.

`--purge` follows the same collapse. It removed two roots, a configuration directory and three named secret files, and it kept the connector catalog on purpose. §3.9 says that it "deletes everything under `~/.lns/`". So it now removes the one directory, and the directory of the socket. The socket belongs to the service, and not to your data. The guard that refuses to `rm -rf` an arbitrary parent stays, for the socket.

## Tests first

`lns-ipc/src/paths.rs`:

- `~/.lns` under a home directory.
- `LNS_HOME` names the root itself.
- No home and no override is an error that names the way out.
- Every path that `lns` keeps for you is inside the one directory.
- A per-machine file falls back to the working directory. It does not fail a read that only wanted a file that can be absent.

`uninstall/mod.rs`:

- Purge takes the one directory.
- An overridden socket loses only its own file, and the log beside it.

## Docs

`cli-reference.md` gets an **Environment and files** section: the variables, and the table of what lives where. `audit.md`, `credentials.md`, `connectors.md`, `getting-started.md` and the `lns config` / `lns login` / `lns connector` paragraphs name the new paths.

### DIFF
--- a/crates/lns-cli/src/login/real.rs
+++ b/crates/lns-cli/src/login/real.rs
@@ -3,7 +3,6 @@ use std::path::PathBuf;
 use anyhow::{Context, Result, bail};
 use clap::FromArgMatches;
 use lns_ipc::{Request, Response, decode_frame, encode_frame, read_frame_bytes_async};
-use lns_policy::registry_auth::default_registry_auth_path;
 use tokio::io::AsyncWriteExt;
 use tokio::net::UnixStream;
 
@@ -15,7 +14,7 @@ pub fn run_login<'a>(matches: &'a clap::ArgMatches, ctx: RunCtx<'a>) -> RunFutur
     Box::pin(async move {
         let args = LoginArgs::from_arg_matches(matches)?;
         let default_registry = configured_default_registry()?;
-        let auth_path = default_registry_auth_path();
+        let auth_path = lns_ipc::registry_auth_path();
         let verifier = RealRegistryVerifier::new(crate::service::socket_path()?);
         let input = ctx.input;
         let mut out = ctx.out;
@@ -36,7 +35,7 @@ pub fn run_logout<'a>(matches: &'a clap::ArgMatches, ctx: RunCtx<'a>) -> RunFutu
     Box::pin(async move {
         let args = LogoutArgs::from_arg_matches(matches)?;
         let default_registry = configured_default_registry()?;
-        let auth_path = default_registry_auth_path();
+        let auth_path = lns_ipc::registry_auth_path();
         let mut out = ctx.out;
         super::logout(&args, &default_registry, &auth_path, &mut out)
     })
--- a/crates/lns-cli/src/uninstall/mod.rs
+++ b/crates/lns-cli/src/uninstall/mod.rs
@@ -19,7 +19,7 @@ pub struct UninstallArgs {
     #[arg(
         long,
         default_value_t = false,
-        help = "Also delete all local data: cached images and layers, named volumes, the audit trail, config, and stored credentials. Without this, only the program is removed and your data is kept."
+        help = "Also delete everything under ~/.lns: cached artifacts and layers, named volumes, the audit trail, config, connectors, and stored credentials. Without this, only the program is removed and the directory is kept."
     )]
     pub purge: bool,
 
@@ -65,7 +65,6 @@ pub struct UninstallPlan {
     pub binaries: Vec<PathBuf>,
     pub purge_dirs: Vec<PathBuf>,
     pub purge_files: Vec<PathBuf>,
-    pub kept_paths: Vec<PathBuf>,
 }
 
 pub struct Deps<'a, S, C, A, F> {
@@ -200,13 +199,6 @@ async fn purge(fs: &impl Fs, plan: &UninstallPlan, writer: &mut impl Write) -> R
     for file in &plan.purge_files {
         removed(fs.remove_file(file).await, file, writer)?;
     }
-    for kept in &plan.kept_paths {
-        writeln!(
-            writer,
-            "kept {} (your own edits; delete it manually to remove)",
-            kept.display()
-        )?;
-    }
     Ok(())
 }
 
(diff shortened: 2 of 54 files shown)


crates/lns-cli/src/uninstall/mod.rs
```diff
@@ -245,56 +237,27 @@ async fn remove_binaries(
...
-fn push_owned(
-    dirs: &mut Vec<PathBuf>,
-    files: &mut Vec<PathBuf>,
-    path: PathBuf,
-    overridden: bool,
-    sibling: Option<PathBuf>,
-) {
-    match (overridden, path.parent()) {
+pub(crate) fn purge_targets(src: PurgeSources) -> (Vec<PathBuf>, Vec<PathBuf>) {
+    let mut dirs = vec![src.lns_home];
```
jakolehm, line 248: This makes the resolved `LNS_HOME` an unconditional recursive-delete target. Because the override is accepted verbatim, values such as `/`, `$HOME`, or `.` make `uninstall --purge` delete unrelated data, while the prompt still describes local Lens data rather than showing the resolved target. Please reject dangerous roots and display the actual path that will be removed.

crates/lns-cli/src/login/real.rs
```diff
@@ -15,7 +14,7 @@ pub fn run_login<'a>(matches: &'a clap::ArgMatches, ctx: RunCtx<'a>) -> RunFutur
     Box::pin(async move {
         let args = LoginArgs::from_arg_matches(matches)?;
         let default_registry = configured_default_registry()?;
-        let auth_path = default_registry_auth_path();
+        let auth_path = lns_ipc::registry_auth_path();
```
jakolehm, line 17: This saves credentials under the CLI process's `LNS_HOME`, but the running service later reads `registry_auth_path()` using its own environment over the unchanged socket. With an already-running or login-agent-managed service, login can succeed here while subsequent pulls authenticate from another root. The data root needs to be part of the service identity, or a mismatch must be detected and rejected.

crates/lns-ipc/src/paths.rs
```diff
@@ -1,52 +1,91 @@
...
-pub fn cache_root() -> Result<PathBuf, CachePathError> {
-    dirs::cache_dir()
-        .map(|p| p.join("lns"))
-        .ok_or(CachePathError::NoCacheDir)
+pub fn lns_home_with(
+    env: impl Fn(&str) -> Option<OsString>,
+    home: Option<PathBuf>,
+) -> Result<PathBuf, NoLnsHome> {
+    match env("LNS_HOME") {
+        Some(overridden) => Ok(PathBuf::from(overridden)),
```
jakolehm, line 21: Please require `LNS_HOME` to be non-empty and absolute, and do not let the no-home fallback below silently become project-relative. These helpers also feed write paths for registry credentials, connector credentials, and grants, so an empty/relative override or unresolved home can write secrets into the working project and can resolve differently between CLI invocations and the service.

docs/getting-started.md
```diff
@@ -212,7 +212,7 @@ lns uninstall --purge
 ```
 
 `--purge` keeps only files you authored yourself, such as your
-`~/.lns-connectors.yaml` catalog and each project's `lns-local-mixin.yaml`, and prints
+`~/.lns/connectors.yaml` catalog and each project's `lns-local-mixin.yaml`, and prints
```
jakolehm, line 215: This file no longer survives: `--purge` now recursively removes the entire `~/.lns` root, including `connectors.yaml`. Only project-local authored files such as `lns-local-mixin.yaml` remain, so this paragraph needs to stop promising that the connector catalog is kept.

crates/lns-cli/src/uninstall/mod.rs
chenhunghan: Fixed in #300 — `purge_targets` now refuses a data root that is relative, the filesystem root, or your home directory itself (`refusing to purge …: the lns data root must be an absolute directory of its own…`), and the `--purge` confirmation prompt names the resolved directories it will delete before asking. Three new refusal tests plus a prompt test pin it. (commit 81452993)

crates/lns-ipc/src/paths.rs
chenhunghan: Fixed in #300 — `lns_home_with` now rejects empty and relative `LNS_HOME` (`LNS_HOME must be a non-empty absolute path`), and the cwd fallback is gone: the six per-machine path helpers return `Result<PathBuf, LnsHomeError>` and every caller surfaces the error instead of silently writing under the working directory. (commit 81452993)

docs/getting-started.md
chenhunghan: Fixed in #300 — the guide no longer promises `connectors.yaml` survives a purge; it now says purge deletes the whole data directory and only project-local authored files (`lns.yaml`, `lns-local-mixin.yaml`) remain. (commit 81452993)

crates/lns-cli/src/login/real.rs
chenhunghan: Fixed in #302 — the service now owns the credential store: it stores the credential itself after verifying (`RegistryLoginVerified` → `RegistryLoginStored`), and `lns logout` / `lns login --list` go through new `RegistryLogout` / `ListRegistryLogins` requests. The CLI never opens `registry-auth.json` anymore, so a CLI/service LNS_HOME mismatch can no longer strand a verified login where no pull will find it. (commit 50fb34a3)

jakolehm (approved):

PR #283, opened 2026-08-24
feat(lns-cli)!: split `lns artifact` from `lns sandbox`

chenhunghan:
Stage 5 of 9. This PR splits `lns artifact` from `lns sandbox`, as §1.1–§1.3 and §3.1 name them.

> **Stacked on #282.** The base of this PR is `cli-spec-s4b-one-root`. The diff shows stage 5 only. Merge #282 first.

## The split

Two things carry a name, and every command acts on one of them. An **artifact** is one `lns.run/v1` document, published as one OCI artifact. A **sandbox** is what a sandbox artifact becomes when it runs.

| Namespace | Verbs |
|---|---|
| `lns artifact` | `init` `validate` `push` `pull` `tag` `ls` `inspect` `rm` `prune` |
| `lns sandbox` | `run` `exec` `kill` `stop` `logs` `attach` `ls` `inspect` `rm` |

`lns sandbox run <REF>` is the one command that crosses from one namespace to the other.

These verbs leave `lns sandbox`: `init`, `validate`, `push`, `pull`, `tag`, `prune` and `ps`.

## What you see

The two `ls` verbs now answer two different questions:

```console
$ lns artifact ls
ARTIFACT  KIND  DIGEST  SIZE  HOLDER

$ lns sandbox ls
ID  NAME  IMAGE  STATE  CPU %  MEM
```

`ls` is the verb that §3.2 names, and `lns ps` is its shortcut. So `lns sandbox ps` is gone:

```console
$ lns sandbox ps
error: unrecognized subcommand 'ps'

Usage: lns sandbox [OPTIONS] <COMMAND>

```

## §1.3 arbitration

`rm` and `inspect` are in both namespaces. Their shortcuts work out which namespace you mean:

1. An operand that looks like a path is a local document, so it is the artifact. `lns` does not ask the service.
2. If it is not a path, `lns` looks the word up in both namespaces.
3. One match runs.
4. Two matches refuse, and `lns` prints the two namespaced commands.
5. No match is an error that names both namespaces as searched.

```console
$ lns rm ghost
Error: no sandbox and no cached artifact named "ghost"; `lns sandbox ls` lists the
sandboxes and `lns artifact ls` the cache

$ echo $?
1
```

A word that both namespaces know is refused, and the message names the two commands:

```text
$ lns rm hermes
error: "hermes" is both a sandbox and a cached artifact; say which with
`lns sandbox rm hermes` or `lns artifact rm hermes`
```

This replaces the old fallback chain. That chain tried the run first, then fell through to the cache without a word. It could never see that both namespaces answered, and a miss named only one of them.

`prune` has no shortcut, because it takes no operand, so nothing could tell the two namespaces apart. `validate` and `ls` also have no shortcut.

A known seam at this stage: the artifact half of the lookup still asks the service to inspect the reference, and the service resolves that against a registry. The first commit of #284 points it at the cache (`ListImages`), where it belongs.

## The module tree follows the surface

`author`, `distribute`, `fileset` and `mixin_plan` move under `artifact/`, together with the verbs that drive them and with their tests. Three items move to a place that both namespaces can reach, so that neither namespace depends on the other:

- `SandboxService` and `TermInfo` → `service::client`
- `RealSandboxService` → `service::real`
- `CannedService` and its fixtures → a shared `test_service`
- `format_bytes` → `output`, where every table renders

`sandbox/mod.rs` goes from 2700 lines to about 1000.

## Tests

All Layer 2 scenarios pass. Five are new, for the arbitration: a path-shaped operand, one match in each namespace, two matches, and no match. Layer 3 tests cover the resolution itself. `sandbox_surface.feature` is written again around the two namespaces: what each `--help` holds, what left the sandbox namespace, and that `lns sandbox ps` is now an unrecognized subcommand. The Layer 1 features move to the namespaces too.

## Docs

`cli-reference.md` — "The sandbox surface" becomes "Two namespaces", with the full shortcut table and the arbitration rules. It gets a new `lns artifact` section, and `lns sandbox` is cut to the instance verbs. `running-workloads.md` opens on the two nouns, and not on two tiers. `getting-started.md` and `policy.md` follow the verbs that moved.


crates/lns-cli/src/shortcut.rs
```diff
@@ -0,0 +1,152 @@
...
+        settled => Ok(settled),
+    }
+}
+
+/// A path-shaped operand is a local document, so it never needs the service asked about it.
+pub fn names_a_document(operand: Option<&str>) -> bool {
+    crate::artifact::author::is_local_inspect(operand)
+}
+
+async fn ask_both(svc: &impl SandboxService, operand: &str) -> Result<Owner> {
```
jakolehm, line 47: `ask_both` treats every response other than the expected success variant as a namespace miss. `InspectRun` also returns `Response::Error` for an ambiguous run-id prefix, so if the artifact probe succeeds, `lns rm <prefix>` can delete that artifact instead of refusing the ambiguity. Please classify only explicit not-found responses as misses and surface ambiguity, operational errors, and unexpected responses.

docs/policy.md
```diff
@@ -218,7 +218,7 @@ section before using it.
 No command edits the rules. A decision is recorded one of two ways: you answer the
 card a run raises, or you open `lns-local-mixin.yaml` and write the rule yourself.
 Both write the same document, and because it is a `kind: mixin` document,
-`lns sandbox validate -f lns-local-mixin.yaml` checks what you wrote and
+`lns artifact validate -f lns-local-mixin.yaml` checks what you wrote and
```
jakolehm, line 221: The next line still recommends `lns sandbox inspect -f`, but sandbox inspection no longer accepts `-f`; this must be `lns artifact inspect -f`. The same namespace migration remains unfinished in `docs/sandbox-spec.md:1410`, `:1471`, and `:1702`, which still name the removed sandbox authoring commands.

crates/lns-cli/src/shortcut.rs
```diff
@@ -0,0 +1,152 @@
...
+
+pub const RM_SPEC: CommandSpec = CommandSpec {
+    name: "rm",
+    augment: augment_rm,
+    run: real::run_rm,
+    announces_update_check: true,
+    owns_terminal: crate::command::never_owns_terminal,
+};
+
+pub fn augment_inspect(app: clap::Command) -> clap::Command {
```
jakolehm, line 85: Building the shortcut from `artifact::InspectArgs` rejects `lns inspect <RUN> --format json` in Clap before owner arbitration, even though §1.3 defines the shortcut as an exact alias and `lns sandbox inspect` accepts this option. The shortcut needs to preserve the sandbox-only output option and validate it after resolving the owner.

crates/lns-cli/src/shortcut/real.rs
```diff
@@ -0,0 +1,76 @@
...
+        let svc = crate::service::real::RealSandboxService::new(crate::service::socket_path()?);
+        match which(&svc, "inspect", &reference).await? {
+            Owner::Artifact => {
+                crate::artifact::real::dispatch(
+                    crate::artifact::ArtifactCommand::Inspect(args),
+                    ctx.input,
+                )
+                .await
+            }
+            _ => {
```
jakolehm, line 58: When this resolves to a live sandbox, the parsed `--mixin` values are discarded. `lns inspect reviewer --mixin ...` therefore succeeds while describing the already-running sandbox as if the flag applied. Preserve the prior explicit refusal for this combination rather than silently dropping user input.

crates/lns-cli/src/artifact/mod.rs
```diff
@@ -0,0 +1,1476 @@
...
+pub(crate) async fn inspect_cached<W: std::io::Write>(
+    svc: &impl SandboxService,
+    reference: &str,
+    mixins: &[String],
+    out: &mut W,
+) -> Result<i32> {
+    match svc
+        .one_shot(Request::InspectImage {
+            image: reference.to_string(),
+            mixins: mixins.to_vec(),
```
jakolehm, line 601: Published artifact inspection now forwards user-named mixins unchanged. With `--mixin ./obs`, the relative path reaches the service and is explicitly refused by `mixin::locate`; the previous inspect wiring rooted these paths first. Please root them against `ctx.cwd()` before dispatch for both the namespaced command and shortcut.

crates/lns-cli/src/shortcut.rs
chenhunghan: Fixed in #301 (stage 17, on top of the stack so no rebase below): `InspectRun` now answers a typed `Response::RunUnknown` for a genuine miss, backed by a typed `ResolveError` in `run_registry`, and `ask_both` treats only that as a namespace miss — an ambiguous prefix, a daemon failure, or an unexpected reply surfaces instead of settling on the artifact. Written red first, reproducing exactly this failure (`Ok(Artifact)` on an ambiguous prefix).

docs/policy.md
chenhunghan: Fixed in #301: `docs/policy.md` now says `lns artifact inspect -f`, and the three `docs/sandbox-spec.md` references (`lns sandbox validate` ×2 and the `lns sandbox init, validate, inspect, push` cross-reference) now use the `lns artifact` spellings cli-spec settled.

crates/lns-cli/src/shortcut.rs
chenhunghan: This was repaired in #288 (stage 9, "Second repair"): the shortcut now carries both namespaces' flags — `--format` parses again — and a flag the settled target does not take is refused by name with exit 2 after owner arbitration, with a truth-table test and a Layer 1 feature pinning it.

crates/lns-cli/src/shortcut/real.rs
chenhunghan: Also repaired in #288: a live-sandbox target now refuses `--mixin` explicitly ("`--mixin` previews a document's composition, so `lns sandbox inspect` takes no `--mixin`…", exit 2) instead of silently dropping it.

crates/lns-cli/src/artifact/mod.rs
chenhunghan: Fixed in #301: both the namespaced `lns artifact inspect` and the shortcut now root path-shaped `--mixin` values against the invocation directory before dispatch — the same `root_named_directories` the run path uses — via the new `InspectArgs::root_mixins`; registry references pass untouched. A new Layer 2 scenario pins that the service receives the absolute path.

jakolehm (approved):

PR #284, opened 2026-08-24
feat(lns-cli)!: give the CLI the stopped sandbox — start, rm -f, ls -a, prune

chenhunghan:
Stage 6 of 9. This PR gives the CLI the stopped sandbox: `start`, `rm -f`, `ls -a` and `prune`, as §1.1 and §3.2.2 name them.

> **Stacked on #283.** The base of this PR is `cli-spec-s5-artifact-namespace`. The diff shows stage 6 only: the closed `lns-start` branch (PR #263), rebased in as its 15 commits, plus the work that makes it agree with the split surface. Merge #283 first.

## What the branch brings

§1.1 says: "A sandbox outlives its workload — when the workload exits, the sandbox stays as **stopped** until you remove it, and `lns start` runs it again."

The branch delivers that: the durable run record, a registry that holds stopped entries across a service restart, a replay that boots back onto the preserved writable layer, fail-closed conflict checks, the prune sweep, and the audit events for the full lifecycle. All of it lands as the branch wrote it.

## What you see

`ls` gets a `STATE` column, and `-a` includes the stopped sandboxes:

```console
$ lns sandbox ls
ID  NAME  IMAGE  STATE  CPU %  MEM

$ lns sandbox ls -a
ID  NAME  IMAGE  STATE  CPU %  MEM
```

A stopped sandbox has no guest, so `lns` lists it **without** a statistics probe. It does not list it with a failed probe. A scenario counts the probes and pins this.

`prune` sweeps the stopped sandboxes. It asks first, and with no terminal to ask at it refuses:

```console
$ lns sandbox prune
Error: this removes every stopped sandbox, writable layers included; there is no
terminal to ask at, so pass --force to confirm
```

`lns start` runs a stopped sandbox again:

```console
$ lns start --help
Run a stopped sandbox again (shortcut for `lns sandbox start`).

Usage: lns start [OPTIONS] <RUN>

Arguments:
  <RUN>  Stopped sandbox id or name to run again.

Options:
  -a, --attach                     Attach to the sandbox's output and adopt the workload's exit code.
  -i, --interactive                With -a: forward stdin to the workload.
      --detach-keys <DETACH_KEYS>  With -a: detach chord; on match the CLI detaches, leaving
                                   the sandbox running. [default: ctrl-p,ctrl-q]
```

## What this PR writes again

The headline CLI change of the branch was a rename: `lns rm` for runs, and `lns rmi` for images. Stage 5 did that job by a different route. `lns artifact rm` **is** `rmi`, and `lns sandbox rm` is the `rm` of the branch. So this PR drops `rmi`, and the `rm` behaviour of the branch lands on `lns sandbox rm`:

- It asks the service to remove the sandbox. It does not inspect first and then decide on a snapshot that can be stale when it acts. The service owns the answer to "is this running".
- `-f`/`--force` stops the sandbox first.

## The §1.3 artifact lookup asks the cache, not a registry

The artifact namespace **is** the local store, so the lookup asks what the cache holds. A first version asked the service to inspect the reference, and the service resolves a reference against a registry. Every short word is a real image on Docker Hub, so that version made `lns rm ghost` say `no such image: docker.io/library/ghost:latest`, and no sandbox could carry a name such as `redis`.

This PR's first commit makes the lookup read `ListImages` and compares the operand with what the cache stores. It compares the reference as you typed it, and also the normalized form that the cache stores.

A Layer 2 scenario could not find this fault, because the fake service answers whatever the scenario stages. A run against the real service found it. There is now a scenario for the behaviour, and four Layer 3 tests for the lookup.

## Two faults that the rebase found

**The audit chain of a sandbox was inside the directory that `rm` deletes.** Before stage 4b, the chain was under `<data_dir>/lns/runs/`, and the writable layer was under `<cache_dir>/lns/runs/`. Those are different roots, so a removal could not touch the chain. The collapse to one root made them the same subtree, without a word. The removal test of the branch found it.

§3.6 says: "A sandbox's chain also outlives the sandbox — removing one does not remove what it did." So the chains move to `~/.lns/audit/<id>/`, and the writable layer keeps `~/.lns/runs/<id>/`. A test in `lns-ipc` pins that the two are not the same subtree, so the invariant is structural, and not remembered.

**`handle_request` went over the cognitive-complexity gate.** Its eight `unreachable!` arms all say the same thing: this request has a streaming handler of its own. They collapse into one arm. The five volume verbs move to `handle_volume_request`.

## Rebase resolutions worth a look

- `upperfs::provision_image` — both sides wanted something. `main` wanted the size parameter, and the branch wanted the guard that never reformats an existing layer. Both are in, with both tests.
- `run_registry` — the branch makes entries `Live | Stopped`, so the exec-session accessors that `main` added now go through `as_live`.
- `ipc/adapter.rs` — the branch divides registration into a fresh path and a restart path. The `exec_sessions` field of `main` is carried into both.
- The documentation rewrites of the branch are written against the surface before the split, so this PR reconciles them by hand instead of by merge.


crates/lns-cli/src/sandbox/mod.rs
```diff
@@ -187,6 +248,15 @@ macro_rules! shortcut_spec {
     };
 }
 
+shortcut_spec!(
+    augment_start,
+    START_SPEC,
```
jakolehm, line 253: `START_SPEC` is defined here but never added to `command::registry()`, so the real binary rejects `lns start <RUN>` as an unknown command. The Layer 2 harness masks this by rewriting top-level starts to `lns sandbox start`; please register the spec and exercise the real `build_cli` path.

crates/lns-cli/src/shortcut/real.rs
```diff
@@ -24,6 +24,7 @@ pub(super) fn run_rm<'a>(matches: &'a clap::ArgMatches, ctx: RunCtx<'a>) -> RunF
                         command: crate::sandbox::SandboxCommand::Rm(
                             crate::sandbox::SandboxRmArgs {
                                 run: args.reference,
+                                force: false,
```
jakolehm: Top-level `rm` still parses `artifact::RmArgs`, which has no `-f`, and this branch hard-codes `force: false`. Consequently `lns rm -f <RUN>` exits 2 instead of acting as the documented exact shortcut for `lns sandbox rm -f <RUN>`.

crates/lns-service/src/ipc/mod.rs
```diff
@@ -504,20 +593,110 @@ fn inspect_resolved(id: &str) -> Response {
...
+            })
+            .await
+    {
+        return Response::Error { message };
+    }
     match crate::run_registry::remove_if_exited(id) {
-        crate::run_registry::RemoveOutcome::Removed => Response::Acknowledged,
+        crate::run_registry::RemoveOutcome::Removed => {
+            note_removed(id, force);
+            crate::run::reclaim_run_dir(remover, cache_root, id);
```
jakolehm: The registry entry is already gone and `reclaim_run_dir`’s failure result is ignored, so `lns rm` reports success and frees the name even when `record.json` and `upper.img` remain. On the next service restart that record resurrects the supposedly removed sandbox, potentially alongside a new sandbox using its name. Make successful/tombstoned durable deletion a condition of acknowledgment.

crates/lns-service/src/ipc/mod.rs
```diff
@@ -526,6 +705,43 @@ fn remove_resolved_run(id: &str) -> Response {
...
+where
+    F: crate::image_store::Fs,
+    R: crate::run::RemoveDir,
+    N: Fn(&[String]),
+{
+    let mut removed = crate::run_registry::prune_exited();
+    for id in &removed {
+        crate::run::reclaim_run_dir(remover, cache_root, id);
+    }
+    let known: std::collections::HashSet<String> = crate::run_registry::snapshot()
```
jakolehm, line 742: This snapshot is not coordinated with the asynchronous directory scan. A fresh run registered after the snapshot can create its run directory before `read_dir` completes; it is then absent from `known` and deleted as an orphan while still running. Revalidate each candidate atomically against registration, or protect registration and orphan reclamation with the same lease.

crates/lns-cli/src/shortcut.rs
```diff
@@ -52,15 +63,10 @@ async fn ask_both(svc: &impl SandboxService, operand: &str) -> Result<Owner> {
...
-    let is_an_artifact = matches!(
-        svc.one_shot(Request::InspectImage {
-            image: operand.to_string(),
-            mixins: Vec::new(),
-            decisions: None,
-        })
-        .await?,
-        Response::ImageInspected { .. }
-    );
+    let is_an_artifact = match svc.one_shot(Request::ListImages).await? {
```
jakolehm, line 79: Artifact arbitration can falsely conclude there is no artifact: `cache_holds` normalizes bare names with Docker Hub defaults instead of `run.registry`/`hub.lns.run`, and every non-`ImageList` response is converted to `false`. With a matching sandbox, `lns rm` can therefore delete it without detecting the cached artifact or when cache listing failed. Qualification must match the CLI default and lookup errors must fail closed.

crates/lns-service/src/run_record.rs
```diff
@@ -0,0 +1,381 @@
...
+
+pub async fn load_all_with<F: Fs>(fs: &F, cache_root: &Path) -> Result<Vec<RunRecord>> {
+    let entries = match fs.read_dir(&cache_root.join("runs")).await {
+        Ok(entries) => entries,
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => return Ok(Vec::new()),
+        Err(e) => return Err(anyhow::Error::from(e).context("listing the runs dir")),
+    };
+    let mut records = Vec::new();
+    for dir in entries {
+        if let Ok(bytes) = fs.read(&dir.join("record.json")).await
```
jakolehm: Unreadable, malformed, and future-version records are silently omitted here. `sandbox prune` then sees their directories as unregistered orphans and deletes their preserved writable layers, turning recoverable or downgrade-visible state into data loss. Keep damaged entries protected and surface their damage instead of treating them as absent.

crates/lns-service/src/run_record.rs
```diff
@@ -0,0 +1,381 @@
...
+            && let Ok(record) = serde_json::from_slice::<RunRecord>(&bytes)
+            && record.version == CURRENT_VERSION
+        {
+            records.push(record);
```
jakolehm: This also restores records whose original request had `auto_remove: true`. If the service exits before normal completion cleanup, a sandbox created with `lns run --rm` comes back as restartable stopped state, contradicting the documented opt-out from persistence. Auto-remove records need cleanup rather than registration during rebuild.

crates/lns-ipc/src/paths.rs
```diff
@@ -40,8 +40,9 @@ pub fn short_run_id(id: &str) -> &str {
     id.char_indices().nth(12).map_or(id, |(i, _)| &id[..i])
 }
 
+/// A sandbox's chain outlives the sandbox, so it cannot live in `runs/`, which is what removing one deletes.
 pub fn audit_runs_root() -> Result<PathBuf, NoLnsHome> {
-    Ok(lns_home()?.join("runs"))
+    Ok(lns_home()?.join("audit"))
```
jakolehm, line 45: Changing the audit root without migrating or reading the old `~/.lns/runs/<id>` location makes existing chains disappear. Worse, `sandbox prune` classifies those legacy audit-only run directories as orphans and deletes them. Migrate the chain before enabling the orphan sweep so the audit history actually outlives the sandbox.

crates/lns-service/src/ipc/mod.rs
```diff
@@ -30,6 +30,82 @@ pub trait RunHost {
...
+where
+    S: AsyncReadExt + AsyncWriteExt + Unpin + Send,
+    H: StartHost,
+{
+    let Ok((run_id, status)) = crate::run_registry::resolve_status(run) else {
+        let _ = write_error(stream, no_startable_run(run)).await;
+        return Ok(());
+    };
+    match status {
+        lns_ipc::RunStatus::Running => {
```
jakolehm, line 88: For `start -a` on an already-running sandbox, this sends the attach handshake and immediately closes the stream. The CLI then enters its attached-session driver and fails on EOF instead of attaching or returning cleanly. Route the request through attach handling, or reject the attached form explicitly.

lns-policy.yaml
```diff
@@ -0,0 +1,19 @@
+network:
+  egress:
+    http:
+    - match: registry.npmjs.org
```
jakolehm: This and the two other newly added `lns-policy.yaml` files look like harness/generated project state, not lifecycle fixtures; this root copy also grants real connectors and broad direct egress. `crates/e2e-tests/lns-policy.yaml` was previously removed as a stray harness-generated file. Please remove these unrelated policy files from the PR.

crates/lns-service/src/run/orchestrator.rs
```diff
@@ -77,18 +77,56 @@ pub async fn handle(
...
-    let result = orchestrate(run_id, microvm, args, prepared, frame_tx.clone(), input_rx)
-        .instrument(tracing::Span::current())
-        .await;
+    let microvm_label = microvm.clone();
+    let result = orchestrate(
+        run_id,
+        microvm,
+        args,
+        prepared,
+        mode,
```
jakolehm, line 92: The new restart/removal lifecycle exposes an ordering hazard later in this call: `orchestrate` publishes `Exited` via `set_exit_code` before `shutdown_after_session` releases the VM and its resources. `stop` can return and `start`, `rm`, or `prune` can then reuse or unlink `upper.img` while the old VM still owns it. Publish the stopped state only after the VM is fully quiesced.

crates/lns-cli/src/sandbox/mod.rs
chenhunghan: Fixed — `START_SPEC` is now in the `command.rs` registry, the test-harness rewrite that masked the gap is deleted, and a grammar scenario pins that `lns start --help` answers as the shortcut for `lns sandbox start`. (commit 181eecba)

crates/lns-cli/src/shortcut/real.rs
chenhunghan: Fixed — the top-level `lns rm` now takes `-f` and routes it through to `SandboxRmArgs`; an artifact operand with `-f` is refused with a message saying force only applies to sandboxes. Scenarios cover the forced passthrough, the artifact refusal, and the grammar. (commit 4b5f9d9b)

crates/lns-service/src/ipc/mod.rs
chenhunghan: Fixed — `remove_run_with` now acks only after `reclaim_run_dir` reports the files durably gone; when deletion fails the entry is restored into the registry and the client gets an error naming the surviving files, so the run stays listed and cannot resurrect beside a namesake. (commit b14c6c72)

crates/lns-service/src/ipc/mod.rs
chenhunghan: Fixed — the sweep now revalidates each candidate against the live registry right before reclaiming (registration precedes scratch-dir creation, so the fresh check closes the window), on top of the record.json guard from the damaged-records fix. A test simulates a run registering mid-sweep and pins that its dir survives. (commit e67d16bf)

crates/lns-cli/src/shortcut.rs
chenhunghan: Fixed — `cache_holds` now qualifies bare names through the configured default registry (`run.registry`, else `hub.lns.run`) instead of the Docker Hub rules, and arbitration fails closed: an Error or unexpected reply from the cache listing aborts the shortcut instead of silently treating the cache as empty. (commit fb7f37de)

crates/lns-service/src/run_record.rs
chenhunghan: Fixed — `load_all_with` now returns damaged records (unreadable, unparseable, wrong version) alongside the good ones instead of dropping them; the orphan sweep only reclaims a dir whose record.json is confirmed absent, anything else is warned about and left alone; startup logs each damaged run. (commit ea531b8d)

crates/lns-service/src/run_record.rs
chenhunghan: Fixed — rebuild no longer registers records with `auto_remove: true`; it hands them back as doomed and startup reclaims their dirs, so a `--rm` run that died with the service is cleaned up instead of revived as a stopped sandbox. (commit c2ff5149)

crates/lns-ipc/src/paths.rs
chenhunghan: Fixed — service start now migrates each legacy `runs/<id>/audit.jsonl` + `audit.anchor` into `audit/<id>/` before any prune can run; an established chain at the new home is never overwritten, and a chain that cannot move is left in place with a warning rather than destroyed. (commit 4f3a5cd9)

crates/lns-service/src/ipc/mod.rs
chenhunghan: Fixed — an attached `start` of an already-running run is now refused with an error pointing at `lns attach <run>` instead of answering a bare RunStarted and closing the stream; the detached form still answers RunStarted. (commit e3dc1cb1)

lns-policy.yaml
chenhunghan: Fixed — the three stray `lns-policy.yaml` files are untracked and the name is gitignored repo-wide so harness runs cannot re-add one. (commit 264b0372)

crates/lns-service/src/run/orchestrator.rs
chenhunghan: Fixed — the exit is now published only after `shutdown_after_session` awaits the VM quiesce (`publish_exit_after_quiesce`), so rm/start/prune cannot grab the writable layer while the microVM still holds it; a failed quiesce still publishes the exit so the run never sticks as Running. (commit 2cd35652)

jakolehm (approved):

PR #288, opened 2026-08-25
fix(lns-cli)!: make the run, exec, and inspect shortcuts exact aliases again

chenhunghan:
A follow-up to the nine-stage cli-spec series. This PR makes the `lns sandbox` spellings of `run` and `exec` answer `125` for a pre-start failure, the way their top-level shortcuts already do.

> **Stacked on #286.** The base of this PR is `cli-spec-s8-connector`. The diff shows stage 9 only. Merge #286 first.

## The gap

§5 says a `run` or `exec` that fails before the workload starts is `lns` failing, never the workload — exit `125`, so it is never mistaken for a workload that exited `1`. §1.3 says a shortcut and its namespaced spelling are the same command — same arguments, same output.

`lns run` and `lns exec` kept that promise. Their `lns sandbox` spellings did not: the alias paths bubbled a pre-start error straight to the generic exit `1`.

```console
$ lns run some-image        # service not running
Error: Lens Sandbox is not running. Run `lns service start` to start it.
$ echo $?
125
$ lns sandbox run some-image
Error: Lens Sandbox is not running. Run `lns service start` to start it.
$ echo $?
1        # ← before this PR
```

## The fix

`sandbox::real::run` now wraps both alias launches in the same `as_pre_start_failure` the top-level `run` and `exec` specs use. The `Exec` interception moves up from `dispatch` to sit beside `Run`, so both spellings share one shape: parse, wrap, launch. `as_pre_start_failure` becomes `pub` so the sandbox wiring can name it.

## Tests

The fix sites are production wiring (`sandbox/real.rs`, `orchestrator.rs` — both IGNORES-exempt), so the pin is Layer 1: a new `pre_start_exit_code.feature` drives all four spellings against a dead service socket and asserts `125` for each. Written red first — the two alias scenarios failed with exit `1` before the fix.

## Second repair: the inspect shortcut lost `--format` in stage 5

Stage 5 gave the `lns inspect` shortcut the artifact namespace's arguments, and `lns artifact inspect` takes no `--format`. So `lns inspect <RUN> --format json` became a parse error, while `lns sandbox inspect <RUN> --format json` works — the same §1.3 break, in the other direction.

The shortcut now carries both namespaces' flags, and refuses the one the settled target does not take — by name, not by parser accident:

```console
$ lns inspect ./lns.yaml --format json
error: a local document renders as its author wrote it, so `lns artifact inspect`
takes no `--format`; for a sandbox: `lns sandbox inspect <RUN> --format json`
$ echo $?
2
```

- A **sandbox** target takes `--format` and passes it through; `--mixin` is refused there the same way, instead of being dropped without a word.
- A **document** or **cached artifact** target refuses `--format`, saying why, exit `2` — the usage-error code its namespaced spelling answers.

The decision (`InspectTarget` × flags → refusal) lives in `shortcut.rs` under the 100% gate, with Layer 3 truth-table tests. A new Layer 1 `inspect_shortcut_flags.feature` pins the document refusal through the real binary (red first: the old message was clap's `unexpected argument`). The stage-5 parse pin flips to assert the shortcut now parses `--format`.

Also swept in: a stale comment in `lns-policy/src/grants.rs` still naming the `LNS_WORKLOAD_GRANTS_PATH` variable stage 4b deleted.

## Third repair: the `add` hint names a spelling the CLI recognizes

A successful `lns connector add` closed with "Connect it to a project with `lns connect <id>`." — but there is no top-level `lns connect`, and §3.3 defines no such shortcut. The next step is now spelled in full: `lns connector connect <id>`. A Layer 3 test pins the hint (red first: the old text named the spelling that does not parse).


chenhunghan:

PR #289, opened 2026-08-25
fix(lns-cli)!: give ls one spelling — the list alias is removed

chenhunghan:
A follow-up to the cli-spec series, stacked on #288.

`lns artifact ls` and `lns sandbox ls` carried a visible `list` alias. The specification names no such spelling: §3.1 and §3.2.2 say `ls`, and §10 says an old spelling is removed rather than quietly kept working. (`lns connector list` and `lns config list` are unaffected — there, `list` **is** the verb the spec names.)

Two parse pins in `parse_errors.feature` went red first — both spellings parsed — and the alias is gone. The flag aliases `--memory` and `--mount` stay: §3.2.1 names them.

> **Stacked on #288.** Merge #288 first.


chenhunghan:

PR #290, opened 2026-08-25
feat(lns-cli): say which command takes a document when a RUN verb gets one

chenhunghan:
Stacked on #289. Closes the last cheap gap from the cli-spec conformance pass.

## The gap

cli-spec §2.4: *"a command takes a REF or a RUN, never either. Given the wrong kind, it says which it wanted and which command takes the one you typed."* Before this, `lns sandbox inspect ./lns.yaml` asked the service for a run named `./lns.yaml` and surfaced whatever error came back — or 125 if no service was running.

A path-shaped operand (`.`, `..`, `./…`, `../…`, `/…`, `lns.yaml`, `…/lns.yaml` — the same `is_definition_path` test the run/shortcut arbitration already uses) names a document, and a document is never a RUN. That makes the RUN verbs the one deterministic wrong-kind case, so they now answer:

```
error: `lns sandbox inspect` takes a RUN — a sandbox id or name — and "./lns.yaml" names a document; `lns artifact inspect` is what reads one
```

exit 2 (unparsable invocation), on stderr, client-side — before `require_running`, so it never depends on a service.

## What's covered

- **Decision** — `document_refusal` / `wrong_kind_refusal` in `sandbox/mod.rs`, 100%-gated: Layer 3 truth table over all eight RUN verbs (exec, kill, stop, start, inspect, logs, attach, rm) plus the boundary (`v1.2-agent` is a name, `ls` has no operand); Layer 2 scenarios pin the user-visible refusal for inspect/rm/logs and that a dotted *name* still reaches the service.
- **Wiring** — three IGNORES-exempt interception points: `sandbox/real.rs` `run()` (the `lns sandbox …` namespace incl. exec), `dispatch_command` (the top-level `lns start/stop/kill/rm/logs/attach` wrappers), and `orchestrator.rs` `exec_command` (top-level `lns exec`). Pinned by a Layer 1 feature proving `lns exec ./lns.yaml -- true` and `lns sandbox logs ./lns.yaml` exit 2 with **no service running** — the check demonstrably precedes any socket contact.
- The Layer 2 harness now folds captured stderr into the assertable output, matching where §4.1 says errors go.

## Not in scope

The REF verbs have no deterministic converse: a bare word is a valid REF, so `lns artifact inspect reviewer` can't know `reviewer` is a run name without asking — the service's not-found answer stays the right response there. `lns run` takes a REF and a document *is* one of its operand kinds, so it's untouched. The top-level `rm`/`inspect` shortcuts already arbitrate path-shaped operands into the artifact namespace (#288).

## Gates

`make lint`, 440 Layer 2 scenarios, full coverage gate, targeted `make e2e` — all green.

### DIFF
--- a/crates/e2e-tests/features/wrong_kind_operand.feature
+++ b/crates/e2e-tests/features/wrong_kind_operand.feature
@@ -1,0 +1,20 @@
+Feature: a RUN verb given a document redirects before touching the service
+
+  §2.4: given the wrong kind of operand, a command says which it wanted
+  and which command takes the one you typed. The redirect is a client-side
+  decision, so it answers 2 even when no service is running — never 125.
+
+  Scenario: lns exec given a document names the artifact reader instead
+    Given a clean lns cache home
+    And no Lens Sandbox service is running
+    When I run "lns exec ./lns.yaml -- true"
+    Then the exit code is 2
+    And the output contains "takes a RUN"
+    And the output contains "lns artifact inspect"
+
+  Scenario: lns sandbox logs given a document is the same refusal
+    Given a clean lns cache home
+    And no Lens Sandbox service is running
+    When I run "lns sandbox logs ./lns.yaml"
+    Then the exit code is 2
+    And the output contains "lns artifact inspect"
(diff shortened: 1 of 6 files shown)


chenhunghan:

PR #296, opened 2026-08-25
fix(lns-cli): list what a prune would remove before asking

### LINKED ISSUE #291: prune asks without listing what it would remove
cli-spec is explicit, in all three homes, that prune shows its work before asking:

- §3.1 / §3.2.2 — artifact + sandbox prune: "Removes every cached artifact nothing holds. **Lists them and asks**, unless `-f`/`--force`." / "Removes every stopped sandbox, writable layers included. **Lists them and asks**, unless `-f`/`--force`."
- §3.6 — volume prune: "Removes every volume no sandbox holds. **Lists them and asks**, unless `-f`/`--force`."

Today all three prunes ask a bare yes/no (or refuse without a terminal) without first printing the candidate list, so the user confirms a deletion they haven't seen. The fix is the same shape in each namespace: fetch the candidates, print them (the same columns the namespace's `ls` uses), then ask; empty set short-circuits to "nothing to remove" without a prompt.

Layer 2 scenarios per namespace: the list precedes the prompt, `-f` skips both, empty set never prompts.

chenhunghan:
Stacked on #293.

The cli-spec prune rows (§ artifact `prune`, § sandbox `prune`, § volume `prune`) all say **"Lists them and asks, unless `-f`/`--force`"** — but every prune asked blind. This closes the gap for all three namespaces.

## What changed

- **The candidate list rides with the prompt on stderr.** The list is context for the question, not the command's answer, so per §4.1 it lands on the same stream as the prompt. A piped stdout still shows both.

  ```
  Would remove:
    hermes:1.0
    scribe:1.0
  This removes every cached sandbox not held by a running one ... Continue? [y/N]
  ```

- **Artifact candidates come from the service** via a new `Request::ListPrunableImages` (reply reuses `Response::ImageList`). The CLI cannot compute them from `ls` data: `ImageInfo.in_use_by` only names a *running* holder, while prune keeps the whole dependency closure of the actively-held roots — a base image whose sandbox is running shows `in_use_by: None` but is not removable. The service-side partition is extracted from `prune_with` into `removable_partition`, shared by prune and the new listing, so the list is exactly what the sweep would remove.

- **Sandbox and volume candidates reuse existing requests** (`ListRuns` filtered to stopped; `ListVolumes` filtered to `in_use_by: None`) — both match the service's own prune semantics exactly.

- **Empty candidate sets:** a sandbox or volume prune with nothing to remove prints the existing no-op line (`No stopped sandboxes.` / `No unused volumes.`) and never asks. An artifact prune still asks with nothing removable, because it also reclaims the provisioned tool cache when no sandbox is live — the prompt already says so.

- **Unchanged:** `-f`/`--force` skips the listing with the question; the §7.2 non-tty refusal still fires before any service call (the "no request reached the service" pins survive verbatim).

## Tests (red-first)

- New Layer 2 feature `prune_lists.feature` (8 scenarios): listing precedes the prompt on stderr (order-asserted), held candidates excluded, empty sets never prompt, `--force` sends no listing request.
- Service Layer 2: three new `image_management.feature` scenarios pinning that the prunable listing names only what prune would remove — including a base image protected solely through a live sandbox's dependency edge — and touches nothing.
- Layer 3: `prunable_references` / `stopped_run_names` / `unused_volume_names` sorting, filtering, short-id fallback, daemon-error and unexpected-variant paths; IPC round-trip pin for the new request; `handle_request` wiring pin.
- Amended pins: the decline scenarios now assert "no `PruneImages`/`PruneRuns` request" (a listing request legitimately precedes the prompt).

`make lint`, `make complexity`, and the full coverage gate are green; 449 lns-cli scenarios + 912 unit tests, 274 lns-service scenarios pass.

Closes #291

### DIFF
--- a/crates/lns-cli/src/artifact/mod.rs
+++ b/crates/lns-cli/src/artifact/mod.rs
@@ -545,6 +545,8 @@ async fn prune<I: std::io::BufRead, W: std::io::Write, E: AsyncWriteExt + Unpin>
                 "this removes every cached sandbox not held by a running one and, when none is live, the provisioned tool cache; there is no terminal to ask at, so pass --force to confirm"
             );
         }
+        let candidates = prunable_references(svc).await?;
+        crate::output::announce_prune_candidates(&candidates, stderr).await?;
         if !confirm_prune(input, stderr).await? {
             return Ok(0);
         }
@@ -564,6 +566,19 @@ async fn prune<I: std::io::BufRead, W: std::io::Write, E: AsyncWriteExt + Unpin>
                 crate::output::format_bytes(reclaimed_bytes)
             )?;
             Ok(0)
+        }
+        Response::Error { message } => bail!("daemon error: {message}"),
+        other => bail!("unexpected response from daemon: {other:?}"),
+    }
+}
+
+async fn prunable_references(svc: &impl SandboxService) -> Result<Vec<String>> {
+    match svc.one_shot(Request::ListPrunableImages).await? {
+        Response::ImageList { images } => {
+            let mut references: Vec<String> =
+                images.into_iter().map(|image| image.reference).collect();
+            references.sort_unstable();
+            Ok(references)
         }
         Response::Error { message } => bail!("daemon error: {message}"),
         other => bail!("unexpected response from daemon: {other:?}"),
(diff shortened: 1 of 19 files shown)


chenhunghan:

Issue #294, opened 2026-08-25
artifact ls --kind cannot name mixin or connector, and the store cannot classify them


chenhunghan:
Raised by review on #280: `lns artifact ls --kind` accepts `image|sandbox`, while cli-spec §3.1 defines KIND as `sandbox|mixin|connector`. Two artifact kinds are unfilterable, and `image` is a CLI value the spec doesn't name.

This isn't a clap-enum tweak — the gap runs through the service:

- `lns_ipc::CachedKind` is `{Image, Sandbox}`; there is no `Mixin`.
- The service infers kind from the dependency edge (`image_store::kind_of`: no dependencies → image, dependencies → sandbox), so it cannot tell a mixin from a sandbox even if the CLI asked.
- Deeper: `pull_mixin_graph` deliberately records **no index entry** for mixins ("nothing runs it, and nothing has to reclaim it") — cached mixins never appear in `ls` at all, so a `--kind mixin` filter today would be a value that can never match.

Closing the gap needs decisions, then plumbing:

1. Should mixins get index entries (visible in `ls`, reclaimable by `rm`/`prune`), or does §3.1's `ls` intentionally cover only runnable artifacts? The spec says "Lists cached artifacts" and mixins are cached artifacts.
2. What happens to `image` as a filter value — plain OCI base images are real cached entries shown today; the spec's KIND doesn't include them.
3. `connector` arrives with the `kind: connector` work and can be added to the filter then.

Mechanically: record the document kind at ingest (the pull path parses it), add the variant(s) to `lns_ipc::CachedKind`, classify instead of inferring, and align `CachedKindFilter` — a breaking, no-shim change per the transitional rules.

chenhunghan:
Both halves now have PRs on the stack: #297 records mixin graph documents as index records, and #299 removes `kind: image` from the artifact surfaces (base images stay cache-internal for GC). This can close when both merge.

PR #297, opened 2026-08-25
feat: record pulled mixin graphs in the artifact index

chenhunghan:
Stacked on #296 (merge that first; this diff is the top commit).

Part of #294 — the first of two steps agreed there: make mixins index-visible. The second step (dropping `kind: image` from the user surface) follows as its own PR.

## Why

`docs/cli-spec.md` §3.1 closes with the decisive sentence: *"Every artifact verb here works on them exactly as it works on a sandbox."* Today a pulled mixin takes no index row — invisible to `ls`, unremovable by name, unprotected from nothing and by nothing. This PR records the pulled graph so the existing store machinery (holder refusal, dependency closure, prune) applies to mixins for free.

## What changed

**lns-service**
- `ImageRecord` now carries an explicit `kind: image | sandbox | mixin`. Kind was inferred from dependency edges, which cannot tell a leaf mixin (no deps, no layers) from a base image. Pre-format records no longer parse; the loader already skips unparseable records, so a stale cache self-heals by re-pulling.
- `pull` of a mixin records one index row per document of the pinned graph (`warm` now returns the graph — pinned identities plus pinned edges — instead of a bare count). Rows are config-only: no layers, digest from the pin.
- A pulled sandbox records dependency edges to the mixins it declares (`PulledSandbox` gains `mixins`), so `rm` of a mixin a cached sandbox layers on is refused naming the sandbox, and prune's closure keeps what a live sandbox reaches.
- A run's record now **merges** with an existing record's edges instead of clobbering them — without this, running a pulled sandbox would silently drop its mixin edges.

**lns-ipc**
- `CachedKind` gains `Mixin` (`"mixin"` on the wire and in words).

**lns-cli**
- `lns artifact ls --kind mixin` filters to cached mixins.

**docs** (guides describe what ships)
- `cli-reference.md`: `ls --kind` grammar + the pull row no longer claims a mixin takes no index entry.

## Tests (red-first per layer)

- Layer 2, lns-service: 3 new scenarios — a cached mixin lists as a mixin; removing a mixin a cached sandbox layers on is refused; a mixin a live sandbox layers on is not a prune candidate.
- Layer 2, lns-cli: `ls --kind mixin` lists only the cached mixins.
- Layer 3: the mixin-pull IPC test flipped from "records no row" to "records its graph in the index" (parent + child rows, kind `mixin`, pinned digests); `mixin_graph_records` fixture/error-path units (unpinned document, invalid reference); run-record merge units (edges kept, fresh record as-is); legacy kindless record skipped by the listing instead of failing it; `warm` returns pinned edges.

## Gates

`make lint`, `make complexity`, `make coverage` all green (lns-ipc 7, lns-service 111, lns-cli 37 files at 100%, 0 failures). 277 lns-service scenarios, 2369 units; 450 lns-cli scenarios, 912 units.


chenhunghan:

PR #301, opened 2026-08-25
fix!: land the stage-5 review repairs — arbitration, --mixin rooting, docs

chenhunghan:
The review of #283 (stage 5, the `lns artifact` / `lns sandbox` split) raised five findings. Two were already repaired by #288 (the `--format` shortcut break and the silently dropped `--mixin`); this PR lands the remaining three, at the top of the stack so no stage below needs a rebase.

> **Stacked on #299.** The base of this PR is `cli-spec-s16-drop-kind-image`. The diff shows stage 17 only. Merge #299 first.

## 1. An ambiguous run prefix could delete a cached artifact

`ask_both` treated every InspectRun reply other than `RunInspect` as "not a sandbox". The service answers an ambiguous run-id prefix with `Response::Error`, so if the operand also matched a cached artifact, `lns rm <prefix>` settled on the artifact and deleted something the user never named.

The wire now says what it means: `InspectRun` answers the new `Response::RunUnknown` for a genuine miss (nothing resolved, or resolved-then-vanished), and `run_registry::resolve` returns a typed `ResolveError` (`Unknown` / `Ambiguous`) instead of a string, so the handler classifies without parsing messages. The shortcut's probe treats only `RunUnknown` as a namespace miss; an ambiguity, a daemon failure, or an unexpected reply surfaces:

```console
$ lns rm 1a2b        # two runs share the prefix; "1a2b" is also a cached artifact
Error: daemon error: ambiguous run id prefix: 1a2b
```

Written red first: the new shortcut test reproduced the review's exact failure (`Ok(Artifact)` on an ambiguous prefix) before the fix. `lns sandbox inspect ghost` now answers `no such run: ghost` from the typed miss; the other run-addressed verbs (stop, kill, logs, attach, exec, stats) keep their `Error` wording unchanged.

## 2. A relative `--mixin` reached the service unrooted

Published-artifact inspection forwarded user-named mixins verbatim, so `lns artifact inspect <REF> --mixin ./obs` handed the service a relative path that `mixin::locate` refuses. Both the namespaced command and the shortcut now root path-shaped mixins against the invocation directory before dispatch — the same `root_named_directories` the run path already uses — via the new `InspectArgs::root_mixins`. A registry reference passes untouched.

## 3. The docs still named the pre-split spellings

`docs/policy.md` recommended `lns sandbox inspect -f`, and `docs/sandbox-spec.md` still said `lns sandbox validate` (×2) and listed `lns sandbox init, validate, inspect, push` in its cross-references. All now use the `lns artifact` spellings `docs/cli-spec.md` settled.

## Tests

- Layer 3: shortcut arbitration truth table (ambiguity surfaces, a miss settles, a failed cache listing fails aloud, an unexpected reply is refused), the typed `ResolveError` in `run_registry`, `inspect_run_request` / `inspect_resolved` answering `RunUnknown`, and `InspectArgs::root_mixins`.
- Layer 2: a new scenario pins that a relative `--mixin` reaches the service as an absolute path; the inspect fixtures now fake a miss with `RunUnknown` instead of a worded `Error`, which is exactly the distinction the review asked for.

### DIFF
--- a/crates/lns-cli/src/artifact/mod.rs
+++ b/crates/lns-cli/src/artifact/mod.rs
@@ -309,6 +309,14 @@ pub struct InspectArgs {
     pub mixins: Vec<String>,
 }
 
+impl InspectArgs {
+    /// A `--mixin` directory is the user's, so it roots where they typed it; the service reads an absolute path or nothing.
+    pub fn root_mixins(&mut self, cwd: &std::path::Path) -> Result<()> {
+        self.mixins = crate::run::target::root_named_directories(&self.mixins, cwd)?;
+        Ok(())
+    }
+}
+
 #[derive(clap::Args)]
 pub struct RmArgs {
     #[arg(value_name = "REF", help = "Cached artifact reference to remove.")]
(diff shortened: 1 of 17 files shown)


chenhunghan:

PR #303, opened 2026-08-25
feat!: keep base images out of the artifact surfaces

chenhunghan:
Closes the second half of #294: `image` is no longer a kind the artifact cache admits to.

`docs/cli-spec.md`'s `--kind` grammar is `sandbox|mixin` (with `connector` to come); the base OCI image a sandbox runs on is a cache implementation detail, not an artifact the user manages. This PR makes that boundary real.

## What changed

- **`lns-ipc`**: `CachedKind` is now `{Sandbox, Mixin}` — the wire protocol has no image kind.
- **`lns artifact ls`**: base images never list; `--kind image` is refused as an invalid value (exit 2). `--kind sandbox` / `--kind mixin` are the only filters.
- **`lns artifact rm <base>`**: answers `no such image`, exactly as for an unknown reference — a base is not directly removable, and its existence is not disclosed.
- **`lns artifact tag <base> <new>`**: refused as `no such cached sandbox`.
- **`lns artifact prune`**: still reclaims an idle base image's record and layers, but never names it in the removed report.
- **GC linkage unchanged**: base-image index records are still written by the pull/run paths, and the dependency closure still protects a live or cached sandbox's base from reclamation.
- **`docs/cli-reference.md`**: ls grammar and row updated; notes that base images are cache-internal.

The `lns inspect <ref>` by-reference surface (which labels a plain OCI image `kind: image` from its registry manifest) is deliberately untouched — that describes what a reference *is*, not what the cache holds.

## Tests (red-first)

Three new Layer 2 scenarios in `lns-service` (`image_management.feature`), watched fail red before the change:
- A base image is not listed as a cached artifact
- Removing a base image directly is answered as unknown
- Pruning reclaims an idle base image without naming it

Two new Layer 2 scenarios in `lns-cli` (`sandbox_manage.feature`): `ls --kind sandbox` filters, and `ls --kind image` is refused with exit 2.

New Layer 3 tests pin the internal arms: `pull_with` refuses an Image record, `tag_with` answers a base source as uncached, prune reclaims without reporting, and rm of a run-recorded base answers as unknown while the record survives. The removal-refusal dependency check is now pinned via a mixin required by a cached sandbox.

Gates: fmt, lint, complexity, coverage all green.


chenhunghan:

PR #316, opened 2026-08-26
feat(connectors)!: rebuild connectors as mixins the user installs

jansav:
## Heads-up for four in-flight PRs

This deletes the model that #154, #158, #176 and #190 are written against. Landing order has been agreed as this PR first, so those will need rebasing onto the new model rather than the other way round. Two notes for their authors:

- **#190** (`offer a connector on first use, whatever the policy allows`) is solving *exactly* the behaviour the new spec calls for — the offer firing on a domain the policy already permits. That work has a direct successor here, not a dead end.
- **#154** (`store credential secrets in the OS keychain`) is the one worth salvaging. The new design still needs a per-machine value store, and a keychain is strictly better than the JSON file it uses today.

## Why

Connectors were deleted wholesale (29,551 deletions) to clear the way for a new design. The old model gave `connector` its own document shape: `displayName`, `egress`, and `methods[]` carrying OAuth device-flow/PKCE endpoints.

**A connector is now a mixin the user installs on their machine.** It is offered when a run reaches the service it covers, and accepting it merges the method into the guest that is already running. Because it applies to a *booted* guest, it may carry only what such a guest can still be given — and that restriction is what makes "the mixin is applied" true rather than aspirational.

## The flow that works

```
lns connector install ./some-provider     # nothing is granted; nothing opens
lns run --name reviewer <image>            # the run reaches api.some-provider.example
                                          # → a card asks: connect some-provider?
                                          #   paste a token, name the connection, Connect
                                          # → the destination opens, the credential is
                                          #   injected on the wire, the fileset is written,
                                          #   the request proceeds
lns sandbox start reviewer                 # the same run again, still granted
```

The same decisions are available without the card: `lns connector connect` / `grant` / `disconnect` / `forget`, and `lns connector list` shows what is installed and which accounts this machine holds.

## Two late changes, both requested

**A connection, not a profile.** The verb beside it was already `connect`, and `cli-reference` had drifted to "a named connection" on its own. `--profile` is now `--connection`; `--as` stays a separate flag, because one flag meaning both "name a new one" and "pick a held one" turns a mistyped label into a second account instead of an error.

**A grant belongs to a run, not to the host working directory.** The working directory now decides one thing only: which local mixin resolves. A grant is keyed by the run, kept across every stop and start, and removed when the run is removed.

Keyed by the run **id**, never its name: `run_registry` frees a name the moment `lns sandbox rm` takes the run, so a name-keyed grant would hand consent to whatever run took the name next. `--project <PATH>` is replaced by a required `--run <RUN>`.

A grant may still be given ahead of the run, to a name no run holds — a **reservation**, taken by the first run the user *explicitly names* that way and consumed once. An automatically generated name never takes one, the generator avoids reserved names, and where a race puts both on one name the reservation is discarded. A rename takes nothing. The take is ordered before the pre-boot disclosure, so §1.5's MUST holds on the first run of a reserved name.

A consequence, accepted deliberately: a fresh auto-named `lns run` in a directory that granted yesterday is asked again. `--name` is how a grant is kept.

## How to read this

Forty-five commits, each one decision. The spec commits are normative — they are what the code implements, not a description of it, and reading them first is the fastest way in.

| Commit | What it decides |
|---|---|
| `refactor!: remove connectors and credentials` | Deletes the old model outright. No compatibility shim — the product is pre-1.0. |
| `docs(spec):` ×4 (early) | §1.4, §3.2, §4.1, §7.1 and the merge order: installed → detected → offered → connected → applied. |
| `feat(connectors): accept a connector document` | The document layer: `serves` + `methods[]`. |
| `feat(connectors): keep the installed set…` | The machine stores at the paths cli-spec §9 names. |
| `feat(connectors): a served destination is held and asked about` | The composition rule that turns an `allow` into an `ask` without reordering the table. |
| `feat(connectors): a served destination offers to connect` | The card, and applying a grant to a running guest. |
| `feat(connectors): a method may write files into a running guest` | Inline filesets, with a 1 MiB per-method ceiling. |
| `docs(spec): a connector connects, so what it holds is a connection` | The terminology, decided before the code moved. |
| `docs(spec): a run id is lowercase hex, so a name is never lowercase hex` | A spec correction: the old rule did not deliver the non-collision it promised. |
| `docs(spec): a grant belongs to a run, not to a directory` | The re-keying, the reservation, and its two rules. |
| `feat(connectors)!: a grant is keyed by the run that holds it` | The store, `--run`, and the adapter. |
| `feat(connectors): a grant may be reserved for a name no run holds` | The reservation row and the two verbs that write and clear it. |
| `feat(connectors): the run the user names takes what was reserved for it` | The claim, the generator skip, and the discard. |
| `feat(connectors): what a run granted goes when the run does` | `rm`, `prune`, and the boot sweep. |

## Decisions worth arguing with

**A hold is an `ask` rule, not a reordering.** Moving served patterns ahead of the table would invert `allow api.example` beside `deny *.example`, because the guest gate is first-match-wins. Instead each `allow` is cloned, narrowed to the destinations it shares with a `serves` pattern, marked `ask`, and inserted immediately before itself — provably a subset, so a `deny` that already shadowed it still wins.

**The wire is three-valued; the authoring type is not.** `lns_policy::Verdict` stays `{Allow, Deny}` so an `ask` can never reach a committable `lns-local-mixin.yaml`. Only the wire types carry `Ask`.

**A grant is bound to a digest.** The card grants against the bytes it disclosed, and a recorded grant whose connector has since been updated supplies nothing and offers again. Consent does not silently widen. A reservation is a grant in this respect.

**The grant key carries its kind rather than inferring it.** `run\0<id>\0<name>` and `name\0<run name>\0<name>`. A two-part key would already be unambiguous — a run name is never all lowercase hex — but that ties the on-disk format to `validate_run_name`, and if that rule ever relaxes, old reservations start being read as run ids. A row under any other prefix is inert: not reported, not deleted, answering for nobody.

**The generator skip is not the guarantee; the discard is.** The reserved set must be read before the name exists, so a concurrent `grant --run` can always land in between. A rule that only holds without a race is not a rule.

**Two connectors serving one destination resolve to neither.** Install refuses an overlap it can see, but `destination_covers` deliberately cannot decide two mid-segment wildcards, so this is reachable. The ordinary card asks instead, with one inform naming both.

**Declining is the only answer that clears a hold.** While a connector is undecided the hold rewrites its destinations to `ask` on every frame, so "always allow" cannot end the question. That is why the card keeps `Never here` even when this version can offer none of the connector's methods.

## What does not ship

**A packed fileset.** Inline filesets are offered and written; a method that packs a *directory* is not offerable, because materialising the layer into a running guest is a separate piece. The card never discloses a write it will not perform, and `grant` refuses it for the same reason.

**A connector fileset under a bind target writes the host.** `exclude` is a read-side mask, so binding a directory and excluding one file does not stop a fileset writing it. The fix needs the guest home, the resolved path and the mount table together, which only come together inside the guest — tracked as **lensapp/lens-sandbox-core#50**, an `st_dev` mount-crossing refusal at prepare time, per entry. **This is a leak until that lands.**

**Arming a document-declared credential.** `lns inspect` and the run summary disclose a sandbox's `spec.credentials` again (see below), but the runtime does not yet inject one. The render describes the document, not the boot.

**OAuth.** `auth.kind: oauth_device` parses and is simply not offerable; an unknown kind is not an error. Token paste is the only mechanism in this slice.

**`--kind connector` is still rejected by clap**, though `docs/cli-spec.md` lists it as grammar. `lns artifact validate` handles a connector completely without the flag, so nothing is weaker.

## Two defects this PR found in itself

**A declared credential was disclosed nowhere.** The first commit deleted the render sites while leaving `spec.credentials` parsed, validated, merged and attributed — so a sandbox declaring a secret path reported `lns.yaml is valid.` and then showed it in neither `lns inspect` nor the run summary. §1.5 approves the resolved sandbox; a path nothing shows is one nobody approved. Restored across five sites, each pinned by a test that dies alone under mutation. Two of those tests exist only because the first round of them passed a broken implementation: the mixin half of the wire, and a credential travelling to two domains where the renderer showed one.

**`lns inspect --mixin` was silently ignored for a connector document.** A flag the user typed, accepted and dropped, exposed by the rebase onto main. It refuses now.

## Follow-up already identified

A docs and dead-code sweep, kept out of these commits so they stay reviewable: `docs/policy.md:134`'s now-false "no credential injection" clause, `crates/lns-cli/README.md` documenting the removed flow, the "connection ledger" vs "durable ledger" naming split, and `dashboard/filter.rs`'s unpopulatable `"credential"` and `"connection"` filter categories. Also: `approval_flow/session.rs` carries both senses of "connection"; the run summary names a credential without its domains, because a contribution carries no injections; `lns artifact init` no longer scaffolds `credentials: []`; and `lns run --help` still promises a name "must not be all digits" where the code refuses all lowercase hex. A grant records no ledger event yet — `ApprovalKind` has only `Network`, and widening it is a cross-crate wire change with no spec text behind it.

## Verification

Rebased onto `f31efb9b`. CI green: `lint`, `test`, `coverage`, `audit`, and both `check-release-build` cross-builds. 518 lns-cli and 238 lns-service cucumber scenarios; every new file at 100%. Locally, `cargo test --workspace` fails only the known environmental root-uid tests, matched by name.

**No `@microvm` scenario, and deliberately not yet.** The live path needs a card answered, and there is no CLI surface for approving or declining a notification — so an end-to-end scenario would have nothing to drive the decision with. That surface is out of scope here; the scenario follows it. When it lands, the run to write is: start a run named `reviewer`, grant mid-run, watch the next request succeed with the injected header and the fileset appear at its `~/`-anchored path, then restart the run and watch the grant still apply. It is also where "a run takes what was reserved for its name" belongs, since the claim site is the socket driver that Layer 2 does not reach — pinned at Layer 3 for now.

Depends on nothing unmerged in `lens-sandbox-core`; the pin is `a0a9578`.


crates/lns-service/src/approval_flow/session.rs
```diff
@@ -180,118 +173,214 @@ impl ApprovalSession {
...
+        };
+        if let Some(port) = self.connectors.get()
+            && let Err(why) = port.decline(&offer.name)
+        {
+            self.notifier.inform(&why);
+            return DecisionOutcome::Resolved;
+        }
+        let held = self.held_for(&offer.name);
+        self.forget_offer(&offer.name);
+        self.represent(&held);
```
chenhunghan, line 254: `decline_offer` drops the connector from `self.offers` and re-presents the held cards, but never calls `send_policy_frame()`. `grant_offer` publishes via `apply_granted_egress`; this path publishes nothing, so the guest keeps enforcing the `ask` rules it was last sent.

`held_patterns()` is derived from `self.offers`, and its own doc comment says "dropping a connector from `offers` lifts its holds on the very next frame". After a decline there is no next frame.

I verified the four outcomes with tests on this branch, using a policy that **allows** the served destination:

| after the decline | frame published? | result |
|---|---|---|
| the decline itself | no | hold stays |
| `Allow once` | no | card returns on the next request |
| `Allow always` | yes | hold cleared, questions stop |
| `Deny always` | no, refused | `already decides this destination` |
| (`Connect` instead) | yes | hold cleared |

So it is not a dead end — `place_approved` returns `Stands` when the shadowing rule carries the same verdict, so allow-always does publish and does clear it. But the user is asked about a destination their own `lns.yaml` already allows, and `Allow once` — the natural answer for a one-off — leaves the hold in place so the card returns every time.

§3.2.4 says "Declining is the only answer that clears a hold." Here it is the only answer that does not.

One line fixes it: `self.send_policy_frame()` after `forget_offer`. The test needs a policy that allows the served destination — the current `offered()` fixture uses an empty policy, so every hold in the suite comes from the default `ask` and this case is never built.

crates/lns-cli/src/connector/mod.rs
```diff
@@ -194,1210 +132,1700 @@ pub const SPEC: CommandSpec = CommandSpec {
...
+    run: &str,
+    prompt: &mut impl Write,
+) -> Result<()> {
+    writeln!(
+        prompt,
+        "granting {} to {run} would give it:",
+        connector.name
+    )?;
+    labelled(prompt, "method", &method.label)?;
+    disclose_list(prompt, "opens", &method.opens)?;
```
chenhunghan, line 491: §3.2.4: "A granted method can still override a document's `deny` … The override MUST be named in the pre-boot disclosure ([§1.5]) and on the consent card that introduces it."

Neither happens today.

**On the card:** `disclose` prints `opens   <pattern>` from `opened_by(method)`, which is the method's `egress` match patterns and nothing else. It has no view of the run's own policy, so it cannot say the destination it is about to open is one the document denies.

**Before boot:** grant state is read in exactly one place — `supervisor/adapter.rs:182`, at supervisor start. Nothing in the resolve path calls `granted_supply_for`, so the resolved sandbox the CLI renders carries the document's egress only. On every later `lns sandbox start`, the pre-boot screen shows the `deny` and the run allows it.

The merge itself is correct — `merge_effective` orders `overlay → connectors → baseline`, so the override is real and the local mixin still wins. It is only the disclosure that is missing.

Mitigating: the user typed `lns connector grant` and confirmed `[y/N]`, so consent was given once and explicitly. What is absent is the reminder on each later start, which is what §1.5 is for.

crates/lns-artifact/src/connector.rs
```diff
@@ -0,0 +1,1035 @@
...
+    Ok(())
+}
+
+fn validate_serves(serves: &[String]) -> Result<()> {
+    if serves.is_empty() {
+        bail!(
+            "a connector must declare at least one serves entry: serves is what decides when it is offered"
+        );
+    }
+    for pattern in serves {
```
chenhunghan, line 221: `validate_serves` checks non-empty and whitespace-free, and nothing else. An entry whose port does not parse as a `u16` falls through `split_port`'s `.parse().ok()?` and is classified as one whole host name.

Verified on this branch: with `serves: ["db.internal:notaport"]`, `offers_for` returns `Offer::One` for the literal string `db.internal:notaport` and `Offer::None` for both `db.internal` and `db.internal:5432`. Same for `api.example:99999`.

So the install succeeds, `lns connector list` shows the connector, no card ever appears, and there is no error at any point. The user checks the run, the policy and the network before the typo.

§5's checklist asks only that `serves` be present and non-empty, so this is conforming rather than a divergence — but the behaviour is unchosen rather than decided, and it is a silent one. Refusing an entry whose port position is non-empty and unparseable would cost one line here.

crates/lns-service/src/approval_flow/protocol.rs
```diff
@@ -1,49 +1,186 @@
...
+}
+
+/// Hold every destination a served pattern shares with something the run allows, by inserting that allow narrowed to the shared destinations ahead of itself (§3.2.1).
+fn held_for_offers(egress: &Egress, serves: &[String]) -> WireEgress {
+    let mut http = Vec::new();
+    for rule in &egress.http {
+        if rule.verdict == Verdict::Allow {
+            for pattern in serves {
+                if let Some(shared) =
+                    lns_policy::matching::intersection(pattern, &rule.match_pattern)
```
chenhunghan, line 122: The hold and the offer answer the same question with different functions, and where they disagree it fails **open**.

Verified on this branch: with the run allowing `*.eu.example` and a connector serving `api.*.example`, the published table holds only the `allow` — no `ask` is inserted, because `intersection` returns `None` for two mid-segment wildcards. But `offers_for("api.eu.example", ...)` returns `Offer::One` for that same connector.

So a request to `api.eu.example` leaves with no token, no card is raised, and the service answers 401.

This is the documented mid-segment blind spot, and refusing to guess is right — `a_wildcard_does_not_shadow_one_reaching_outside_its_suffix` gives the reason. Worth noting the asymmetry though: on the offer side a miss costs an ambiguous card (`Offer::Ambiguous`, which is pinned), while on the hold side it costs the hold itself. Only the first is tested.

Probably not worth code today given how rare a mid-name wildcard on both sides is — but the composed behaviour deserves a test, since both halves are green individually and the combination is what fails.

chenhunghan (commented):
**Problem:** The old `kind: connector` model is deleted and rebuilt: a connector is a mixin the user installs on their machine, detected via `serves`, offered when a run reaches a covered destination, and merged as source 4 when the run grants a method. Grants are keyed by run id (with a name-keyed reservation for pre-run grants) and bound to the installed digest.

**Approach:** A hold is an `ask` rule cloned from each `allow` and narrowed to `intersection(serves, match)`, inserted immediately ahead of itself — provably a subset, so a `deny` that already shadowed the rule still wins and the first-match table is never reordered. The authoring `Verdict` stays two-valued so `ask` cannot reach a committable `lns-local-mixin.yaml`. Grant keys carry their kind, rows a build cannot parse are inert rather than deleted, and every payload type has a hand-written `Debug` that cannot print a token.

**Product alignment:** Squarely "policy you run into, not write" and "real secrets stay outside the workload" — the fileset carries the placeholder, the value stays on the host. The spec work is genuinely normative and the code follows it closely.

The design is sound and the matching layer is unusually well-pinned — `an_intersection_is_always_covered_by_both_patterns_it_came_from` states the safety property the whole ask layer rests on as an invariant rather than as examples.

**Verdict:** Needs fixes. Four findings, none blocking. Each was verified by writing a test against this branch and running it, not by reading alone.

### Should fix

**1. Declining a connector is the one answer that does not clear its hold** — `session.rs:252`

**2. A granted method that overrides a document `deny` is disclosed nowhere** — `lns-cli/src/connector/mod.rs:491`. §3.2.4 states this as a MUST, twice.

### Minor

**3. A `serves` entry with an unparseable port matches nothing real, silently** — `lns-artifact/src/connector.rs:221`

**4. Two mid-name wildcards get no hold, so that request leaves unauthenticated** — `protocol.rs:122`

### Two test gaps behind these

- **No test declines and then asserts what is published to the guest.** Every hold test calls `seeded_holding(policy, &explicit_list)` directly, and the `offered()` fixture uses an empty policy — so the case the hold exists for (the run's own egress *allows* the destination) is never constructed. That is why finding 1 was missed.
- **Layer 2 has no scenario for the hold at all.** `connector_install.feature` and `connector_connect.feature` cover installs and grants only; "a run reaches a served destination and is held" is pinned at Layer 3 alone, against CLAUDE.md's weighting of Layer 2 as the primary coverage-bearing layer.

crates/lns-service/src/approval_flow/session.rs
jansav: Fixed — `decline_offer` now calls `send_policy_frame()` after `forget_offer` and before `represent`, mirroring grant's publish-then-release order. Your one-line diagnosis was exactly right, and no policy recompute is needed: a decline stores no granted layer, so `effective_over` returns the same policy and only the derived `held` list shrinks.

One correction to the citation — the rule is §3.2.1 ("while a run has neither granted nor **declined** this connector, every destination `serves` covers is held") rather than §3.2.4.

You were right that the fixture was the obstacle: `offered()` uses an empty policy, so every hold in the suite coincides with the default ask and the bug is invisible. The new test uses a policy that allows the served destination and asserts the whole verdict list, so it also fails if the published frame drops the allow instead of merely lifting the ask. Watched red first — "expected a frame: Empty", no frame at all.

`declining_keeps_the_request_and_lets_the_ordinary_card_ask` had to change too: it asserted `rx.try_recv().is_err()` under the message "no decision frame", which only held because nothing else published. It now drains the channel and asserts no `RequestDecision` is among the frames. (commit 069c75cc)

crates/lns-artifact/src/connector.rs
jansav: Fixed. Both shapes confirmed, plus two more of the same class: `[::1]:notaport` (the `]:` branch of `split_destination` returns `Some` without applying `port_shaped`) and `db.internal:` (empty tail).

The refusal went into `lns-policy` as `unusable_port`, called from `validate_serves`. It deliberately does not reuse `split_destination`: folding an unparseable tail back into the host name is precisely the behaviour being guarded against, so the splitter cannot answer this question. I also left `split_destination` alone rather than refactoring a shared seam — it mirrors lens-sandbox-core's `parse_matcher`, its `]:`-without-`port_shaped` asymmetry is deliberate, and it sits on the matching path the guest agrees with.

One consequence worth flagging: `parse` is also the read path for stored connectors, so a connector installed before this change with a bad `serves` becomes an unreadable row and blocks later installs until it is uninstalled. That is the no-shims migration path, and the existing error already tells the user what to do.

Still open, same class, out of scope here: `api.example:0` parses as `0u16` and is accepted, though `parse_port`'s message already claims `(1-65535)`. Pre-existing and symmetric with egress. (commit 81b8f22b)

crates/lns-service/src/approval_flow/protocol.rs
jansav: Agreed on all counts — test only, no code. Added `two_mid_segment_wildcards_leave_a_served_destination_unheld`, which pins both halves in one place: the published table carries no ask, and the session still matches the connector to `api.eu.example`.

One correction to the described effect, and it is worse than a missing card. Because no ask is inserted, the guest allows the request under the run's own rule and never raises one — so `offers_for` is never reached and no 401 is involved. The traffic simply leaves unauthenticated with nothing on screen. The test says so, and names its own closing signal: delete it when `intersection` can express that shared set.

It was green on arrival, so I mutation-checked it — swapping the connector's `serves` to `*.eu.example`, a pair that does intersect, makes it fail with `[("*.eu.example", Ask), ("*.eu.example", Allow)]`. So the assertion discriminates held from unheld rather than merely passing. (commit 43faf090)

crates/lns-cli/src/connector/mod.rs
jansav: Half fixed, half deferred to #356 — and your "the merge itself is correct, only the disclosure is missing" was the right read.

**The GUI consent card now names the override.** `ConnectorMethodView` gained `overrides`, computed in `ApprovalSession` when the card is raised rather than when the offer is held, since the decisions file changes under a running guest.

One thing worth recording: the comparison is against the **shipped baseline only**, not the run's effective policy. `merge_effective` orders overlay → connectors → baseline, so a `deny` the developer typed in `lns-local-mixin.yaml` still wins and is not an override; comparing against the effective policy would cry wolf on it. `overrides` is three-valued — these are overturned, nothing is, or no document was in hand — so "clean" and "unchecked" cannot render alike.

**`lns connector grant` and the §1.5 pre-boot disclosure are deferred**, because both need the run's shipped baseline somewhere it does not exist: `set_shipped_policy` has one caller, inside a live `ApprovalSession`, it is never persisted, and nothing registers a run → session lookup. #356 carries both, plus a related gap they share a fix with — nothing calls `apply_granted_egress` off the IPC path, so a grant to a live run reaches the guest only on its next start. (commit 4ed94c70)

chenhunghan (approved):

PR #343, opened 2026-08-27
fix(sandbox): Land a fileset a volume mount would otherwise hide

### LINKED ISSUE #341: A fileset guestPath nested under a volume target validates, then is shadowed at boot
## Summary

A fileset whose `guestPath` falls **under** a volume `target` passes validation, then gets shadowed by the volume mount at boot. The workload sees no file, and the run reports nothing wrong.

Reproducer shape:

```yaml
spec:
  volumes:
    - name: node-home
      target: /home/node
  filesets:
    - hostPath: ~/.config/tool.md
      guestPath: /home/node/.config/tool.md
      optional: true
```

## Mechanism

Fileset files are baked into the composefs runtime layer on the host (`crates/lns-service/src/runtime_layer/mod.rs` `build()`, consumed by `crates/lns-service/src/composefs/descriptor.rs`), so they exist in the rootfs the moment the overlay mounts at `newroot` (`crates/lns-init/src/mount.rs:864`).

`mount_volumes` runs *after* that (`crates/lns-init/src/mount.rs:891`) and mounts the volume's ext4 over `/home/node`. The fileset file is still present in the lower layer — just covered.

One exception makes this look flaky rather than broken: `seed_volume_if_pristine` (`crates/lns-init/src/mount.rs:689` → `crates/lns-init/src/seed.rs:15` `prepare_pristine_volume`) copies `newroot/<target>` into the volume when the volume is pristine (empty except `lost+found`). So:

- **first** run against a brand-new named volume → the file *is* visible, copied into the volume;
- **every later** run → the volume is no longer pristine, so the stale first-run copy persists and host-side edits are never picked up;
- **bind** volumes → `mount_binds` (`crates/lns-init/src/mount.rs:611`) does no seeding at all, so the file is never visible.

## Why validation doesn't catch it

`validate` only rejects an **exact** string collision between a volume `target` and a fileset `guestPath` — both go into one `BTreeSet` of exact paths (`crates/lns-artifact/src/sandbox.rs:361-390`, fileset side at `:386-389`). `/home/node` and `/home/node/.config/tool.md` are not equal, so the document validates clean.

Behaviour coverage is exact-collision only: `crates/lns-cli/tests/behaviours/sandbox_author.feature:224`.

## Docs

`docs/sandbox-spec.md:675` says `guestPath` "MUST be unique across **all** `volumes` and `filesets`", and `:1189` describes them as sharing one namespace — neither states whether the check is exact or prefix-based. §3.1.11 (`:708`) says "a fileset is not a mount… the run writes its files into the guest's filesystem", which reads as a promise that the file always lands.

## Not the cause here, but worth ruling out first

Two other paths skip a `hostPath` fileset silently-by-design, and both emit one log line, so a run log distinguishes them from shadowing:

- `optional: true` and the host file is genuinely absent → `skipping optional hostPath … — not present on this host` (`crates/lns-service/src/artifact/fileset.rs:183-191`)
- pulled sandbox, path not granted by this machine → `skipping hostPath … this machine did not grant it` (`crates/lns-service/src/artifact/fileset.rs:150`)

Tilde expansion itself is fine (`crates/lns-service/src/artifact/fileset.rs:203-211` `resolve_host_source`).

## Expected

Either fail closed at validate time, or make the write land. Suggested split:

1. **Validation (this issue).** Refuse a fileset `guestPath` nested under a volume `target`, and a volume `target` nested under a fileset `guestPath`, with a message naming both entries. Fail-closed matches the rest of that function, and the alternative — a warning — still leaves a run whose files silently aren't there.
2. **Docs.** State that the shared-namespace rule is prefix-based, not exact.
3. **Separate feature, not part of this.** "Seed a file into a volume on every run" is a real use case (dotfiles into a persisted home) and needs its own design — seeding currently happens once, only on a pristine volume. It shouldn't arrive by way of a shadowed fileset.

## Workaround

Land the file outside the volume (e.g. under `/etc`) and copy or symlink it into place from a `pre-start` script, so it re-applies every boot instead of once at seed time.

---

Agent: Sandy

jakolehm:
## Summary

Closes #341.

A fileset whose `guestPath` fell under a volume `target` validated clean, then got shadowed by the volume mount at boot. The workload saw no file, and the run reported nothing wrong.

Fileset files are baked into the composefs runtime layer, so they exist the moment the overlay mounts at `newroot`. `mount_volumes` runs after that. A pristine named volume hid the bug further, because `seed_volume_if_pristine` copied the covered path into the volume on the first run only — so the file appeared once and never updated again.

### 1. The write lands

The host stages a write a writable volume would cover under `/.lens/fileset-deferred` instead of writing it at its final path. `lns-init` copies the staged tree onto those paths after the volumes and binds mount, then applies the chown manifest a second time — the first pass runs before the mount, when a path inside a volume does not exist yet.

The copy repeats on every boot. That is what a fileset already did: the upper layer is provisioned per run, so the host file was always what a boot got.

A symlink the last run left at a directory position is unlinked before the directory is created, so a workload cannot redirect where root writes.

### 2. Two mounts refuse instead

| Mount | Why it refuses |
|---|---|
| bind | The write would put the file in the host directory the bind shares, and leave it there after the run. A fileset never writes to the host filesystem. |
| read-only volume | It takes no write, so the file could never land. |

`lns validate` refuses a document that nests one, naming both entries. The service refuses a launch whose own `-v` flag adds one, before anything boots.

### 3. Guest-path claims compare on segments

`/home/node` and `/home/node/` are one claim, not two, so a trailing slash no longer buys a second claim on one path.

## Docs

`docs/sandbox-spec.md` §3.1.10 and §3.1.11 carry the decision, in their own commit. §3.1.10 also states that uniqueness is decided on path segments, and that a volume under a volume and a fileset under a fileset stay allowed. `docs/running-workloads.md` describes what ships.

## Follow-ups

- The stage → layer → land → chown round trip spans three crates. It belongs in the `@microvm` suite.
- `mount_run_tmpfs` runs after the landing, so a `guestPath` under `/run` is still covered by the tmpfs. Pre-existing, same bug family.
- At the flag level, a bind or read-only target that exactly equals a fileset `guestPath` is neither refused nor staged; the mount syscall refuses the boot with a worse message.

### DIFF
--- a/crates/lns-init/src/staged.rs
+++ b/crates/lns-init/src/staged.rs
@@ -1,0 +1,204 @@
+#![cfg_attr(not(target_os = "linux"), allow(dead_code))]
+
+use std::io;
+use std::path::Path;
+
+/// Guest writes a volume mount would have covered: the host stages them here instead, and `land` copies them onto the paths they name once the volumes are mounted.
+pub const STAGED_ROOT: &str = "/.lens/fileset-deferred";
+
+pub fn land(newroot: &str) -> io::Result<()> {
+    let staged = Path::new(newroot).join(&STAGED_ROOT[1..]);
+    if !staged.is_dir() {
+        return Ok(());
+    }
+    land_tree(&staged, Path::new(newroot))?;
+    consume(&staged);
+    Ok(())
+}
+
+/// Leaving the landed tree behind would show the workload a root-owned duplicate of every file, and it has nothing left to do once the copies are in place.
+fn consume(staged: &Path) {
+    if let Err(err) = std::fs::remove_dir_all(staged) {
+        eprintln!("lns-init: could not remove {}: {err}", staged.display());
+    }
+}
+
+fn land_tree(from: &Path, onto: &Path) -> io::Result<()> {
+    for entry in from.read_dir().map_err(|err| blaming(from, err))? {
+        let entry = entry.map_err(|err| blaming(from, err))?;
+        let staged = entry.path();
+        let landing = onto.join(entry.file_name());
+        let staged_is_dir = entry
+            .file_type()
+            .map_err(|err| blaming(&staged, err))?
+            .is_dir();
+        if staged_is_dir {
+            create_dir_if_absent(&landing).map_err(|err| blaming(&landing, err))?;
+            land_tree(&staged, &landing)?;
+        } else {
+            replace(&staged, &landing).map_err(|err| blaming(&landing, err))?;
+        }
+    }
+    Ok(())
+}
+
+/// A boot refusal has to name the path that refused, or the operator is left with a bare errno.
+fn blaming(path: &Path, err: io::Error) -> io::Error {
+    io::Error::new(err.kind(), format!("{}: {err}", path.display()))
+}
+
+/// A directory a mount already owns keeps the mode and owner it came with, but a symlink the last run planted there must not redirect where root writes.
+fn create_dir_if_absent(path: &Path) -> io::Result<()> {
+    if std::fs::symlink_metadata(path).is_ok_and(|found| found.file_type().is_symlink()) {
+        std::fs::remove_file(path)?;
+    }
+    match std::fs::create_dir(path) {
+        Err(err) if err.kind() == io::ErrorKind::AlreadyExists => Ok(()),
+        other => other,
+    }
+}
+
+fn replace(staged: &Path, landing: &Path) -> io::Result<()> {
+    if std::fs::symlink_metadata(landing).is_ok() {
+        std::fs::remove_file(landing)?;
+    }
+    if staged.symlink_metadata()?.file_type().is_symlink() {
+        return std::os::unix::fs::symlink(std::fs::read_link(staged)?, landing);
+    }
+    std::fs::copy(staged, landing).map(|_| ())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::os::unix::fs::PermissionsExt;
+
+    fn stage(newroot: &Path, guest_path: &str, body: &[u8], mode: u32) {
+        let staged = newroot.join(&STAGED_ROOT[1..]).join(&guest_path[1..]);
+        std::fs::create_dir_all(staged.parent().unwrap()).unwrap();
+        std::fs::write(&staged, body).unwrap();
+        std::fs::set_permissions(&staged, std::fs::Permissions::from_mode(mode)).unwrap();
+    }
+
+    #[test]
+    fn a_staged_write_lands_on_the_path_it_names_with_its_mode() {
+        let root = tempfile::tempdir().unwrap();
+        stage(root.path(), "/home/node/.config/tool.md", b"read me", 0o600);
+
+        land(root.path().to_str().unwrap()).unwrap();
+
+        let landed = root.path().join("home/node/.config/tool.md");
+        assert_eq!(std::fs::read(&landed).unwrap(), b"read me");
+        assert_eq!(
+            std::fs::metadata(&landed).unwrap().permissions().mode() & 0o777,
+            0o600
+        );
+    }
+
+    #[test]
+    fn a_staged_write_replaces_what_the_last_boot_left_in_the_volume() {
+        let root = tempfile::tempdir().unwrap();
+        let volume = root.path().join("home/node");
+        std::fs::create_dir_all(&volume).unwrap();
+        std::fs::write(volume.join("tool.md"), b"stale").unwrap();
+        stage(root.path(), "/home/node/tool.md", b"fresh", 0o644);
+
+        land(root.path().to_str().unwrap()).unwrap();
+
+        assert_eq!(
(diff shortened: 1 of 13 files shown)


crates/lns-init/src/staged.rs
```diff
@@ -0,0 +1,204 @@
...
+    Ok(())
+}
+
+/// A boot refusal has to name the path that refused, or the operator is left with a bare errno.
+fn blaming(path: &Path, err: io::Error) -> io::Error {
+    io::Error::new(err.kind(), format!("{}: {err}", path.display()))
+}
+
+/// A directory a mount already owns keeps the mode and owner it came with, but a symlink the last run planted there must not redirect where root writes.
+fn create_dir_if_absent(path: &Path) -> io::Result<()> {
```
jansav, line 76: **`create_dir_if_absent` unlinks a symlink at every level from `newroot`, not just inside the volume.**

The symlink defense is meant for what the previous run's workload planted inside the volume, but `land_tree` walks from `newroot` and applies it to every ancestor directory.

**Why:** `do_mkdir_p` traverses an image-shipped directory symlink (`/lib -> usr/lib` on Debian, `/home -> var/home` on ostree images), so the volume is already mounted at the resolved path. `land` then deletes that symlink, creates an empty real directory in its place, and copies the file into a directory nothing is mounted on. The workload loses both the file and the image's directory layout. Restrict the unlink to positions at or below the volume target.

jansav (commented):
**Problem:** A fileset whose `guestPath` fell under a volume `target` validated clean and was then hidden by the volume mount, so the workload silently never saw the file.

**Approach:** The host rewrites such specs to a staging root (`/.lens/fileset-deferred`); `lns-init` copies the staged tree onto the real paths after `mount_volumes`/`mount_binds` and re-applies the chown manifest. Binds and read-only volumes can take no such copy, so they are refused — offline by the parser, at launch by the orchestrator for `-v` mounts. Guest-path uniqueness now compares path segments.

**Product alignment:** Correct — "policy you run into, not write": a nesting nobody can satisfy is refused with both entries named, and a nesting that works just works.

**Verdict: Needs fixes.**

### Should fix

**1. Merge still keys mounts on the raw string, so a trailing slash from a mixin reopens the bug**
`crates/lns-artifact/src/merge.rs:393` (not in the diff, so no inline comment)

`fold_mounts` claims targets by `volume.target.clone()` / `fileset.guest_path.clone()`. The parser's new segment comparison only sees one document; merge output never goes back through `parse_of_kind`. §3.3.2, which this PR edits to say "compared on path segments", is not implemented.

**Why:** A sandbox with volume `target: /home/node` plus a local mixin with `guestPath: /home/node/` yields two surviving mounts instead of last-wins. `encloses("/home/node", "/home/node/")` is then false, nothing is staged, and the volume hides the fileset — the exact failure this PR closes, reachable through a mixin.

**2.** See the inline comment on `crates/lns-init/src/staged.rs`.

crates/lns-init/src/staged.rs
jakolehm: Good catch — fixed in 8e3fb392.

You are right about the mechanism. `land` now takes the run's volume targets, `land_tree` carries the guest path alongside the landing path, and the unlink fires only at or below a volume target — the only surface a previous run could have written to. Above a volume an image symlink survives and the walk follows it, so the file lands where the volume is actually mounted. `create_dir_if_absent` is back to its plain form.

Two tests hold it in tension, and I mutation-proved both: `a_directory_symlink_the_image_ships_above_the_volume_survives_and_is_followed` builds the ostree `/home -> var/home` shape and fails if the unlink runs everywhere, and `a_symlink_the_last_run_left_at_a_directory_position_does_not_redirect_the_landing` fails if it runs nowhere.

Binds are not a second gap here: `refuse_writes_a_mount_would_hide` refuses a write under a bind or a read-only volume before the run starts, so every staged entry is at or below a writable volume target by construction.

jansav (approved):
Re-reviewed at 8e3fb392.

**Finding 2 — fixed.** `land` takes the run's volume targets, `land_tree` carries the guest path alongside the landing path, and `inside_a_volume` gates the unlink to at-or-below a target. Checked the segment comparison: `/home` against target `/home/node` is false, so an image symlink above the volume survives and the walk follows it to where the mount actually is. `a_directory_symlink_the_image_ships_above_the_volume_survives_and_is_followed` pins it.

**Finding 1 — my consequence was wrong.** `refuse_what_no_sandbox_could_be` runs `parse_resolved` on the *merged* document (`crates/lns-service/src/artifact/mixin.rs:476`), so the new segment-based duplicate check does fire after merge. A mixin spelling `/home/node/` against a sandbox volume `/home/node` does not silently hide the fileset — it fails validation. What is left is smaller: `fold_mounts` still keys on the raw string, so the two spellings never collapse into one last-wins claim the way §3.3.2 now describes, and the author gets a refusal where the spec promises an override. Pre-existing in the merge keyspace, not a blocker for this PR.

**Verdict: Ship it.** ✅

PR #344, opened 2026-08-28
fix(audit): drop the hidden --json alias the spec forbids

chenhunghan:
## Problem

`docs/cli-spec.md` §4.3 says: "there is no separate `--json` switch." The code still had one. `lns audit` carried a hidden `--json` flag that aliased `--format jsonl`.

## Change

This commit deletes the flag. `--format <table|jsonl>` is the only output switch.

## Before

`--json` worked, but no help text showed it:

```
$ lns audit --json
{"class_uid":3002,"unmapped":{"lns_kind":"connection"}, ...}
```

## After

Clap rejects the argument:

```
$ lns audit --json
error: unexpected argument '--json' found

  tip: to pass '--json' as a value, use '-- --json'

Usage: lns audit [OPTIONS] [SANDBOX]

```

## Unchanged

`--format jsonl` emits the same bytes as before:

```
$ lns audit --format jsonl
{"class_uid":3002,"unmapped":{"lns_kind":"connection"}, ...}
```

`--format` still defaults to `table`.

## Tests

Layer 3, inline in `crates/lns-cli/src/audit/`:

- Added `there_is_no_separate_json_switch` — pins the clap `UnknownArgument` error. It failed red before the flag was deleted.
- Added `omitting_the_format_flag_selects_the_table` — pins the default.
- Renamed `the_json_flag_emits_one_raw_event_per_line` to `the_jsonl_format_...`; it now drives `--format jsonl`.
- Renamed `the_json_flag_labels_a_sandbox_run` and `json_emits_one_ocsf_object_per_event_in_sorted_order` the same way.
- Replaced the hidden-flag help test with `the_format_flag_is_the_advertised_output_switch`.
- Deleted the alias-equivalence test and the `--json`/`--format table` conflict test. Both described a flag that no longer exists.

Layer 1: `crates/e2e-tests/features/audit_verification.feature` now runs `lns audit --format jsonl`.

## Docs

- `docs/cli-reference.md` — removed both statements that `--json` still works.
- `crates/lns-cli/README.md` — the example now pipes `--format jsonl` into `jq`.

`docs/cli-spec.md` is not touched. The spec already states the target.

### DIFF
--- a/crates/e2e-tests/features/audit_verification.feature
+++ b/crates/e2e-tests/features/audit_verification.feature
@@ -68,10 +68,10 @@ Feature: lns audit shows one unified timeline across all sandboxes
     And the output does not contain "some-provider"
     And the output shows "some-oauth" before "api.example.test"
 
-  Scenario: --json emits raw events and suppresses the table
+  Scenario: --format jsonl emits raw events and suppresses the table
     Given a clean lns cache home
     And a connection ledger with sample events
-    When I run "lns audit --json"
+    When I run "lns audit --format jsonl"
     Then the exit code is 0
     And the output contains "some-oauth"
     And the output contains "class_uid"
--- a/crates/lns-cli/README.md
+++ b/crates/lns-cli/README.md
@@ -184,7 +184,7 @@ lns stop scraper
 
 ```bash
 lns logs claude-code --follow
-lns audit claude-code --follow --json | jq '.event'
+lns audit claude-code --follow --format jsonl | jq '.event'
 lns audit claude-code --denied        # just the blocks
 ```
 
--- a/crates/lns-cli/src/audit/mod.rs
+++ b/crates/lns-cli/src/audit/mod.rs
@@ -17,12 +17,9 @@ pub struct AuditArgs {
     #[arg(
         long,
         value_enum,
-        conflicts_with = "json",
         help = "Output format. jsonl emits one raw JSON event per line; its shape may change before v1.0."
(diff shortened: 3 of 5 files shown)


jansav (approved):
**Problem:** `docs/cli-spec.md` §4.3 forbids a separate `--json` switch, but `lns audit` carried a hidden `--json` alias for `--format jsonl`. This PR deletes it.

**Approach:** Removes the field and the `conflicts_with`, collapses `format()` to `unwrap_or(Table)`, deletes the two alias-only tests, adds a red-first test that pins clap's `UnknownArgument`, and re-points the docs and the Layer 1 feature. No compatibility shim — correct for pre-1.0.

**Product alignment:** Moves toward the spec target and deepens no divergence. Verified locally: no `--json` reference for any `lns` command remains in the tree; the audit unit tests, clippy, and fmt are green.

**Verdict:** Ship it. ✅

---

**Disclaimer — one thing to fix, here or in a follow-up:**

`crates/lns-cli/README.md:187-188` still advertises `lns audit --follow` and `lns audit --denied`. Neither flag exists: `AuditArgs` has only `sandbox`, `connector`, `kind`, and `format`, and `docs/cli-spec.md` §3.6 grants `lns audit` neither one. Line 187 is rewritten by this diff, so the `--json` half is fixed and the `--follow` half survives. A reader who copies the example gets `error: unexpected argument '--follow' found` — the same defect class this PR removes.

PR #346, opened 2026-08-28
fix: remove every user-visible Docker mention

chenhunghan:
The product has its own language. The word "docker" must not reach the user. This PR removes every Docker mention from the surfaces a user sees: CLI help, the scaffolded `lns.yaml`, and the guides.

### `lns sandbox --help`

Before:

```
  kill     Send a signal to a running sandbox (`docker kill`-style).
```

After:

```
  kill     Send a signal to a running sandbox.
```

### `lns sandbox attach --help`

Before:

```
      --detach-keys <DETACH_KEYS>  Detach chord; on match the CLI detaches and returns, leaving the run running (docker-attach style — no signal is sent). [default: ctrl-p,ctrl-q]
```

After:

```
      --detach-keys <DETACH_KEYS>  Detach chord; on match the CLI detaches and returns, leaving the run running — no signal is sent. [default: ctrl-p,ctrl-q]
```

### `lns volume --help`

Before:

```
Manage the named volumes used with `lns run -v` (`docker volume`-style).
```

After:

```
Manage the named volumes used with `lns run -v`.
```

### `lns init`

Before:

```
$ lns init
✓ created lns.yaml — your sandbox definition, every field ready to edit

  1. set spec.image (scaffolded to alpine:3.20)
$ head -5 lns.yaml
apiVersion: lns.run/v1
kind: sandbox
name: sandbox
spec:
  image: docker.io/library/alpine:3.20
```

After:

```
$ lns init
✓ created lns.yaml — your sandbox definition, every field ready to edit

  1. set spec.image (scaffolded to alpine:3.20)
$ head -5 lns.yaml
apiVersion: lns.run/v1
kind: sandbox
name: sandbox
spec:
  image: alpine:3.20
```

The scaffold text now agrees with the next-steps line, which already said `alpine:3.20`. A bare reference resolves the same way it did before, so the scaffold boots the same image.

## Tests

Red first, in the layer that owns the behaviour. All are Layer 2 (Gherkin, `crates/lns-cli/tests/behaviours/`):

| Feature file | Scenario | Pins |
| --- | --- | --- |
| `sandbox_cli.feature` | the sandbox help describes the verbs in the product's own words | `lns sandbox --help` and `lns sandbox attach --help` contain no "docker" |
| `volume_cli.feature` | the volume family lists its verbs in help | `lns volume --help` contains no "docker" |
| `sandbox_author.feature` | the scaffolded image reads in the product's own words | the scaffold writes `image: alpine:3.20` and contains no "docker" |

The three new assertions failed before the code change (493 scenarios, 3 failed). They pass after it (493 scenarios, 493 passed).

## Review follow-ups

- The scaffold's bare `spec.image` differs from the same string as a run operand (a document field passes through as written; an operand qualifies against `run.registry`, else the hub) — `docs/getting-started.md` now states that split beside the scaffold.
- The `does not contain "docker"` guards are now case-insensitive via a new `in any casing` step, so "Docker-style" fails them too.
- The claude-code example keeps its explicit `docker.io/library/...` base image and names the host in prose — the reader needs the egress destination.
- The e2e inspect scenario also asserts `does not contain "docker.io"`, so a revert of the scaffold goes red.

## Registry default

`docs/cli-spec.md` §3 says a bare reference is qualified by `--registry`, then `run.registry`, then the Lens hub. The code agrees: `DEFAULT_REGISTRY` is `hub.lns.run` (`crates/lns-cli/src/config/mod.rs`). No code path defaults a bare reference or a registry to `docker.io`. No divergence, so no change was needed here.

One related branch stays: `resolve_default_registry` does not prepend a registry the user themselves set to a Docker Hub alias, because the OCI parser already assumes that host and prepending it would break implicit `library/` namespacing. That is compatibility for a host the user typed, not a default we choose.

## Inventory

Every `grep -rni docker` hit in the repo, classified.

### Fixed — user-visible

| Location | Hit | Change |
| --- | --- | --- |
| `crates/lns-cli/src/sandbox/mod.rs:27` | `kill` about: "(`docker kill`-style)" | removed |
| `crates/lns-cli/src/sandbox/mod.rs:148` | `--detach-keys` help: "docker-attach style" | removed |
| `crates/lns-cli/src/volume/mod.rs:80` | `volume` about: "(`docker volume`-style)" | removed |
| `crates/lns-cli/src/artifact/author.rs:26` | scaffold `image: docker.io/library/alpine:3.20` | `image: alpine:3.20` |
| `docs/cli-reference.md` (5 rows) | "Docker keyed syntax", "Docker-style `--rm`", "`docker tag`-style", "docker-attach style", "`docker volume`-style" | removed |
| `docs/running-workloads.md` (12 places) | "`docker build -f`", "Docker-style" user/memory/volumes/binds, "divergence from `docker run`", "docker-style detach", "docker-tag style", "top-level docker-style verb", "docker-attach style", `docker.io/library/...` images | removed or replaced with plain description |
| `docs/getting-started.md:131` | scaffold sample image | `alpine:3.20` |

| `docs/examples/claude-code/lns.yaml` + `README.md` | `docker.io/library/debian:bookworm-slim`, "(`docker.io`)" prose | **kept, per review** — a third-party base image's registry host is the fact this network-policy example teaches, not product surface |

### Left — internal, compatibility, or not a user surface

| Location | Hit | Why it stays |
| --- | --- | --- |
| `crates/lns-policy/src/registry_auth.rs` | `docker.io`, `index.docker.io`, `registry-1.docker.io`, `registry.hub.docker.com` alias folding | Registry hosts a user types. OCI compatibility. |
| `crates/lns-cli/src/config/mod.rs` | Docker Hub alias short-circuit and its doc comments | Compatibility for a user-set registry, plus internal doc comments. |
| `crates/lns-artifact/src/registry.rs`, `memory.rs` | doc comments naming Docker/containerd behaviour | Internal doc comments. Not printed. |
| `crates/lns-service/src/image/mod.rs`, `crates/e2e-tests/tests/support/registry.rs` | `IMAGE_DOCKER_LAYER_*` media types, `Docker-Content-Digest` header | Wire constants. |
| `crates/lns-service/mise.toml`, `crates/lns-service/src/tools/mise.rs`, `crates/bump-mise/` | `docker.io/library/...` provisioner base pins | Internal pins. Not printed. |
| `crates/lns-service/src/supervisor/mod.rs:29`, `crates/lns-service/build.rs:146` | "(Docker required)" in the `LNS_NFT_BIN` error | Developer escape hatch. The message already names a repo script that no user install carries. |
| `crates/lns-cli/src/run/host_bind.rs:330`, `crates/lns-artifact/src/sandbox.rs:37` | `.docker` in the secret-shaped path list | A real host path name, not product language. |
| `crates/lns-supervisor/src/dispatcher/runtime.rs:88`, `crates/lns-service/src/run_registry.rs:95`, `crates/lns-cli/src/run/target.rs:62`, `crates/lns-cli/src/shortcut.rs` | doc comments | Internal. Not printed. |
| `crates/*/tests/**`, `crates/*/src/**` test modules, `crates/e2e-tests/**` | fixtures, feature-file descriptions, `find_docker()` oracles, `public.ecr.aws/docker/library/...` images | Test-internal. Never rendered to a user. `sandbox_cli.feature` keeps its "output does not contain docker exec/kill" guards. |
| `scripts/build-static-nft.sh`, `scripts/static-nft.Dockerfile`, `scripts/verify-static-nft.sh`, `.github/workflows/*` | build tooling | Out of scope. |
| `CHANGELOG.md` | historical entries | Generated by release-please. History is not rewritten. |

### Flagged for you — not changed

| Location | Text | Why I left it |
| --- | --- | --- |
| `crates/lns-cli/README.md:202` | "There is no Docker, no Apple Container CLI, no Lima, no other dependency." | This is a deliberate positioning claim about what the product does **not** need, not Docker as product surface. Removing it weakens the "no system dependencies" statement. Tell me if you want it reworded. |
| `crates/lns-cli/README.md:210` | "No Docker daemon, no always-on service." | Same reason. |
| `docs/sandbox-spec.md:798` | `.docker` in the secret-shaped file list | Spec text, and a real host path name. Spec edits are your call. |

No sentence in `docs/cli-spec.md` mentions Docker.


crates/lns-cli/src/artifact/author.rs
```diff
@@ -23,7 +23,7 @@ const SANDBOX_SCAFFOLD: &str = "apiVersion: lns.run/v1
 kind: sandbox
 name: sandbox
 spec:
-  image: docker.io/library/alpine:3.20
+  image: alpine:3.20
```
jansav, line 26: **The scaffold now teaches a reference the CLI resolves elsewhere.**

`spec.image` is never qualified — it reaches `oci-client`, so `alpine:3.20` still means `docker.io/library/alpine:3.20`. But `lns run alpine:3.20` on the command line qualifies to `hub.lns.run/alpine:3.20` (`crates/lns-cli/src/config/mod.rs:400`, `docs/cli-spec.md` §2.4). The first document a user ever sees now models a bare reference whose meaning flips when they type the same string as an operand.

**Why:** The user copies the scaffold's pattern into `lns run`, or into a project with `run.registry` set, and gets a different registry with no error that explains why. Either keep the scaffold explicit, or state the document-vs-operand split in `docs/getting-started.md` next to the scaffold.

crates/lns-cli/tests/behaviours/sandbox_cli.feature
```diff
@@ -15,6 +15,14 @@ Feature: managing running sandboxes from the CLI
     And the output contains "inspect"
     And the output contains "rm"
 
+  Scenario: the sandbox help describes the verbs in the product's own words
+    When I run "lns sandbox --help"
+    Then the exit code is 0
+    And the output does not contain "docker"
```
jansav: **The regression guards are case-sensitive.**

`then_output_does_not_contain` (`crates/lns-cli/tests/behaviours/steps/sandbox_cli.rs:279`) uses `str::contains`, so a help string saying "Docker-style" passes the guard. Prose capitalises the word; that is the form most likely to come back.

**Why:** The guard's stated job is "the word docker never reaches the user", and it does not catch the spelling the next author will write. Lowercase both sides in the step, or add the capitalised needle.

docs/examples/claude-code/README.md
```diff
@@ -63,7 +63,7 @@ lns push "ghcr.io/$(gh api user --jq .login | tr A-Z a-z)/claude-code:0.1.0"
 
 One artifact uploads: the sandbox definition, including its inline fileset. The
 base image is referenced as written in `spec.image`, not re-uploaded — a
-consumer pulls it from its origin (`docker.io`). `lns push`
+consumer pulls it from its origin registry. `lns push`
```
jansav: **The example lost the fact the reader needed.**

"pulls it from its origin (`docker.io`)" became "pulls it from its origin registry", and line 101 likewise. This example is about network policy: the reader has to know which host to allow, and `spec.image: debian:bookworm-slim` no longer names one anywhere on the page.

**Why:** A consumer following the example cannot work out the egress destination. Qualify `spec.image` in this example instead of stripping the host from the prose — it is a third-party base image, not product surface.

crates/e2e-tests/features/sandbox_authoring.feature
```diff
@@ -54,5 +54,5 @@ Feature: authoring a sandbox definition offline
     And I run "lns artifact inspect" in the project directory
     Then the exit code is 0
     And the output contains "Sandbox: sandbox"
-    And the output contains "docker.io/library/alpine:3.20"
+    And the output contains "alpine:3.20"
```
jansav, line 57: **This assertion no longer pins what its commit says it pins.**

`the output contains "alpine:3.20"` also passes against the old `docker.io/library/alpine:3.20`.

**Why:** Commit `77ba139` claims to pin the new scaffold value, but the assertion would stay green through a revert. Add `And the output does not contain "docker.io"`.

jansav (commented):
**Problem:** The product's own language leaks: `lns` help, the scaffolded `lns.yaml`, and the guides all name Docker. This PR removes every user-visible mention.

**Approach:** Three commits — CLI help strings and the scaffold image, then the guides, then an e2e assertion. Three Layer 2 Gherkin guards assert the help and the scaffold carry no "docker".

**Product alignment:** Correct. The vision and `docs/` own the terminology, and comparison-to-Docker framing is not it.

**Verdict:** Needs fixes. Four findings, all "should fix" — no blocker.

crates/lns-cli/src/artifact/author.rs
chenhunghan: Fixed — kept the scaffold bare (that is the shipped `spec.image` semantics: the field is passed through as written, and `docs/sandbox-spec.md` §3.1.1 accepts any OCI reference form without promising hub qualification) and stated the document-vs-operand split in `docs/getting-started.md` directly beside the scaffold: a bare `spec.image` resolves as every OCI client reads it, while a bare run operand qualifies against `run.registry`, else the hub — with the advice to write the registry host into `spec.image` when you mean a different one. (commit 971784b)

crates/lns-cli/tests/behaviours/sandbox_cli.feature
chenhunghan: Fixed — added an `in any casing` variant of the does-not-contain step (lowercases both sides) rather than changing the generic step's semantics for every other scenario, and switched the docker guards to it: `lns sandbox --help`, `lns sandbox attach --help`, `lns volume --help`, and the scaffolded-file guard in sandbox_author.feature. "Docker-style" now fails the guard. (commit 79b69c2)

docs/examples/claude-code/README.md
chenhunghan: Fixed — qualified the example's `spec.image` back to `docker.io/library/debian:bookworm-slim` and restored the host in both prose spots (line 66 and the package-visibility note), since a third-party base image's registry is the fact this network-policy example exists to teach, not product surface. (commit 1457154)

crates/e2e-tests/features/sandbox_authoring.feature
chenhunghan: Fixed — added `And the output does not contain "docker.io"`, so the scenario goes red on a revert to the old scaffold value. (commit a06e27e)

jansav (approved):

PR #347, opened 2026-08-31
feat(connectors): github reaches the npm and maven package registries

jakolehm:
The github connector asked only for `repo` and `read:org`, and routed only `api.github.com` and `github.com`. A workload could not install a private package from GitHub Packages.

This adds:

- the `read:packages` scope
- routes for `npm.pkg.github.com` and `maven.pkg.github.com`
- a `bearer_header` injection for each of those two hosts

An existing connection gets the new scope only when the user connects again. That is inherent to OAuth scoping.

`ghcr.io` is out of scope here: a private image pull goes through `lns login`, which is a separate, per-machine store.

### DIFF
--- a/crates/lns-policy/src/connectors.rs
+++ b/crates/lns-policy/src/connectors.rs
@@ -987,6 +987,35 @@ mod tests {
     }
 
     #[test]
+    fn bundled_github_reaches_the_npm_and_maven_package_registries() {
+        let gh = bundled_connectors()
+            .iter()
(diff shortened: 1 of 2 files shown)


chenhunghan (approved):

jansav:
I'd say not worth it. Connectors will be removed as part of https://github.com/lensapp/lens-sandbox/pull/316 from the binary anyway.

PR #364, opened 2026-09-01
fix: a -t=false run hands its workload pipes, not a PTY

### LINKED ISSUE #359: -t=false runs still hand the workload a PTY, so CLIs page and hang
## What happens

Start a run with \`lns run -q -t=false --yes -f <doc>\`. The in-guest supervisor logs:

\`\`\`
starting agent process (PTY) ... interactive=true
\`\`\`

The workload detects a terminal on stdout. Two observed consequences:

- \`git log\` colors its output and spawns \`/usr/bin/pager\`. The pager waits for input that never arrives. The run hangs silently. \`lns logs\` shows only pager control sequences (\`--More--\`).
- Any pager-inclined or prompt-inclined CLI can stall a headless run the same way.

## What should happen

\`-t=false\` must mean the workload cannot detect a TTY. Allocate pipes, not a PTY, when the run requests no terminal. Headless runs then behave like every other CI or pipe context. Pagers never spawn.

## Workaround

Set \`GIT_PAGER=cat\` and \`PAGER=cat\` in the run env.

Observed on 0.21.0, 2026-09-01. Diagnosis of the first hang took 10 minutes.

chenhunghan:
## What was wrong

`lns run -t=false` still gave the workload a PTY. The in-guest supervisor logged:

```
starting agent process (PTY) ... interactive=true
```

The workload detected a terminal on stdout. `git log` then colored its output and spawned `/usr/bin/pager`, which waited for input that never arrives. The run hung silently and `lns logs` showed only pager control sequences.

## Where the signal was lost

The `-t` answer already reaches the guest correctly:

1. `lns-cli` sends `tty = !detached && --tty && stdin_is_tty()` in the run request.
2. `lns-service` puts it on the wire as `ClientFrame::OpenSession { tty, .. }`.
3. `lns-session-broker` honours it: `run_tty_session` dups a PTY slave onto the supervisor's fds 0/1/2, `run_pipe_session` dups pipes.

Then the supervisor threw the answer away. It decided PTY-vs-pipes from its own `TERM`:

```rust
fn agent_wants_pty() -> bool {
    match std::env::var("TERM") { Ok(v) if v.is_empty() || v == "linux" => false, Ok(_) => true, Err(_) => false }
}
```

`lns-service` appends `TERM=xterm-256color` to every supervised run's env (last-wins, so `-e TERM=…` cannot clobber it). So `agent_wants_pty()` was always true and every run — headless included — got `openpty()`.

**No protocol change was needed.** The supervisor's own stdio *is* the run's `-t` answer.

## The fix

`crates/lns-supervisor`:

- New host-testable decision in `dispatcher/agent.rs`:

  ```rust
  pub(crate) fn agent_io(session_has_terminal: bool, devpts_mounted: bool) -> AgentIo
  ```

  PTY only when the broker gave us a terminal *and* devpts is mounted; pipes otherwise. The `TERM` sniff is gone.
- `dispatcher/runtime.rs` calls it with `std::io::stdout().is_terminal()` and `pty_available()`.
- The pipe path now relays the workload's bytes **verbatim** (`relay_verbatim`), the way the PTY drain did. Before this change the pipe path was unreachable in production and would have prefixed every line with a dim `[stdout]` / red `[stderr]` tag — that would have been a visible output regression for every headless run, so it is replaced rather than reused. Line-prefixing (`stream_output`) stays where it belongs: `pre-start` script output.

`crates/lns-service` (`workload_env.rs`): the constant that claimed to be a supervisor PTY opt-in no longer is one. Renamed `SUPERVISOR_PTY_OPT_IN_TERM` → `WORKLOAD_TERM` and corrected the comments. The value is unchanged and still exported to a supervised workload — it is a terminal *type*, not a terminal.

## Exec sessions

Not affected. `lns exec` never goes through the supervisor: the broker forks the command itself and already picks `run_pipe_session` for `tty=false`. `exec_session_env` also strips the injected `TERM` before an exec session sees it. Checked, no change needed.

## Before / after

Guest-observable, from `crates/e2e-tests/features/microvm_run.feature`:

```
lns run -q -t=false -- /bin/sh -c 'if [ -t 1 ]; then echo stdout-is-a-terminal; else echo stdout-is-a-pipe; fi'

before:  stdout-is-a-terminal
after:   stdout-is-a-pipe
```

## Tests

Red first, in the layer that owns the decision (Layer 3, `dispatcher/agent.rs` — `dispatcher/runtime.rs` is guest-only and IGNORES-listed):

- `a_run_that_declined_a_terminal_gets_pipes_even_where_devpts_is_mounted` — failed with `left: Pty, right: Pipes` against a seam stubbed with the old rule, which is exactly the reported defect.
- `a_run_that_asked_for_a_terminal_gets_a_pty`, `a_terminal_run_falls_back_to_pipes_without_devpts`.
- `a_headless_workloads_output_is_relayed_byte_for_byte` — failed with `[stdout] commit abc\r\n…` before `relay_verbatim` existed. This pins that the fix does not reformat headless output.

Layer 1: one `@microvm` scenario in `microvm_run.feature` (nightly `e2e-microvm`, not the PR gate).

## Verification

Scoped, as run in this environment (aarch64 Linux):

```
cargo test -p lns-supervisor          75 passed, 0 failed
cargo clippy -p lns-supervisor --all-targets -- -D warnings    clean
cargo clippy -p lns-supervisor -- -D clippy::cognitive_complexity    clean
cargo fmt --all
```

`lns-service` does not build in this sandbox (missing `pango`/`atk` system libraries, and no apt egress), so its rename is unverified locally — CI is the oracle for that crate. The rename is confined to `workload_env.rs`; `grep` confirms no other reference to the old name in the tree.

## Side effect worth knowing

The headless workload's stdin is now the session's stdin pipe instead of an unwritten PTY slave. With `-i=false` the broker has already closed the write end, so a read gets EOF at once. With `-i` (stdin, no tty) host stdin now actually reaches the workload — previously the non-bridged PTY path dropped it. Both match `docker run` semantics.

Closes #359

### DIFF
--- a/crates/e2e-tests/features/microvm_run.feature
+++ b/crates/e2e-tests/features/microvm_run.feature
@@ -43,3 +43,9 @@ Feature: a command runs to completion inside a real microVM
     When the user runs a microVM command "/bin/sh -c 'echo final-$((9*9)); exit 0'"
     Then the exit code is 0
     And the output contains "final-81"
+
+  Scenario: a headless run hands the workload pipes, so no pager can spawn
+    Given the Lens Sandbox service is running
+    When the user runs a microVM command "/bin/sh -c 'if [ -t 1 ]; then echo stdout-is-a-terminal; else echo stdout-is-a-pipe; fi'"
+    Then the exit code is 0
+    And the output contains "stdout-is-a-pipe"
--- a/crates/lns-service/src/workload_env.rs
+++ b/crates/lns-service/src/workload_env.rs
@@ -75,7 +75,8 @@ pub struct ToolRuntime {
     pub env: Vec<(String, String)>,
 }
 
-pub const SUPERVISOR_PTY_OPT_IN_TERM: &str = "xterm-256color";
+/// The terminal type a supervised workload is given; PTY-vs-pipes is decided by the run's own `-t` answer in the guest, never by this value.
+pub const WORKLOAD_TERM: &str = "xterm-256color";
 
 /// The vars the supervisor handshake owns; they describe the workload's own session, so replaying them into an exec would tell a piped command it has a colour terminal and hand it the workload's command line.
 fn is_session_only(entry: &str) -> bool {
@@ -83,7 +84,7 @@ fn is_session_only(entry: &str) -> bool {
     match key {
         "AGENT_COMMAND" | "WORKSPACE_PATH" => true,
         // Only the value we injected: a workload that declares its own TERM meant it.
-        "TERM" => value == SUPERVISOR_PTY_OPT_IN_TERM,
+        "TERM" => value == WORKLOAD_TERM,
         _ => key.starts_with("LENS_SANDBOX_"),
     }
 }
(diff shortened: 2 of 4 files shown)


chenhunghan:
CI note: the first `coverage` run failed on two `lns-service` prune tests, not on this change.

```
ipc::tests::handle_request_prune_runs_removes_exited_runs_and_reports_their_ids
ipc::tests::prune_leaves_a_registered_runs_dir_alone
expected RunsPruned, got Error { message: "LNS_HOME must be a non-empty absolute path, got \"relative/nowhere\"" }
```

`LNS_HOME=relative/nowhere` comes from the registry-login tests, which are `#[serial_test::serial(env)]`. The prune tests are `#[serial_test::serial(global_runs)]` — a different lock, so the two groups run at the same time and the env var leaks across. A pre-existing race, unrelated to the PTY fix (this PR touches `lns-service` only to rename one constant). Re-run of the failed jobs is green. Worth its own issue: put every test that reads `LNS_HOME` in the `env` group.

Issue #380, opened 2026-09-03
A guest gets no DHCP lease: give it its address from the host instead


chenhunghan:
## Goal

The lns service gives the guest its IP address, its default route and its DNS
server. The guest does not speak DHCP at all. The guest boot no longer depends on the host DHCP server or its shared
content-filter flow. This removes that dependency; it does not promise that
all network failures are impossible.

## Status and direction, 2026-09-09

**Current status, 2026-09-10, round 3:** Copilot completed the vsock
bootstrap change through `bd4a96ca` on lens-sandbox PR
https://github.com/lensapp/lens-sandbox/pull/395. The agent reports that VZ
now starts before network observation and configuration is sent through
`ClientFrame::ConfigureNetwork` before workload execution. Protected-plist
and cached-observation paths were removed. All required CI checks are green
and the latest review thread is marked resolved. Direct master re-review
and actual macOS acceptance remain pending; the prototype experiments do
not substitute for them. The PR is open, not merged, and the full issue's
external DHCP coexistence/default-enablement criteria remain open.

**Completion design correction, 2026-09-10:** head `252d9d6c` remains
incomplete: remembered host-network settings require a previous boot and do
not solve a fresh host. The next fix uses the existing vsock control channel
to supply static settings after VZ creates the network, before any workload
or provisioner command executes. Vsock control does not require guest IP
configuration. This supersedes the earlier kernel-command-line transport
assumption, but does not replace Apple NAT or proxy workload egress over
vsock. Remove dependence on a root-readable plist and old observations.
The master will review the implementation and validate actual lns startup
on macOS 15. External DHCP coexistence and default enablement remain open;
this correction alone is not full issue completion.

**Direct verification update, 2026-09-10:** Copilot completed fixes through
`63e20a21`, all required CI checks are green, and the three review threads
are marked resolved. Direct re-review has started. The new cold-start
fallback reads `/Library/Preferences/SystemConfiguration/com.apple.vmnet.plist`
using `std::fs::read_to_string`. Reading this file as the normal user on the
macOS 15.7.9 host returned permission denied. The fallback swallows the read
error and returns to waiting for the absent bridge. Cold-start readiness is
therefore not established; normal-user access must be addressed before
acceptance. Exit-code and spawned-reader instrumentation changes are present.
No claim of completed macOS acceptance or merge readiness is made.

**Current execution status, 2026-09-10 06:13 UTC:** at the user's request,
retried the Copilot direct-review fix round without waiting for all other
[...]

PR #388, opened 2026-09-04
fix(session): Keep DHCP leases active

chenhunghan:
## Summary

The guest stopped all DHCP traffic after its first lease. The host `bootpd` process could then stop after five idle minutes. A later guest in the same vmnet session could get no lease.

This PR keeps the first `udhcpc` process alive. The broker sends `SIGUSR1` every 120 seconds. Each signal makes `udhcpc` renew its lease with the host.

Part of #380

## Design

The broker removes `-q` from the initial `udhcpc` command. BusyBox then moves the client to the background after the first lease. The command writes its process ID to `/tmp/lns-udhcpc.pid`.

The initial command keeps `-n`. This option lets the client return an error when it gets no first lease. BusyBox clears `-n` after a successful lease and before it moves to the background. This behavior preserves the failure path for PR 2.

The broker owns a keepalive guard for the full run. The guard sends `SIGUSR1` through the bundled BusyBox every 120 seconds. The guard stops and joins its thread before the broker shuts down.

The hook stores the IP address, mask, router, and DNS values that it applied. A matching `bound` or `renew` event exits before it changes the address, route, or DHCP DNS file. A changed lease applies the address once and replaces the default route.

## Work items and commits

- `8ef1fa1` keeps `udhcpc` alive, schedules renewal, and makes the hook idempotent.
- `8ef1fa1` owns the keepalive guard in the broker for the full run.
- `a697391` records the manual macOS vmnet check in the e2e feature notes.

## Verification

The renewal test first failed because no 120-second wait or signal existed:

```text
left: []
right: [120s, 120s]
```

The hook test first failed because no lease comparison existed:

```text
lease comparison
```

The persistent-client test then failed against the timer-based prototype. It expected a `SIGUSR1` command but got a new `udhcpc` command. The implementation now sends the signal.

The changed-address test also passed a mutation check. A duplicate address application made it fail with `left: 2, right: 1`.

These scoped checks pass:

```text
cargo test -p lns-session-broker
123 passed; 0 failed

cargo clippy -p lns-session-broker -- -D warnings
Finished dev profile

cargo fmt --check
passed
```

## Manual verification

This check needs macOS vmnet. It did not run in this Linux environment.

1. Start an anchor sandbox: `lns run -d --name dhcp-anchor alpine:3.20 -- sleep 900`.
2. Wait seven minutes.
3. Start a second sandbox: `lns run --name dhcp-probe alpine:3.20 -- /.lens/guest-tools/bin/busybox ip -4 addr show dev eth0`.
4. Confirm that the second command shows an `inet` address for `eth0`.
5. Stop the anchor: `lns sandbox stop dhcp-anchor`.

## Deviations

The final implementation uses the preferred persistent client instead of the timer-based prototype. It keeps `-n` for the first lease because BusyBox clears this option after success. This choice keeps the no-lease failure path available for PR 2.

## Left out

This PR does not implement any PR 2 work item. It does not change the first-lease retry count. It does not add the fatal message, DNS warning, service exit reason, or audit event.

The macOS vmnet check remains manual. CI cannot run it.

## Screenshots

Not applicable. This change has no visual surface.

### DIFF
--- a/crates/e2e-tests/features/microvm_egress.feature
+++ b/crates/e2e-tests/features/microvm_egress.feature
@@ -12,6 +12,10 @@ Feature: the network cage blocks denied egress from a real guest
   reliable assertion needs a deterministic host-side endpoint rather than a
   real-internet host, whose reachability makes the test flaky.
 
+  DHCP keepalive needs a manual macOS vmnet check. Start one sandbox and leave
+  it running. Start a second sandbox seven minutes later. Confirm that the
+  second sandbox gets a network address and can reach an allowed destination.
+
   Scenario: a deny-all policy blocks the workload's outbound connection
     Given the LNS service is running
     And a network policy that denies all egress
--- a/crates/lns-session-broker/src/main.rs
+++ b/crates/lns-session-broker/src/main.rs
@@ -42,9 +42,13 @@ fn main() -> ExitCode {
 
 #[cfg(target_os = "linux")]
 fn run() -> Result<i32, String> {
-    if let Err(e) = network::bring_up_eth0() {
-        eprintln!("lns-session-broker: best-effort network setup failed: {e}");
-    }
+    let dhcp_keepalive = match network::bring_up_eth0() {
+        Ok(()) => Some(network::start_dhcp_keepalive()),
+        Err(e) => {
+            eprintln!("lns-session-broker: best-effort network setup failed: {e}");
+            None
+        }
+    };
     // Must run after bring_up_eth0: it consumes the DNS the udhcpc hook stashed.
     if let Err(e) = network::configure_dns() {
         eprintln!("lns-session-broker: best-effort DNS setup failed: {e}");
@@ -85,6 +89,7 @@ fn run() -> Result<i32, String> {
     drop(accept_thread);
 
     reap_exec_sessions(&exec_sessions);
+    drop(dhcp_keepalive);
 
     match primary_outcome {
         Ok(out) => Ok(out.exit_code),
--- a/crates/lns-session-broker/src/network.rs
+++ b/crates/lns-session-broker/src/network.rs
@@ -2,13 +2,18 @@
 
 mod real;
 #[cfg(target_os = "linux")]
-pub use real::{bring_up_eth0, configure_dns};
+pub use real::{bring_up_eth0, configure_dns, start_dhcp_keepalive};
+
+use std::time::Duration;
 
 pub const BUSYBOX: &str = "/.lens/guest-tools/bin/busybox";
 pub const UDHCPC_SCRIPT_PATH: &str = "/tmp/lns-udhcpc.script";
+pub const UDHCPC_PID_PATH: &str = "/tmp/lns-udhcpc.pid";
+pub const UDHCPC_RENEW_COMMAND: &str = r#"pid=$("$1" cat "$2") && "$1" kill -USR1 "$pid""#;
 
 pub const DHCP_DNS_PATH: &str = "/.lens/dhcp-dns";
 pub const RESOLV_CONF_PATH: &str = "/etc/resolv.conf";
+pub const DHCP_RENEW_INTERVAL: Duration = Duration::from_secs(120);
 
 pub const FALLBACK_DNS: &[&str] = &["1.1.1.1", "8.8.8.8"];
 
@@ -19,10 +24,15 @@ deconfig)
     "$BB" ip addr flush dev "$interface" 2>/dev/null
     ;;
 bound|renew)
+    lease="${ip}|${mask:-24}|${router}|${dns}"
+    previous_lease=$("$BB" cat /.lens/dhcp-lease 2>/dev/null || true)
+    if [ "$previous_lease" = "$lease" ]; then
+        exit 0
+    fi
     "$BB" ip addr flush dev "$interface" 2>/dev/null
     "$BB" ip addr add "${ip}/${mask:-24}" dev "$interface"
     if [ -n "$router" ]; then
-        "$BB" ip route add default via "$router" dev "$interface" 2>/dev/null || true
+        "$BB" ip route replace default via "$router" dev "$interface"
     fi
     "$BB" mkdir -p /.lens
     : > /.lens/dhcp-dns
@@ -31,6 +41,7 @@ bound|renew)
             echo "$d" >> /.lens/dhcp-dns
         done
     fi
+    printf '%s\n' "$lease" > /.lens/dhcp-lease
     ;;
 esac
 "#;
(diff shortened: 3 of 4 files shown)


chenhunghan (commented):
Review from a read-only lns reviewer run on branch `fix/dhcp-keepalive`, with lens-sandbox-core cloned beside it because the renew crosses its nftables cage. It ran nothing; CI is the evidence for the test claims. Findings ranked by severity.

### 1. High: the 120 s renew is a unicast UDP datagram to port 67, and the guest cage denies UDP with no rule

`crates/lns-session-broker/src/network.rs:74`. The first lease lands before the supervisor installs the cage; the renew does not. The cage's output chain is `policy drop`, unmarked UDP goes to the userspace relay, and `udp_egress_verdict` returns Deny with `no-udp-rule` when nothing claims the destination (lens-sandbox-core `network.rs:139`, `:177`, `proxy.rs:1865`). lns has no UDP egress table at all (`crates/lns-policy/src/lib.rs:1018`). Only UDP 53 escapes, via the DNS redirect. Scenario: `lns run ... -- sleep 900`; at t=120 s the renew to `192.168.64.1:67` is rejected, bootpd sees nothing, and `lns audit` gains one `policy-deny` line every 120 s for the run's life. The one open point: BusyBox's SIGUSR1 renew goes through `udhcp_send_kernel_packet`, which netfilter sees, unless this build's rebind path is the raw socket. A capture with `sudo tcpdump -nli bridge100 port 67 or port 68` past 120 s settles it. Fix: a cage rule `meta mark != $mark udp dport 67 accept` ahead of the queue rule in lens-sandbox-core, landed before this PR, or the capture proving the renew leaves the guest today.

### 2. High: `deconfig` flushes the address but keeps the lease cache, so a rebind to the same lease configures nothing

`network.rs:24`, `:27`, `:44`. The `bound|renew` arm exits before `ip addr add` when the cached string matches, and `deconfig` does not remove `/.lens/dhcp-lease`. The cache is on the preserved writable layer, not on boot-scoped `/tmp`. Scenario A: renew and rebind both unanswered, udhcpc enters init, `deconfig` flushes eth0, discovery wins the same address, the hook sees an identical string and exits 0: eth0 stays address-less for the rest of the run. Scenario B: `lns stop` then `lns start` boots the same layer, the host hands back the same address, and the restarted run never gets an address. Fix: `rm -f` the cache in `deconfig`, and keep the cache under `/tmp`.

### 3. Medium: every keepalive failure is discarded

`network.rs:74`, `let _ = run_check(...)`. If udhcpc dies (OOM at the default 512 MiB, say), the pidfile is gone or stale, `cat` fails, the error is dropped, and the loop waits 120 s forever. The original bug returns silently. Fix: report the first failure through the broker's stderr path and re-run udhcpc instead of only signalling.

### 4. Medium: the lease key includes DNS, so a DNS-only renew flushes and re-adds the address

`network.rs:27`, `:32`, `main.rs:53`. The key is `ip|mask|router|dns`. A changed resolver list makes the arm `ip addr flush` and re-add the same address, dropping the connected route for a moment, while `/etc/resolv.conf` stays as it was since `configure_dns` runs once. Fix: key the address work on `ip|mask|router`; write the DNS file only when it changed.

### 5. Medium: the renew signals whatever PID the file names

`network.rs:12`, `:116`. No check that the PID is still udhcpc. SIGUSR1 terminates a process that does not handle it, so a stale pidfile plus PID reuse kills an unrelated workload child every 120 s. Fix: compare `/proc/$pid/comm` with `udhcpc` before signalling; a mismatch is a dead client per finding 3.

### 6. Low: the tests pin the script's text, and the lifetime tie is unpinned

`network.rs:376`, `:390`. Two of the three new tests do substring arithmetic on the script constant and pass with finding 2 present. `keepalive_renews_after_120_seconds` is the one behavioural test. "The keepalive dies with the broker" has no test; `DhcpKeepalive` and its `Drop` are pure channel logic but sit in `network/real.rs`, which the coverage gate ignores with a reason that no longer describes the file (`scripts/coverage-floor.sh:46`). Fix: move the guard behind an injected spawner into `network.rs`, test that dropping it stops and joins the thread, update the ignore reason.

### 7. Low: the conflict with https://github.com/lensapp/lens-sandbox/pull/389 can silently disable the keepalive

`network.rs:106`. This branch drops `-q` and adds `-p`; PR 389 changes `-t 2` to `-t 7` two lines below on a base that still has `-q`. Whichever merges second conflicts, and keeping PR 389's side restores `-q`, so udhcpc exits after the first lease and finding 3 swallows the failed renews. Fix: assert in the argv test that `-q` is absent, with the reason, and state the merge order in one PR body.

### 8. Low: the manual-check note sits in the deny-egress feature and contradicts it

`crates/e2e-tests/features/microvm_egress.feature:15` asks to confirm the second sandbox "can reach an allowed destination", which the feature's own description defers as flaky, and the PR body's steps check the address instead. Fix: address only, in its own feature file or a `docs/` section.

### Questions

1. Which socket does BusyBox 1.36.1 use for the SIGUSR1 renew, and is there a capture of it leaving a caged guest?
2. Where are the two load-bearing BusyBox claims pinned: that `-n` stops applying after the first lease, and that the `bound` hook runs before udhcpc backgrounds, which keeps `configure_dns` at `main.rs:53` from reading an empty DNS file?
3. Was `/.lens` deliberate for the lease cache, given it survives `lns stop` plus `lns start`?
4. The hook script and pidfile live in world-writable `/tmp` and now run as PID 1's client every 120 s for the run's life; acceptable for a root workload, or move both under a root-only path?
5. The SAFETY note at `forker.rs:20` claims a single-threaded caller, which the new timer thread makes less true; correct it here?

chenhunghan:
Closing. The cause this PR attacks is not the cause of #380.

The host measurements are in #380. The macOS kernel drops the guest's DHCP
request before bootpd reads it, in the socket content filter, because all
guests share one flow record on the DHCP server's socket and that record
caches a filter set that goes stale. bootpd idling out is harmless: launchd
respawns it from the inherited socket in under 100 ms, and a probe that
booted into a network whose bootpd had been dead for 85 s got its OFFER 20 ms
later.

Three further problems, each settled by reading:

1. **The keepalive cannot refresh the poisoned record.** It renews on a
   different flow record, keyed by the guest's own address rather than by the
   broadcast pair that every first request uses.
2. **The renewals are refused by our own cage.** The filter chain has policy
   drop and sends every unmarked UDP datagram to the relay at
   `lens-sandbox-core/crates/lens-sandbox-core/src/network.rs:177`, where a
   policy with no UDP rule returns `Deny("no-udp-rule")` at
   `src/proxy.rs:1870`. The boot lease is unaffected, because the cage is
   installed by the supervisor long after the broker takes it, but every
   later renewal meets it. The renewal only reaches the host because SIGUSR1
   clamps the remaining lease to 30 s at busybox `networking/udhcp/dhcpc.c:1580`,
   so the refused unicast falls through to a broadcast rebind about three
   seconds later. That path also files one policy-deny audit record and one
   kernel log line every 120 s.
3. **It introduces a new outage.** With the client kept alive, a host that
   stops answering runs the hook's `deconfig` branch and flushes the guest's
   address after about twelve seconds. The `-q` behaviour left the address
   alone.

The replacement is in #380: the guest stops speaking DHCP and takes its
address, route and DNS server from the lns service.

PR #389, opened 2026-09-04
fix(network): Refuse egress runs without DHCP

chenhunghan:
## Summary

A guest could start without a DHCP lease. An egress workload then failed later with an unrelated connection error.

This change waits for a lease for at least 20 seconds. It refuses an egress run when the host gives no lease. It keeps the best-effort path for a policy that allows no egress.

The user now sees:

```text
lns-session-broker: the guest got no DHCP lease from the host network after 20 s
  the host DHCP server (bootpd) answers a session's first guest and may stop after five idle minutes
  remedy: stop every run, then start again
```

The broker sends a typed `NoDhcpLease` refusal to the service. The service exits with code 1. It records one `no_dhcp_lease` audit event. It does not report that a workload exited because the workload did not start.

Part of #380

## Work item to commit mapping

- `4f3963a` changes `udhcpc` from `-t 2 -T 3` to `-t 7 -T 3`.
- `de1b187` adds the policy branches, typed broker refusal, service exit handling, audit event, and Layer 2 scenario.
- `70ab550` documents the macOS host DHCP session and recovery.
- `9529230` covers the broker refusal display path that the coverage gate identified.

## Verification

- `cargo test -p lns-session-broker` passed. It ran 125 tests.
- `cargo test -p lns-session` passed. It ran 13 tests.
- `cargo test -p lns-ocsf` passed. It ran 31 tests.
- `cargo clippy -p lns-session -p lns-session-broker -p lns-ocsf --all-targets -- -D warnings` passed.
- `cargo fmt --all -- --check` passed.
- `git diff --check` passed.
- The local `lns-service` test and clippy commands could not start. This container does not have the GTK development packages for `gdk-3.0`, `atk`, and `pango`. CI is the service build and test oracle.

## Deviations

- The wait is 21 s, seven discovers three seconds apart, and the message and the docs say 21 s. Issue #380 says 20 s; the flags are the fact.
- The refusal is rendered on the host from the typed reason on the wire, as a run log line at level Error, instead of guest-authored stderr text. That is what keeps the three lines aligned on a raw-mode terminal and lets a detached run's log carry them too.
- The tool provisioner boots with no session; it fetches tools over the network, so a missing lease is fatal for it as well, and its marker says so.
- The DNS warning `no DHCP lease: DNS upstream falls back to 1.1.1.1` is logged by this repository when the DHCP DNS list is empty; the fallback list is ours (`FALLBACK_DNS` in `network.rs`), not lens-sandbox-core's.

## Left out

- The DHCP keepalive is https://github.com/lensapp/lens-sandbox/pull/388; this branch no longer describes it. That PR edits the same udhcpc argument list and the first statements of `run()`, so whichever merges second rebases; the argument-list test is the single place that pins `-q -n -t 7 -T 3 -s`.
- A live macOS vmnet check: this environment has no Virtualization framework. Steps below.
- `no_dhcp_lease` is in the OCSF kind list; the CLI `KindArg` and the dashboard `KINDS` filters list no exit kinds today, so it was not added there.

## Review follow-up

| Finding | Commit |
| --- | --- |
| 1 (any bring-up failure reported as a missing lease) and 10 (the WARN was ours to add) | `fix(session-broker): name the network step that failed` |
| 2 (egress marker read before the grant layer) | `fix(service): read the egress marker after the grant layer` |
| 3 (staircase on a raw terminal) and 4 (refusal skipped the host-drain wait) | `fix(run): render the broker refusal on the host` |
| 5 (conflict with #388 reverting the wait), 6 (`make lint` on macOS), 8 (decision logic in an ignored file) | `refactor(session-broker): move the network decisions to network.rs` |
| 7 (Layer 2 scenario mocked itself) | `test(service): drop the scenario that mocked itself` |
| 9 (docs described the keepalive) | `docs: describe the DHCP wait that ships` |
| 11 (wrappers with no caller) and 12 (dead 137 guard) | `refactor(service): delete the exit wrappers with no caller` |
| Question 1 (tool provisioner) | `fix(service): state the egress marker on a session-less boot` |
| coverage gate after the move | `test: pin the two lines the coverage gate found` |

Questions: 1, the provisioner refuses too, see above. 2, 21 s everywhere. 3, the policy format has no udp egress table, so `policy_allows_any_egress` has two arms and gains a third the day one lands. 4, see Left out.

### Before

Not applicable. This is a terminal and audit behavior change.

### After

The exact terminal output is in the Summary section.

## Test instructions

1. Start one sandbox on macOS. Wait more than five minutes. Start an egress-enabled second sandbox while the host does not give it a lease. Confirm that the three-line message appears. Confirm that the command exits with code 1. Confirm that the workload does not start.
2. Run `lns audit <run>`. Confirm that one event has kind `no_dhcp_lease`. Confirm that no workload exit event replaces it.
3. Start a sandbox whose effective policy denies all egress. Make DHCP fail. Confirm that the broker logs the existing best-effort setup failure and starts the workload.


chenhunghan (commented):
Review from a read-only lns reviewer run on branch `fix/no-dhcp-loud-failure` at 9529230, with lens-sandbox-core cloned beside it. It ran nothing; CI is the evidence for the test claims. Findings ranked by severity.

### 1. High: any network bring-up failure is reported as a missing DHCP lease, and the real error is discarded

`crates/lns-session-broker/src/main.rs:32`: `Err(_) if egress_allowed => Err(NO_DHCP_LEASE_MESSAGE.to_string())`. `bring_up_eth0_with` fails for five reasons before udhcpc runs: `ip link set lo up`, `ip link set eth0 up`, a busybox spawn error, a hook-script write failure, then udhcpc (`network.rs:58-86`). Scenario: guest tools missing, `run_check` says `spawn 'ip link set lo up': No such file or directory`, and the user is told bootpd idled out and to restart everything, which cannot help. Fix: claim a missing lease only when the udhcpc step failed, and carry the underlying error into the message or the console.

### 2. High: `egress_allowed` is computed before the run's granted connector egress is applied

`crates/lns-service/src/supervisor/adapter.rs:219` reads the policy from `running_policies`; `install_connectors` at `:248` then applies every granted connector through `apply_granted_egress`, which the marker never sees. Scenario: a document with a catch-all deny plus a granted connector; on `lns stop` then `lns start` the marker says no egress, the guest boots best-effort, gets no lease, and the connector's destinations fail with a connection error, the exact silent failure #380 removes. Fix: compute the marker after `install_connectors` from the session's effective policy.

### 3. Medium: the three-line message staircases on a raw-mode terminal

`main.rs:20` via `crates/lns-cli/src/service/orchestrator.rs:934`. Bare `\n` travels as `StderrBytes` and is written verbatim; `lns run` defaults to `--tty` and raw mode clears `OPOST`, so line 2 starts at column 78. The codebase already fixes this for status lines with `lf_to_crlf` at `:994`; the stderr arm has no conversion because no tty session produced stderr frames before. Fix: render the message host-side from the typed reason the wire carries, as `Response::RunLog { level: Error }`, which is CRLF-corrected and also reaches a detached run's log.

### 4. Medium: the refusal path skips the host-drain wait, while the host now suppresses its own error

`session/real.rs:205` writes two frames, closes, and `main()` proceeds to `reboot(RB_POWER_OFF)`; both normal exits wait `HOST_DRAIN_GRACE` first (`:353`, `:394`). `emit_completion` suppresses the `RunLog` error for a `BrokerRefusal`, so the guest frame is the only message. If it is still in the vsock TX queue at power-off, `lns run` exits 1 with nothing printed. Fix: wait for the host to close as the exit paths do, or render host-side per finding 3.

### 5. Medium: the conflict with https://github.com/lensapp/lens-sandbox/pull/388, and a naive resolution reverts the 20-second wait

`network.rs:70-86`, `:307-322`, `main.rs:64-72`. PR 388 removes `-q` and adds `-p` in the same array and rewrites the same first statements of `run()`. Two hazards past the textual conflict: keeping 388's side of the `-t` line restores about 6 s while the message still says 20 s, and nothing tests it because the argv test is itself in the conflict; and `bring_up_network_with` returns `Ok(())` on a failed bring-up with no egress, so 388's keepalive would start on a failed bring-up and shell out to a failing command every 120 s. Fix: land one, rebase the other with one reconciled argv, and keep the argv test as the single place that pins `-q`, `-n`, `-t 7`, `-T 3`, `-s`, `-p`.

### 6. Medium: `make lint` fails on macOS

`main.rs:20`, `:22`, `:26` are ungated and only reachable from the `cfg(target_os = "linux")` `run()`, so `dead_code` fires under `-D warnings` on the platform the product targets. CI lints on ubuntu only (`.github/workflows/ci.yml:79`). Not verified by execution. Fix: `#![cfg_attr(not(target_os = "linux"), allow(dead_code))]`, or move the helpers to `network.rs`, which already carries it (see finding 8).

### 7. Medium: the Layer 2 scenario does not exercise the service path

`crates/lns-service/tests/behaviours/steps/run_dhcp_refusal.rs:9-26` builds an error in the step body and calls two pure functions; it never goes through `handle_request`, `emit_completion` or the audit writer, and would pass if `orchestrator.rs:89` stopped passing the reason. It duplicates the Layer 3 test at `ocsf_audit.rs:304`. Fix: drive the refusal through IPC dispatch and assert the exit code plus the written audit line, or delete the feature.

### 8. Low: new decision logic sits in a file the coverage gate ignores

`main.rs:22`, `:26` against `scripts/coverage-floor.sh:41`, whose reason ("only runs inside the microVM") is now false for these two pure helpers. Fix: move them and the constant into `network.rs`, which is at 100%, with their tests. Also resolves finding 6.

### 9. Low: the docs describe the keepalive, which is not on this branch

`docs/running-workloads.md:1068-1071`. CLAUDE.md's transitional rule: guides describe what ships. The keepalive is PR 388. Fix: drop the paragraph here, add it in 388. The rest of the section checks out against #380.

### 10. Low: the WARN work item was implementable here

The Deviations section says the 1.1.1.1 fallback belongs to lens-sandbox-core. Core's `discover_upstream` reads `/etc/resolv.conf` only; the fallback is this repository's `FALLBACK_DNS` at `network.rs:13`, chosen by `select_dns` at `:107`. Fix: warn in `select_dns` or `configure_dns_with` when the DHCP list is empty; reword the deviation.

### 11. Low: two wrappers with no production caller

`audit.rs:253` `record_run_exited` and `:278` `record_run_exited_at` survive only for two tests. CLAUDE.md: delete the old spelling. Fix: delete both, pass `None` at the two test sites.

### 12. Nit: `audit.rs:298`, `reason.is_none() && exit_code == 137` cannot change the outcome; drop the guard.

### Questions

1. The tool provisioner boots with no session (`tools/provisioner/real.rs:173`), so its guest treats a DHCP failure as fatal and the user sees `broker refused the run: no_dhcp_lease` rather than the three lines. Intended?
2. `-t 7 -T 3` is about 21 s; the message says 20 s. Rounding, or say 21?
3. The format has no udp egress table (`crates/lns-policy/src/lib.rs:100-104`). Is the issue's udp question dead, or does `policy_allows_any_egress` need a third arm when one lands?
4. Should `no_dhcp_lease` join `KindArg` in the CLI audit filter and `KINDS` in the dashboard filter?

chenhunghan:
Follow-up on the review: every finding is closed on the branch, and the mapping is in the PR body under "Review follow-up". In short: 1 and 10 by naming the network step that failed and logging the DNS fallback here; 2 by reading the egress marker after the grant layer; 3 and 4 by rendering the refusal on the host from the typed reason; 5, 6 and 8 by moving the network decisions into `network.rs`, where the argument-list test pins `-t 7`; 7 by dropping the scenario that mocked itself; 9 by removing the keepalive paragraph; 11 and 12 by deleting the wrappers and the dead guard. The tool provisioner now refuses without a lease too, and the wait is stated as 21 s throughout.

chenhunghan:
Closing as written. The idea survives and is the first work item in #380,
because a typed refusal is much better than the connection error that follows
a missing lease. Four things in this diff have to change, so a smaller PR
replaces it rather than a rebase.

1. **The wider window buys nothing.** The measurements in #380 show one flow
   record poisoned for 22 minutes, with every request inside it dropped by the
   kernel. Widening `-t 2` to `-t 7` would not have saved any observed
   failure. Keep the schedule as it is.
2. **The message names the wrong cause and the wrong mechanism.** bootpd does
   not stop answering after five idle minutes. The host kernel drops the
   request before bootpd reads it. The remedy in the message is right by
   accident: letting every run exit works because it tears the shared network
   down, which frees the socket and the poisoned flow record with it.
3. **The refusal fires on every bring-up step.** The Linux backend attaches no
   network device at all, since no `--net` is emitted anywhere in
   `crates/lns-service/src/vm/cloud_hypervisor/`, so `ip link set eth0 up`
   fails on every Linux boot. As written, every Linux run whose policy allows
   egress would refuse at boot. Refuse only on the DHCP step.
4. **An empty DHCP DNS list does not prove a missing lease.** The broker
   accepts an empty list and falls back at
   `crates/lns-session-broker/src/network.rs:107-114`, and `/.lens/dhcp-dns`
   survives a restart on the preserved upper layer, so the list can be
   non-empty with no lease. The state is "no DHCP-supplied resolver".

The typed reason, the host-side renderer and the audit event are all worth
keeping, so the replacement PR should lift them from here.

Issue #393, opened 2026-09-08
PRD: spec.image may name a Containerfile or Dockerfile — lns builds it in a guest with its own executor, lns push publishes it, lns pull fetches it


chenhunghan:
## In plain terms

A sandbox document says which image the sandbox boots from. Today that has to be an image somebody already built with Docker and pushed to a registry. Anything the image lacks is installed by a script at boot, every time, in every sandbox.

This proposal lets the document point at a Dockerfile sitting next to it instead. lns builds the image itself, inside one of its own micro-VMs, under the same network rules the sandbox has. When you push the document, the built image goes with it, and the published document names the image by digest. Someone who pulls the document never builds anything.

For the mahler kit this means the four install scripts, their locks and their cache mounts become one Dockerfile, and a run starts in the time it takes to mount a layer instead of a minute of npm.

## Problem Statement

A sandbox document names the image it runs (`spec.image`) and the tools the service provisions (`spec.tools`, #173). Everything else a workload needs at boot has no place in the document, so it happens in `scripts` at `pre-start`, on every boot, in every guest.

The mahler kit (lensapp/mahler, four agents run headless against lensapp repositories) is the measure of the gap. Its definitions install at pre-start, each with a mkdir lock and a host directory bound as a cache: the Claude Code, Codex, Copilot and pi CLIs; Terraform at the version atlantis.yaml pins; yq, kustomize and kubeconform; the Rust toolchain. That is a hand-rolled image layer per tool, rebuilt in shell, with the lock and the cache standing in for what a layer would give for free. The same shape recurs for anything an agent should have before it starts, such as a skill or a plugin.

The way out today is a Dockerfile: build it with the Docker CLI, push it to a registry lns does not manage, and name the result in `spec.image` by digest. That works, and teams do it. It has three costs:

1. **Two toolchains and two logins.** The document is declarative and published with `lns push`; the image it runs on is built with `docker buildx`, pushed with `docker push` to ghcr.io, and pinned by a digest nothing in the document explains. `lns inspect` and the run summary see the digest and nothing behind it.
2. **No egress discipline at build time.** A Dockerfile built on the host fetches from anywhere. The same install done at pre-start goes through the guest's gate and produces a `Connect to <host>?` card; the kit answered one such card by hand three times before the host went into the document (`pkg-npm.githubusercontent.com`).
3. **Docker on every machine that publishes.** A laptop without Docker Desktop cannot produce the image a document needs, although it can run the document.

[...]

PR #394, opened 2026-09-08
[reading only, merges via #422] feat(sandbox): spec.image may name a Containerfile beside the document

chenhunghan:

Part of #393.

## The problem, and what changes

`spec.image` accepts only an OCI reference. #393 decides it may also name a Containerfile beside the document, which lns builds itself. The whole PRD is several PRs of work (a Containerfile parser, an in-guest instruction executor, per-instruction layer snapshots, a cache key, an image index per architecture, a `lns sandbox build` verb). **This PR lands the document layer only, and nothing else**: the grammar decides which of the two forms `spec.image` is in, the offline verbs answer for the path form, and the two verbs that would need a builder refuse it by name instead of handing a path to a registry.

Before: `image: ./image` was passed through as a registry coordinate — `lns validate` said the document was valid, and the run died later inside a pull with a reference-parse error naming nothing the author wrote. After: the path form is a recognised form with its own rules, and a form lns cannot build yet fails at the CLI, before an upload or a launch, saying so.

## Acceptance criteria → change

The PRD's testing decisions that this slice is answerable for:

| Criterion (#393) | Where |
|---|---|
| "Validation refuses `spec.image: ../elsewhere`" | `crates/lns-artifact/src/image.rs:24` `validate`, called from `crates/lns-artifact/src/sandbox.rs:313`, so every load path (validate, inspect, run preflight, push) runs it |
| "…`~/image`" | same, the home-anchored arm |
| "…a directory with neither `Containerfile` nor `Dockerfile`, each with the field name" | `crates/lns-cli/src/artifact/image_build.rs:10` `containerfile`, wired into validate at `crates/lns-cli/src/artifact/author.rs:173` |
| "A directory with both uses `Containerfile`" | `CONTAINERFILE_NAMES` (`crates/lns-artifact/src/image.rs:4`) is searched in order; observable because inspect discloses the chosen file (`crates/lns-cli/src/artifact/author.rs:337`) |
| "A file of any name is the Containerfile and its parent directory is the context" | `image_build.rs:24-27` — an existing non-directory path is the Containerfile as written |
| PRD field table: "A relative path beside the document, no `..`, not home-anchored" | `image.rs:24-49` |

Not attempted here, and so not claimed: the build guest, the instruction executor and its subset gate, the layer store import, publishing the image and rewriting `spec.image` to a digest with `imageSource`, the cache key, the architecture index, `lns sandbox build`, and the host-Docker switch.

## Implementation decisions

**One predicate decides the form, because validation, the run and the push must not disagree about it** — the same shape `names_a_local_path` already gives mixins, plus `~`:

```rust
pub fn source(image: &str) -> ImageSource<'_> {
    let leading = image.trim_start().chars().next();
    if matches!(leading, Some('.' | '/' | '~')) {
```

That makes `~/image` and `/srv/image` *path* forms that are then refused with a reason, rather than references that fail obscurely at pull time — which is what the PRD's acceptance list asks for.

**The path rules live in the grammar crate, so a document is refused wherever it is loaded.** `crates/lns-artifact/src/sandbox.rs:313`:

```rust
    crate::image::validate(&doc.spec.image)?;
```

**Resolving *which* file needs the filesystem, so it sits in the CLI behind the existing `Fs` seam** (the same one path filesets are checked through), never in the parser (`image_build.rs:13-27`):

```rust
    if fs.is_dir(&named) {
        return CONTAINERFILE_NAMES
            .iter()
            .find(|name| fs.exists(&named.join(name)))
```

**Run and push refuse rather than pretend.** `crates/lns-cli/src/run/target.rs:49` and `distribute.rs:333` / `:385` (both `push` and `push --dry-run`, beside `refuse_unpushable_tools`, so nothing is uploaded first):

```rust
            "spec.image {path:?} names a Containerfile, and building one is not yet supported; publish the image yourself and name it by reference"
```

A push that published a path would publish a document no consumer could run, which §6 forbids; a run that passed one to the service would fail inside a pull.

## Specification

`docs/sandbox-spec.md` is the normative target and the code does not reach it (AGENTS.md, "Transitional mode"), so the spec change is its own commit (e66a2f1f) and states the whole target — both forms of `image`, the build under the document's own `egress`, the push-time rewrite to a digest plus `imageSource`, and the field's appearance in §3.1, §5 and §6. The divergence this PR leaves is deliberate and named above. No user guide was touched: the path form does not ship yet, and a guide answers "how do I use this today".

## Tests (red first)

- Layer 3, `crates/lns-artifact/src/image.rs:52+` — 9 tests: classification of both forms, each refusal (escape, home anchor, absolute, control character), the accepted spellings, and that a reference is held to no path rule. Written first; watched fail (`cannot find function 'source' in this scope`, 11 errors) before the implementation existed.
- Layer 3, `crates/lns-artifact/src/sandbox.rs:1441+` — the refusal reaches the whole document at `parse`, and a path beside the document parses.
- Layer 2, `crates/lns-cli/tests/behaviours/image_containerfile.feature` — 10 scenarios over the real CLI verbs with the in-memory `Fs`: validate accepts a directory holding a Containerfile; refuses one holding neither, `../elsewhere`, `~/image`, and a path naming nothing; inspect discloses `built from ./image/Dockerfile` and prefers `Containerfile` when both exist; a named file of any extension is accepted; run refuses; push refuses and uploads nothing. Watched fail: 5 scenarios red before the CLI wiring, all green after.

| Command | Result |
|---|---|
| `cargo test -p lns-artifact` | 383 passed |
| `make test-crates CRATES="lns-cli lns-artifact"` | pass — cucumber 545 scenarios / 2655 steps, all green |
| `cargo fmt --all -- --check` | clean (after `cargo fmt`) |
| `cargo clippy -p lns-artifact -p lns-cli --all-targets -- -D warnings` | clean |
| `cargo clippy -p <each> -- -D clippy::cognitive_complexity` (target/complexity) | clean |
| `make coverage COVERAGE_CRATES="lns-artifact lns-cli"` | new files at 100%: `lns-artifact/src/image.rs` 46/46, `lns-cli/src/artifact/image_build.rs` 24/24. `crates/lns-cli/` 43 at 100%, 0 failures |

**Limitations of the local run, stated plainly:**

- `make lint` as a whole fails on this machine at `scripts/affected-crates.test.sh` (4 cases expect a crate list, get `__FULL__`). It fails identically on unmodified `main` here (verified by stashing), so it is environmental, not this change. The Rust half of `lint` (fmt + clippy) was run directly and is clean.
- `cargo clippy --workspace` cannot complete here: `lns-service` depends on `gdk-sys`/`pango-sys` and this container has no GTK development libraries. So `lns-service` was **not** compiled, tested, or coverage-measured locally.
- The narrowed coverage run reports two pre-existing FAILs in `lns-artifact` (`connector.rs` 477/478, `tools/mod.rs` 203/212). Their uncovered lines are `Method::label`, `SafeVersion` and `is_safe_bin_path` — `pub` API exercised by `lns-service`'s suite, which the narrowed run does not execute. Neither file is touched by this PR.
- CI is the oracle for all three, and it has now answered: **every required check is green** on `d588e8f0` — `lint` 2m09s (so the `affected-crates.test.sh` failure really is local-only), `test` 3m23s, `coverage` 3m13s and `coverage-as-root` 4m05s (full workspace, so `lns-service` compiles, its suite passes, and the two `lns-artifact` files above are at 100% there), plus `audit`, `changes`, `labels`, `check`.

No fixture in the tree carried a path-shaped `spec.image` before this change (`grep` over `crates/`, `docs/`), so nothing that used to parse stops parsing.

## Deviations and omissions

- **`lns-service` is unchanged.** It never receives a path form: push refuses to publish one, and the local run path refuses before the launch request. Defence-in-depth there is left to the slice that adds the builder.
- **`spec.imageSource` is specified but not implemented** — it is written by the push-time build, which does not exist yet. Adding a field no producer writes would be speculative.
- **`lns sandbox build` is not added**, and `lns run` does not build on first run. Both need the executor.

### DIFF
--- a/crates/lns-artifact/src/image.rs
+++ b/crates/lns-artifact/src/image.rs
@@ -1,0 +1,133 @@
+use anyhow::{Result, bail};
+
+/// The names a build context may hold, in the order a directory is searched — `Containerfile` first, as Podman does.
+pub const CONTAINERFILE_NAMES: [&str; 2] = ["Containerfile", "Dockerfile"];
+
+/// What `spec.image` names: an image to pull, or a Containerfile beside the document for lns to build (§3.1.1).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ImageSource<'a> {
+    Reference(&'a str),
+    Containerfile(&'a str),
+}
+
+/// A value beginning `.`, `/` or `~` is a path — no OCI reference starts that way — so one predicate decides the form for validation, the run, and the push.
+pub fn source(image: &str) -> ImageSource<'_> {
+    let leading = image.trim_start().chars().next();
+    if matches!(leading, Some('.' | '/' | '~')) {
+        ImageSource::Containerfile(image)
+    } else {
+        ImageSource::Reference(image)
+    }
+}
+
+/// The path rules a Containerfile-form `spec.image` is held to: the artifact ships what it names, so it names something beside the document (§3.1.1).
+pub fn validate(image: &str) -> Result<()> {
+    let ImageSource::Containerfile(path) = source(image) else {
+        return Ok(());
+    };
+    if path.starts_with('~') {
+        bail!(
+            "spec.image {path:?} is built from a Containerfile beside the document, so it cannot be home-anchored; publish the image yourself and name it by reference instead"
+        );
+    }
+    if path.starts_with('/') {
+        bail!(
+            "spec.image {path:?} is built from a Containerfile beside the document, so it must be relative, such as ./image"
+        );
+    }
+    if path.split('/').any(|segment| segment == "..") {
+        bail!(
+            "spec.image {path:?} leaves the document's directory, and the artifact ships what it names; move the Containerfile into the project"
+        );
+    }
+    if path.chars().any(char::is_control) {
+        bail!("spec.image {path:?} must not contain control characters");
+    }
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn a_registry_coordinate_is_an_image_to_pull() {
+        for reference in [
+            "alpine:3.20",
+            "ghcr.io/acme/base@sha256:abc",
+            "docker.io/library/node:24-bookworm",
+        ] {
+            assert_eq!(
+                source(reference),
+                ImageSource::Reference(reference),
+                "{reference} is a reference, so nothing about it is built"
+            );
+        }
+    }
+
+    #[test]
+    fn a_path_shaped_value_names_a_containerfile_to_build() {
+        for path in [
+            "./image",
+            ".",
+            "./build/api",
+            "../elsewhere",
+            "~/image",
+            "/srv/image",
+        ] {
+            assert_eq!(
+                source(path),
+                ImageSource::Containerfile(path),
+                "{path} is path-shaped, so it names a build rather than a registry coordinate"
+            );
+        }
+    }
+
+    #[test]
+    fn a_containerfile_path_that_leaves_the_document_is_refused_by_the_field_it_was_written_in() {
+        for path in ["../elsewhere", "./build/../../elsewhere"] {
+            let refusal = format!("{:#}", validate(path).unwrap_err());
+            assert!(
+                refusal.contains("spec.image") && refusal.contains(path),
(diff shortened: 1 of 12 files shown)


chenhunghan:

PR #395, opened 2026-09-09
feat(service): reserve a guest address on the host instead of hoping DHCP answers

chenhunghan:
Part of lens-sandbox issue lensapp/lens-sandbox#380

## Before / after

**Before.** Every guest booted by asking the shared macOS network for a lease. When the reply never came (the flow the issue reports), `udhcpc` failed, the broker logged a line into the guest console nobody reads, and the workload started anyway on an unaddressed interface — the run looked alive and could reach nothing. Nothing on the host knew which addresses existed.

**After, default path (unchanged addressing).** A guest still asks for a lease, but a failure is now a *typed refusal*: the workload never starts, the CLI prints the cause and a remedy, and the audit log records one named failure event instead of a fake workload exit.

```
the guest got no address from the host DHCP server
  remedy: let every run exit, which tears the shared network down, then start again
```

**After, opt-in static path (`LNS_STATIC_GUEST_NET=1` in the service's environment).** The guest boots with no address and no DHCP client — its kernel command line carries a marker and never an address — and waits. Its own boot is what creates `bridge100`, so the host reads that live interface *afterwards*, which is the first moment anything readable describes the shared network. It then excludes everything it can observe to be taken, reserves three candidates, and sends them with the prefix, gateway and resolver to the guest as a typed `ConfigureNetwork` frame on a dedicated vsock connection — before the run's workload, and before any provisioner command, is started. The guest ARP-probes a candidate, announces it, confirms the gateway, and answers with the address it took. Only then is the reservation narrowed to that one address and the workload opened. The CLI prints, in the run's own log:

```
Address  guest assigned 192.168.64.254
```

This works on a cold host, first boot: nothing is remembered between runs, no prior DHCP boot is needed, and no file of Apple's is read or written.

If every candidate is taken, if the gateway is silent, if the host cannot address the guest at all — no bridge, no address left, unreadable lease or neighbour state — or if the guest never answers, answers with something else, or answers with an address the host does not hold for it, the run refuses by name **before the workload starts** and exits 125, with the cause and a remedy instead of a call stack, and the reservation goes back:

```
the host shared network could not be discovered: the shared network is not active (ifconfig bridge100 exited with 1)
  remedy: let every run exit, then start again, or unset LNS_STATIC_GUEST_NET to boot on the host DHCP server
```

**There is no DHCP fallback in the static path.**

## Review round 1

1. `b6a2a3c7` adds a VM-lifecycle conflict monitor. `conflict_monitor_checks_throughout_one_guests_lifecycle_and_routes_its_warning` pins periodic scheduling, run routing, deduplication, and stop. Static addressing remains off by default.
2. `cff0e354` returns 125 for typed broker refusals. `a_dhcp_refusal_is_rendered_by_the_host_and_not_as_a_workload_exit` pins the CLI frame. Existing nonzero workload tests pin genuine workload codes. The same typed reason continues into the audit event.
3. `1b9c32e8` moves the lease and monitor into `boot_with_owner`, which owns them in the blocking VMM task until confirmed teardown. `vm_owner_lives_until_delayed_vmm_teardown_is_confirmed` and `vm_owner_is_released_after_a_failed_launch` pin delayed stop and failed launch. Workload and provisioner boot paths use this owner. `58aca738` fixes the test scheduler handoff.
4. `6b9e213d` replaces plist/default guessing in allocation with an observation of active `bridge100`. It retries while the VZ network appears. Tests pin a nondefault gateway, missing and malformed observations, and the rejection of ambiguous multi-subnet plist data.
5. `f49ddc7e` makes the high-to-low usable range lazy. `a_broad_subnet_yields_its_first_candidates_without_materialising_the_range` pins bounded consumption of a `/1`.
6. `07047bf3` adds a command seam for ARP observation. Spawn, nonzero exit, UTF-8/parse errors now refuse allocation. `neighbor_observation_propagates_spawn_exit_and_parse_failures` pins these cases outside real command glue.
7. `74b8004d` removes reservation persistence and startup reconciliation. The service process owns all VMM tasks, so service exit tears down every possible holder. `a_new_service_process_starts_with_no_reservations` pins this lifecycle. There is no file to truncate and no discarded persistence error.
8. `2ae8990b` adds the guest `NetworkApplied` report for workload and sessionless provisioner sessions. `assigned_address_report_is_validated_before_normal_session_frames` rejects malformed and unoffered addresses. `7726bcc2` confirms the address in the allocator, releases spare candidates, routes the host-visible address line, and limits conflict checks to the selected address. `a_confirmed_guest_address_releases_spares_and_rejects_unoffered_reports` pins narrowing and reuse.
9. **CI was red on the coverage gate**, not on a review point: `vm/guest_addr.rs` (10 lines) and `vm/session_client.rs` (1 line) had dropped below 100%. The uncovered lines were the ones review round 1 added — the `AllocError::Neighbors` refusal message from point 6, the `SelectError` messages and the per-run `ConflictSource` view from points 8 and 1, and the confirmation call itself. `59abe201` gives `AddressSelection` the allocator it confirms against instead of reaching the process-wide one, which is what made the confirmation testable at all (`a_reported_address_narrows_the_host_reservation_and_an_unheld_one_fails_the_session`); `35674d1f` pins the three refusal/report messages (`a_neighbor_table_that_cannot_be_observed_stops_allocation_instead_of_guessing`, `a_refused_confirmation_says_whether_the_address_or_the_reservation_was_wrong`, `a_runs_monitor_is_told_about_its_own_conflict_and_no_one_elses`). `03869b19` rewrites the `guest_addr/real.rs` ignore reason, which still named the `FileStore` and startup reconcile that point 7 deleted. No production behaviour changes: the seam is the same allocator, reached by a field instead of a global.

10. **P1 — static cold start.** `60fc7d8b`. Discovery required a live `bridge100`, which does not exist until a shared-network guest boots — so the first guest, whose own boot creates it, waited 20 × 250 ms and refused. Red first: `the_first_static_boot_reads_the_declared_network_the_bridge_has_not_been_created_from_yet` and `a_workload_and_a_provisioner_both_boot_first_on_a_host_with_no_bridge_yet`. `host_network_for_boot` now prefers the live bridge and falls back to `/Library/Preferences/SystemConfiguration/com.apple.vmnet.plist` — reading `Shared_Net_Address` as the gateway and `Shared_Net_Mask` as the mask, not inferring either. A host that neither runs nor declares a shared network is still refused by name (`a_host_that_neither_runs_nor_declares_a_shared_network_says_so`); the retry loop remains as the last resort and its error names both sources. `reserve` became the testable `reserve_on`, leaving only the two constants in `real.rs`.
11. **P2 — host network failures are pre-workload failures.** `a3134566`, test tightened in `63e20a21`. Red first: `a_host_that_cannot_address_the_guest_fails_before_the_workload_not_as_one`. `broker_exit_reason` became `refusal_of`, returning a `Refusal { reason, summary }` for any pre-workload refusal — `BrokerRefusal` as before, and now `ReserveError::{Network, Alloc}`, which `downcast_ref` finds through the orchestrator's `.context(...)`. `emit_completion` returns 125 for any refusal and `failure_message` renders `ReserveError::explain()` (cause plus remedy) rather than the error chain. Each cause carries its own stable audit name — `host_network_undiscoverable`, `host_addresses_exhausted`, `host_address_state_unreadable` — pinned by `every_reason_a_host_cannot_address_a_guest_has_its_own_audit_name` (`73f280d9`); `audit::record_run_exited_with_reason*` and `ocsf_audit::refusal_event` (renamed from `broker_exit_event`) take them.
12. **P3 — the address report reaches the run's log.** `dc626d3b`. `tokio::spawn` does not carry the run span, so the reader task's `log::info!("Address", …)` had no `RunFrameTx` for `FrameForwardLayer` to find and the line reached only the developer trace stream. Red first: `the_spawned_reader_reports_the_assigned_address_into_its_own_runs_log`, on a new `log::testing::capture_run_frames_spawned` harness that spawns inside the span and polls outside it — the obvious harness (entering the span around `block_on`) is green on arrival and does not model production, which is the trap this one avoids. The spawn moved behind `session_client::spawn_server_frame_reader`, instrumented with `tracing::Span::current()`; `real.rs` calls the seam. `docs/running-workloads.md` was reconciled in the same commit: discovery order, the exit-125 paragraph, the address line, and the honest statement of what the reservation lock does and does not cover.

## Review round 2

1. **P1 — the cold-start declaration is not readable by the service user.** `252d9d6c`. Correct on a real host: `/Library/Preferences/SystemConfiguration/com.apple.vmnet.plist` is `root:wheel` mode `0640`, so `RealHostFiles::read` returns `Permission denied`, round 1's fallback produced `None`, and the first static guest after a shared-network teardown went back to waiting for a `bridge100` that nothing was creating.

   Red first, and each failed for its own reason before the fix: `a_declaration_this_user_may_not_read_falls_back_to_the_network_this_host_was_seen_running` (host layer) and `a_first_static_boot_is_addressed_when_the_declaration_is_root_only` (`reserve_on`, the path both boot paths take). Both were mutation-proved after: deleting the recall fallback turns exactly those two red and nothing else.

   What changed. The declaration is no longer the last source, and no longer the only cold-start one:

   - `declared_host_network` returns `io::Result<Option<HostNetwork>>`. `Err` (unreadable) and `Ok(None)` (readable, declares nothing) are different answers and are reported as different sentences — `… is not readable by this user (permission denied)`, `… does not exist`, `… declares no shared network`. Pinned by `a_declaration_this_user_may_not_read_is_named_as_that_and_not_as_an_absent_one` and `a_host_that_declares_nothing_is_told_apart_from_one_whose_declaration_is_unreadable`. The failure is named, not hidden.
   - A third source: what this host was last **observed** running. `NetworkMemory` (`recall`/`remember`) keeps the live bridge's own address and mask in `~/.lns/host-network`; `host_network_for_boot` asks live bridge → declaration → memory → the retry loop, and remembers every live observation on the way (`a_live_observation_is_remembered_so_the_next_cold_start_is_not_a_guess`, `a_bridge_that_appears_while_the_host_waits_is_remembered_too`). A record that is not one whole subnet with its gateway inside it is refused rather than booted onto (`a_remembered_network_survives_a_round_trip_and_a_damaged_record_is_no_answer`); `RealNetworkMemory` is round-tripped over a tempdir `LNS_HOME` in `real.rs`.
   - **Every** boot teaches the host, not just static ones. A guest with a shared-network interface is what creates `bridge100`, so `boot_with_owner` calls `host_net::real::learn_shared_network()`, which on macOS observes and records in the background. One ordinary DHCP run is therefore enough to teach a host that has never completed a static boot — which is the state the reviewed host is in. `learn_host_network` is pinned by `any_boot_teaches_the_host_what_its_shared_network_is` and `a_boot_that_brings_no_shared_network_up_teaches_nothing` (no observation never erases the last one).

   Not done, deliberately: no permission change to an Apple-owned file, no privileged helper, no `sudo`, no guessed subnet, and no default. A remembered network is an observation of this host, and the guest still proves it before it commits — ARP probe, gratuitous announce, gateway check — so a stale record refuses the run by name instead of addressing a guest wrongly.

   Honest remaining gap: a host that has **never** had a shared network up under this user and whose declaration is unreadable still cannot be discovered, and the run refuses naming all three sources. There is nothing readable left to read on such a host; one ordinary run puts it right permanently. `docs/running-workloads.md` states the discovery order, the permissions reality, and this gap.

   `reserve_on` grew past clippy's argument limit, so the three ports travel together as `host_net::HostView { live, files, memory }` — no behaviour change.

## Review round 3

1. **P1 — complete the cold start instead of requiring a prior DHCP boot.** `349ba189`. Correct, and the round-2 answer was the wrong shape: `NetworkMemory` left first use broken when `bridge100` was absent, the plist unreadable and nothing cached, and a cached record could be reused after the host's subnet changed. The implementation decision recorded on the thread — use the existing guest control channel — is what landed.

   **The ordering is inverted.** The host no longer needs to know the network before the VM starts. It boots the guest (the MAC is derived from the run id, so it is known without a reservation), waits for the connector, observes the live `bridge100` that this guest's own attachment created, reserves on it, and then sends the plan over vsock. Nothing is guessed, nothing is remembered, and no file of Apple's is read.

   Red first, bottom-up, each failing before its production code existed:

   - `lns-session`: `the_plan_reaches_the_guest_over_the_session_channel_after_it_has_booted` (the new `ClientFrame::ConfigureNetwork(GuestNet)` round trip) and the `GuestNet::validate` tests (`a_plan_the_host_could_have_reserved_is_accepted`, `a_plan_with_no_candidate_leaves_the_guest_nothing_to_claim`, `a_prefix_that_leaves_no_usable_address_is_refused`, `a_candidate_outside_the_gateways_subnet_is_not_this_networks_address`, `each_refusal_says_what_was_wrong_with_the_plan`). `GuestNet` is now serde-derived; the ad-hoc `to_cmdline_value`/`parse` codec and both its error variants are deleted with their tests. `GUEST_NET_ENV` became `GUEST_NET_BOOTSTRAP_ENV`, which carries only `1`.
   - `lns-session-broker`: new `bootstrap.rs`, pinned by `the_plan_the_host_sends_is_applied_and_the_address_reported_back`, `nothing_but_a_plan_opens_this_channel`, `a_host_that_hangs_up_before_it_sends_a_plan_leaves_a_refusal_not_a_wait`, `a_plan_the_guest_cannot_use_is_refused_before_the_interface_is_touched`, `a_guest_that_cannot_take_any_offered_address_refuses_by_name` and `each_outcome_gets_one_console_line`. `main.rs` now accepts the bootstrap connection *first* and only accepts the primary session after a successful apply; a refusal exits before any workload can be opened. `NetworkResult::AwaitingBootstrap` replaces the old pre-listen apply.
   - `lns-service`: new `vm/net_bootstrap.rs` (the host half of the same exchange) with `the_guest_is_sent_the_plan_and_reports_the_address_it_took`, `a_guest_that_refuses_gives_the_host_its_typed_reason`, `an_address_the_host_never_offered_is_not_taken_as_an_answer`, `an_address_that_is_not_an_address_is_named_as_malformed`, `a_guest_that_hangs_up_before_answering_never_starts_a_workload`, `a_guest_that_never_answers_gives_up_instead_of_holding_the_run_open`, `a_frame_that_is_not_an_answer_to_the_plan_is_refused_by_name`, `a_channel_that_cannot_be_written_to_fails_before_anything_waits`.
   - `vm/guest_addr.rs`: `address_guest` is the whole sequence between a started VMM and a started workload, pinned by `the_guest_is_told_the_plan_and_the_reservation_narrows_to_the_address_it_took`, `a_bridge_the_vmm_is_still_creating_is_waited_for_rather_than_guessed_at`, `a_guest_that_never_answers_gives_the_address_back_instead_of_holding_it`, `a_guest_that_refuses_the_plan_stops_the_run_with_its_own_reason`, `a_guest_the_host_cannot_reach_holds_no_address_and_starts_no_workload`, `a_host_that_cannot_discover_its_network_never_opens_a_control_channel` and `the_address_the_guest_took_is_reported_into_its_own_runs_log`. Every failure after the reserve releases it.
   - `vm/mod.rs`: `what_the_run_acquires_after_the_vmm_started_is_still_released_with_it` — the lease is now acquired *after* `boot_with_owner` was spawned, so it is handed to the already-running VMM task through a `oneshot` whose receiver that task owns. Round 1's ownership property survives the reordering: released on confirmed teardown, not before.
   - `run/mod.rs`: `a_guest_that_refuses_its_plan_ends_the_run_with_its_own_reason_not_a_workload_code` — `Refusal::of` and `failure_message` now also unwrap the new `AddressError`, so a guest's own typed reason keeps its name and a host-side bootstrap failure gets the new audit name `guest_address_not_applied` (added to `every_reason_a_host_cannot_address_a_guest_has_its_own_audit_name`). Still exit 125, still pre-workload.

   **Deleted, with their tests and their docs:** `NetworkMemory`, `RealNetworkMemory`, `~/.lns/host-network`, `learn_shared_network` and its call in `boot_with_owner`, `parse_declared_network` / `declared_host_network` / `plist_string` / the vmnet plist path, `host_network_for_boot`, `HostView`, `GuestNet::{to_cmdline_value, parse}`, `SessionParams::{expected_guest_addresses, address_selection}`, `AddressSelection`, and `log::testing::capture_run_frames_spawned` (round 1 P3's harness, unused once the address line stopped living in a spawned reader). `NetworkApplied` arriving on a workload session is now an error, pinned by `an_address_report_on_a_workload_session_is_refused_rather_than_believed`.

   Round 1's P3 property is preserved by construction, not by a harness: `address_guest` is awaited inline inside the run span, so the `Address` line reaches the run's own log. `the_address_the_guest_took_is_reported_into_its_own_runs_log` pins it there instead.

   One deliberate simplification, stated rather than hidden: in the static path a guest-side failure is now always a typed refusal. The old "egress not required, so degrade quietly" distinction remains for DHCP and is gone for static, where the host is waiting for an answer and a silent degrade would be a lie. The plan is validated on both sides (`GuestNet::validate`) before an interface is touched.

## Work-item mapping

| Issue work item | Where | State |
|---|---|---|
| Refuse loudly and precisely at startup | `lns-session/src/lib.rs` `BrokerExitReason` (4 variants) → `ServerFrame::Refused` → `run/mod.rs::failure_message` → `ocsf_audit::broker_exit_event` | done |
| Host network parameters and lease expiry | `vm/host_net.rs` (`observe_host_network` against the live bridge the running guest created, `parse_leases`, `active_leases`, `LeaseExpiry`) | done |
| Stable VM identity where needed | `vm/guest_addr.rs::mac_for` + `VmSpec.mac` + `vz/ffi.rs::guest_mac` | done |
| Atomic candidate reservations | `vm/guest_addr.rs::Allocator::reserve` (single lock spans every exclusion read and the write) | done |
| Configuration carried before network setup | `ExecSpec::with_guest_net_bootstrap` (marker only) → `network::plan_from_env` → `ClientFrame::ConfigureNetwork` over vsock, applied before the workload session is opened | done |
| Guest ARP conflict check and announcement | `lns-session-broker/src/network.rs::claim_address` (`arping -D`, then `arping -U`) | done |
| Gateway check | `network.rs::confirm_gateway` | done |
| DHCP client deleted in the completed path | **deviation — see below** | partial |
| Reporting and docs | `guest_addr/real.rs` log lines + conflict warnings; `docs/running-workloads.md` "How a guest gets its address (macOS)" | done |
| Reserve per booting VM, startup failure, detach, provisioning, reuse, service restart | `Lease` owned by the VMM task + process-local table + provisioner owner key | done |

Both guest kinds are covered: the ordinary workload boot (`run/orchestrator.rs`) and the tool-provisioner boot (`tools/provisioner/real.rs`), which boots without a session and therefore always declares that it needs egress. Both now address the guest after the connector arrives and before their first command.

## Design decisions

**1. The plan travels on the kernel command line, not over the session channel.** The guest must be addressed before any host channel exists, and `lns-init` already ignores unknown keys.

`crates/lns-service/src/vm/mod.rs`
```rust
    pub fn with_guest_net(mut self, net: Option<&lns_session::GuestNet>) -> Self {
        if let Some(net) = net {
            self.kernel_env.push((lns_session::GUEST_NET_ENV.into(), net.to_cmdline()));
```
The encoding is whitespace-free by construction (`"a1,a2,a3/24|gw|dns"`), which is what `build_kernel_cmdline`'s existing `debug_assert` demands.

**2. Three candidates, chosen from the top, under one lock.** The guest cannot ask again once its command line is built, so it carries spares; the lock is what makes two simultaneous boots safe.

`crates/lns-service/src/vm/guest_addr.rs:110`
```rust
        let neighbours = self.neighbors.observed();
        let mut held = self.held.lock().expect("reservation table poisoned");
        let mut taken: HashSet<Ipv4Addr> = HashSet::new();
```

**3. An unreadable or absent lease expiry means "held", never "free".**

`crates/lns-service/src/vm/host_net.rs`
```rust
            LeaseExpiry::Unreadable => true,
```
Guessing the other way hands out an address someone is using.

**4. The guest proves the address before it takes it, and never falls back.**

`crates/lns-session-broker/src/network.rs`
```rust
    for address in &plan.candidates {
        if probe_is_free(runner, address)? {
```
followed by a gratuitous ARP announce and a gateway check; exhaustion returns `BrokerExitReason::NoStaticAddress { offered }`, which stops the run.

**5. A reservation is RAII, so a failed launch gives the address back immediately.**

`crates/lns-service/src/vm/guest_addr/real.rs`
```rust
impl Drop for Lease {
    fn drop(&mut self) {
        release(&self.owner);
```

**6. The MAC is derived, not random**, because a lease follows the hardware address and must not move under a restarting guest (`guest_addr.rs::mac_for` → `02:00:xx:xx:xx:xx`, locally administered, unicast, stable per run id).

## The coexistence argument (the unresolved risk, stated plainly)

macOS 15 exposes no supported way to tell Apple's `bootpd` that an address is spoken for, and modifying `/etc/bootpd.plist` or the lease file is out of bounds. So the race in the issue is real and **is not fixed here**:

- *Mitigated at pick time*: lns allocates from the **top** of the range while `bootpd` allocates from the bottom; it excludes the network, broadcast and gateway addresses, every unexpired lease, every ARP-visible neighbour, and every address reserved for another guest — the reservation table read and write happen under one lock. The lease file and neighbour table are read *before* that lock is taken, so a lease that appears in the same instant is narrowed by the guest's ARP probe rather than excluded by the host.
- *Mitigated in the guest*: an ARP probe before commit and a gratuitous announce after, so a guest never silently shares an address that is already answering.
- *Detected, not prevented, afterwards*: `Allocator::conflicts()` re-reads the live lease file and reports any active lease on a reserved address held by a different MAC; the service logs it naming the address and the other holder. **Lease history and ARP silence are not exclusive ownership** and are not treated as such anywhere in this change.
- *Therefore*: the static path is **off by default**, high addresses are **not** claimed as reserved by anyone but lns's own table, and the default DHCP path is untouched. Turning it on is an explicit, documented opt-in with the limitation spelled out in `docs/running-workloads.md`.

Targeted tests for exactly this: `an_address_leased_to_another_guest_after_we_reserved_it_is_reported`, `a_lease_for_our_own_mac_is_not_a_conflict`, `a_lease_that_has_not_expired_is_excluded`, `an_unreadable_expiry_is_treated_as_held`, `two_guests_booting_together_never_get_the_same_address`, `a_released_address_is_offered_again`, `a_restarted_service_keeps_the_addresses_its_live_runs_hold`.

## Deviations

1. **The DHCP client is not deleted.** The issue's completed path removes `udhcpc` from the guest. That is only correct once static allocation is the *only* path — and the task forbids enabling it by default while the external-DHCP race is unresolved. So the guest keeps both paths and chooses by whether the host gave it a plan. The stronger rule *is* honoured: **inside the static path there is no DHCP fallback** — a static failure is a typed refusal, never a lease attempt. Deleting `udhcpc` is the follow-up that lands with the default flip.
2. **No `log show` content-filter probe.** The refusal names only what the guest observed; it never blames a component it cannot see.
3. **No new Layer 2 `.feature` for the service refusal.** The behaviour is pinned by unit tests in `run/mod.rs`, `audit.rs`, `ocsf_audit.rs` and `vm/session_client.rs`; adding service Gherkin glue was not needed to pin any new decision.

## What I actually ran

**Round 3**, in the same stubbed container (stub `.pc` + `.so` for the GTK/pango/atk/xdo libraries, `--unresolved-symbols=ignore-all` at link time):

```
cargo test -p lns-service -p lns-session -p lns-session-broker --all-targets
  → 2460 + 8 + 23 + 144 passed, 0 failed
cargo clippy -p lns-service -p lns-session -p lns-session-broker --all-targets
  -- -D warnings -D clippy::cognitive_complexity                 → clean
cargo fmt --all -- --check                                       → clean
cargo llvm-cov -p lns-service --all-targets        → scripts/coverage-floor.sh: 116 at 100%, 0 failures
cargo llvm-cov -p lns-session-broker --all-targets → scripts/coverage-floor.sh:   9 at 100%, 0 failures
```

Commits: `349ba189` (the re-architecture), `f0e46065`, `b8a3c434`, `bd4a96ca` (test hardening — the per-file coverage gate rejects never-taken closure bodies and `else { panic! }` arms in tests, so those assertions were reshaped into total helpers that exercise both outcomes). `gh pr checks 395` is green, `coverage` and `coverage-as-root` included.

The two IGNORES reasons that named deleted items (`learn_shared_network`, `RealNetworkMemory`, `AddressSelection::confirm`) were rewritten to name what actually lives in those files now. The new modules — `vm/net_bootstrap.rs` and the broker's `bootstrap.rs` — went in at 100% and are not in IGNORES. **Round 3 is not validated against a real macOS host either** — see the acceptance list at the end.

**Round 2**, in the same stubbed container (stub `.pc` + `.so` for the GTK/pango/atk/xdo libraries, `--unresolved-symbols=ignore-all` at link time):

```
cargo test  -p lns-service --all-targets                          → 2457 + 8 passed, 0 failed
cargo clippy -p lns-service --all-targets -- -D warnings          → clean
cargo clippy -p lns-service --all-targets -- -D clippy::cognitive_complexity → clean
cargo fmt --all -- --check                                        → clean

cargo llvm-cov --lcov -p lns-service --all-targets
  | coverage-strip-ast | scripts/coverage-floor.sh … crates/lns-service/
    [crates/lns-service/] 115 at 100%, 35 ignored, 0 failures
```

`vm/host_net.rs` 100% (499/499), `vm/guest_addr.rs` 100% (592/592), `vm/mod.rs` 100% (794/794). The `host_net/real.rs` IGNORES reason was rewritten to name what now lives there (`learn_shared_network`, `RealNetworkMemory`) instead of a stale list. **Round 2 is not validated against a real macOS host either** — see the acceptance list at the end.

### Round 1

`lns-service` **does** now build and test in this container: stub `.pc` files plus stub `.so` files for the GTK/pango/atk/xdo libraries it lacks, and `--unresolved-symbols=ignore-all` at link time. Nothing calls into those symbols in the tests that matter, so the crate's own suite runs for real.

```
cargo test  -p lns-service --all-targets                          → 2448 + 8 passed, 0 failed
cargo clippy -p lns-service --all-targets -- -D warnings          → clean
cargo clippy -p lns-service --all-targets -- -D clippy::cognitive_complexity → clean
cargo fmt --all -- --check                                        → clean

cargo llvm-cov --lcov -p lns-service --all-targets
  | coverage-strip-ast | scripts/coverage-floor.sh … crates/lns-service/
    [crates/lns-service/] 115 at 100%, 35 ignored, 0 failures
```

The three tests in `35674d1f` were green on arrival — the code they cover already existed — so each was mutation-proved instead: dropping the cause from the `Neighbors` message, swapping the `SelectError` arms, and removing the owner filter from `conflicts_for` each turned exactly one named test red, and green was restored after each. The `session_client` test in `59abe201` was genuinely red first (it did not compile until `AddressSelection::new` existed) and was mutation-proved too, by deleting the `selection.confirm` call.

Earlier rounds, unchanged:

```
cargo test -p lns-session                → 24 passed
cargo test -p lns-session-broker         → 139 passed
cargo test -p lns-ocsf                   → 32 passed
```

`vm/host_net.rs` and `vm/guest_addr.rs` were originally written implementation-and-tests together rather than strictly red-first, and were proved by mutation at the time: removing the gateway exclusion, removing the other-owner exclusion, treating `LeaseExpiry::Unreadable` as free, and counting an own-MAC lease as a conflict — each caught by a named test.

Everything else in this branch was written red first: the `GuestNet` wire format and the refusal variants were observed failing to compile before the types existed, and the guest static path failed (including one genuine assertion failure in the fake's outcome stack) before `claim_address`/`confirm_gateway` existed.

**CI evidence** — [run 34362569225](https://github.com/lensapp/lens-sandbox/actions/runs/34362569225): every required job green — `lint`, `test`, `coverage`, `coverage-as-root`, `audit`, `shell-tests-macos`, `check` and both `check-release-build` matrix legs (**including macos-latest**, which is what compiles the new `VZMACAddress::initWithString` path). The coverage gate reports `[crates/lns-service/]` with no failures and the two IGNORES entries (`vm/host_net/real.rs`, `vm/guest_addr/real.rs`) carrying their reasons.

## Remaining acceptance work — pending a real macOS host

Nothing in this PR has been run against Vz. The following need a macOS 15 host and are **not** claimed as done:

- a static guest booting end to end (gateway, DNS, HTTPS) through `VZMACAddress::initWithString` with the derived MAC;
- concurrent static guests, a fresh-MAC replacement reusing a released address, and a provisioner guest boot;
- the refusal path as a user sees it (`lns run` output and the audit event) when every candidate is taken;
- the conflict warning firing against a real `bootpd` assignment;
- **round 3**: the true cold start — as the ordinary (non-root) macOS user, with no `bridge100` up and nothing cached (there is no cache any more), `LNS_STATIC_GUEST_NET=1 lns run …` boots, is addressed over the control channel, and reaches the gateway; then the refusal path with a guest that cannot take any offered address.

The host experiments recorded in the issue on 2026-09-09 prove the *approach* is feasible on macOS 15; they are not evidence about this code, and I have not treated them as such.


crates/lns-service/src/vm/guest_addr/real.rs
chenhunghan, line 1: High: conflict monitoring only runs when another guest reserves an address. A long-lived run can acquire an external DHCP conflict with no warning if no later guest boots. Add lifecycle-scoped monitoring and deliver the warning to the affected run, with tests of scheduling and routing outside coverage-exempt glue. Merely documenting startup-only checks does not satisfy ongoing detection. Keep default enablement blocked while coexistence remains unresolved.

Review of lens-sandbox PR lensapp/lens-sandbox#395 for lens-sandbox issue lensapp/lens-sandbox#380.

crates/lns-service/src/run/mod.rs
chenhunghan, line 1: High: typed pre-workload network refusals currently return exit 1. The CLI specification reserves 125 for lns failure before workload start. Return 125 for these typed refusals, preserve genuine workload exit codes, and verify the observable CLI/audit behavior with an appropriate regression.

Review of lens-sandbox PR lensapp/lens-sandbox#395 for lens-sandbox issue lensapp/lens-sandbox#380.

crates/lns-service/src/run/orchestrator.rs
chenhunghan, line 1: Medium: the address Lease is owned by the orchestrator, which can return after the shutdown grace expires. Requesting VM stop is not confirmation of network detach. A later run may reserve the address while the old VM still exists. Tie reservation lifetime to confirmed VMM teardown, for example by owning it in the VM task, and test delayed stop and failed launch for both workload and provisioner VMs.

Review of lens-sandbox PR lensapp/lens-sandbox#395 for lens-sandbox issue lensapp/lens-sandbox#380.

crates/lns-service/src/vm/host_net.rs
chenhunghan, line 1: Medium: global string splitting of bootpd plist keys does not preserve subnet dictionaries or select the active VM interface. It can mix subnet and mask entries or select an unrelated Internet Sharing network; network+1 is also an inferred gateway. Discover the actual VZ shared network parameters through a supported host observation, accounting for when the network first appears, rather than guessing from arbitrary plist matches. Add multiple-subnet, nondefault subnet, missing data and malformed-source tests; do not silently treat errors as the default network.

Review of lens-sandbox PR lensapp/lens-sandbox#395 for lens-sandbox issue lensapp/lens-sandbox#380.

chenhunghan, line 1: Medium: usable_high_to_low eagerly materializes the entire subnet although reservation needs only three candidates. Accepted broad masks can allocate gigabytes and terminate the service. Use a lazy iterator or a justified bounded scan; test broad subnets without proportional allocation.

Review of lens-sandbox PR lensapp/lens-sandbox#395 for lens-sandbox issue lensapp/lens-sandbox#380.

crates/lns-service/src/vm/host_net/real.rs
chenhunghan, line 1: Medium: arp spawn errors become an empty neighbor set, and nonzero process exit is ignored. This silently removes an allocation exclusion. Propagate an explicit observation failure and refuse allocation or implement a documented conservative policy. Test spawn, exit-status and parse failures through a command seam; keep policy decisions out of coverage-exempt glue.

Review of lens-sandbox PR lensapp/lens-sandbox#395 for lens-sandbox issue lensapp/lens-sandbox#380.

crates/lns-service/src/vm/guest_addr/real.rs
chenhunghan, line 1: Medium: reservation persistence truncates in place and save/load errors are discarded. If persistence is needed, use atomic replacement and explicit failure handling. Also resolve the contradiction that startup reconciliation uses the empty process-local run registry: either establish actual surviving-VM state, or remove unnecessary persistence with a documented lifecycle argument. Do not claim restart recovery based only on a synthetic state production cannot produce.

Review of lens-sandbox PR lensapp/lens-sandbox#395 for lens-sandbox issue lensapp/lens-sandbox#380.

crates/lns-service/src/vm/guest_addr.rs
chenhunghan, line 1: Medium: the host retains all three candidates and never learns which address the guest selected. This produces warnings for unused candidates and prevents the planned host-visible assigned-address message. Report the selected address safely before normal operation, including the sessionless provisioner case, then narrow the reservation only after confirmed selection. Test validation of that report, unused-candidate release and actual-address conflict detection.

Review of lens-sandbox PR lensapp/lens-sandbox#395 for lens-sandbox issue lensapp/lens-sandbox#380.

crates/lns-service/src/vm/guest_addr/real.rs
chenhunghan: Fixed in b6a2a3c7. The test `conflict_monitor_checks_throughout_one_guests_lifecycle_and_routes_its_warning` pins periodic scheduling, routing to the affected run, deduplication, and lifecycle stop.

crates/lns-service/src/run/mod.rs
chenhunghan: Fixed in cff0e354. The test `a_dhcp_refusal_is_rendered_by_the_host_and_not_as_a_workload_exit` now pins exit 125. Existing completion tests preserve genuine workload exit codes.

crates/lns-service/src/run/orchestrator.rs
chenhunghan: Fixed in 1b9c32e8 and 58aca738. The tests `vm_owner_lives_until_delayed_vmm_teardown_is_confirmed` and `vm_owner_is_released_after_a_failed_launch` pin delayed teardown and launch failure. Both workload and provisioner boot paths place the lease in the VMM task.

crates/lns-service/src/vm/host_net.rs
chenhunghan: Fixed in 6b9e213d. The allocator now observes the active bridge100 address and mask, retries while VZ creates it, and does not infer the gateway or use a default. Tests pin a nondefault gateway, missing and malformed data, and ambiguous multi-subnet plist input.

chenhunghan: Fixed in f49ddc7e. `usable_high_to_low` is lazy. The test `a_broad_subnet_yields_its_first_candidates_without_materialising_the_range` consumes only three addresses from a /1.

crates/lns-service/src/vm/host_net/real.rs
chenhunghan: Fixed in 07047bf3. ARP observation now uses a command seam and refuses allocation on spawn, nonzero exit, UTF-8, or parse failure. The test `neighbor_observation_propagates_spawn_exit_and_parse_failures` pins these policies outside real glue.

crates/lns-service/src/vm/guest_addr/real.rs
chenhunghan: Fixed in 74b8004d. Reservation persistence and startup reconciliation are removed. The service process owns every VMM, so no guest survives the process that owns the process-local table. The test `a_new_service_process_starts_with_no_reservations` pins this lifecycle.

crates/lns-service/src/vm/guest_addr.rs
chenhunghan: Fixed in 2ae8990b and 7726bcc2. The guest reports `NetworkApplied` before normal session frames for workload and provisioner sessions. The tests `assigned_address_report_is_validated_before_normal_session_frames` and `a_confirmed_guest_address_releases_spares_and_rejects_unoffered_reports` pin validation, narrowing, reuse, and selected-address conflict scope.

crates/lns-service/src/vm/guest_addr/real.rs
chenhunghan, line 1: Direct master review of lens-sandbox PR lensapp/lens-sandbox#395 against lens-sandbox issue lensapp/lens-sandbox#380.

P1 — Fix static cold-start ordering. reserve observes bridge100 (20 attempts, 250 ms apart) before the workload orchestrator creates VmSpec/starts boot_with_owner, and the provisioner has the same ordering. The NAT attachment is only created later in vm/vz/ffi.rs. With no existing shared-network guest, bridge100 is absent, so the first static run fails before reaching the operation that creates it. Establish and hold the actual network before observing/allocating, or use a supported correct pre-attachment discovery mechanism. Longer retries or assuming another VM is running are not fixes. Add a regression for absent bridge on first workload AND provisioner boot. Retain macOS 15 support.

The user requests full delivery of lens-sandbox issue lensapp/lens-sandbox#380, not merely a green experimental PR. Verify every acceptance criterion and give a concrete macOS acceptance recipe for this exact implementation. Resolve external DHCP coexistence before enabling static allocation by default or deleting DHCP. If the constraints prevent safe full delivery, state the exact blocker and required decision; do not weaken safety or silently stop at an opt-in subset. The master will independently review and run host acceptance before calling this complete.

crates/lns-service/src/run/mod.rs
chenhunghan, line 1: Direct master review of lens-sandbox PR lensapp/lens-sandbox#395 against lens-sandbox issue lensapp/lens-sandbox#380.

P2 — Classify host network setup failures as pre-workload failures too. emit_completion returns 125 only for BrokerRefusal. Errors from guest_addr::real::reserve, including exhaustion, unreadable neighbor/lease state and bridge discovery failure, propagate as ordinary errors and return 1 although no guest workload started. docs/cli-spec.md:624 requires 125 for run/exec pre-workload failure. Preserve the detailed host cause and useful audit classification. Add regressions exercising the host completion path for exhausted allocation and observation failure, while genuine workload exit 1 remains unchanged.

crates/lns-service/src/vm/session_client/real.rs
chenhunghan, line 1: Direct master review of lens-sandbox PR lensapp/lens-sandbox#395 against lens-sandbox issue lensapp/lens-sandbox#380.

P2 — Route NetworkApplied address reporting to the actual run log. The reader task spawned at lines 65–76 does not inherit/instrument the run span. Its log::info in session_client.rs:164 therefore lacks the RunFrameTx context required by log.rs:267–276 and :312–313 and is not forwarded to the CLI/run log. Instrument the reader future with the run span or explicitly send the selected-address RunLog through frame_tx. Add a regression through the spawned-reader path that asserts the owning run receives the message; a pure reservation-narrowing test does not establish logging. Also reconcile user docs and PR claims: production now observes bridge100, not bootpd.plist/defaults, and external observations occur outside the reservation mutex.

crates/lns-service/src/vm/guest_addr/real.rs
chenhunghan: Fixed in 60fc7d8b.

Red first: `the_first_static_boot_reads_the_declared_network_the_bridge_has_not_been_created_from_yet` and `a_workload_and_a_provisioner_both_boot_first_on_a_host_with_no_bridge_yet` (both in `crates/lns-service/src/vm/`) failed before the change — the first boot on a host that has never run a shared-network guest saw no `bridge100`, exhausted the 20 attempts, and refused.

What changed: discovery no longer depends on the bridge existing. `host_network_for_boot` prefers the live `bridge100` when it is up and otherwise reads `/Library/Preferences/SystemConfiguration/com.apple.vmnet.plist` — the declaration macOS builds the shared network *from* — taking `Shared_Net_Address` as the gateway and `Shared_Net_Mask` as the mask. No inference, no default subnet: a host that neither runs nor declares a shared network is named as such and refuses (`a_host_that_neither_runs_nor_declares_a_shared_network_says_so`). The retry loop stays as the last resort, and its error now names both sources. `reserve` is now `reserve_on(...)` so the ordering is testable without a Vz host; `real.rs` keeps only the constants.

crates/lns-service/src/run/mod.rs
chenhunghan: Fixed in a3134566 (test tightened in 63e20a21).

Red first: `a_host_that_cannot_address_the_guest_fails_before_the_workload_not_as_one` in `crates/lns-service/src/run/mod.rs` — it asserted 125 and a named audit reason for an exhausted allocator and for an undiscoverable network, and failed on both counts (exit 1, reason `None`, and the user saw the anyhow context chain).

What changed: `broker_exit_reason` became `refusal_of`, returning a `Refusal { reason, summary }` built from *any* pre-workload refusal — `BrokerRefusal` as before, plus `ReserveError::{Network, Alloc}`, which `downcast_ref` still finds through the orchestrator's `.context(...)`. `emit_completion` returns 125 for any `Refusal`, and `failure_message` renders `ReserveError::explain()` (cause plus remedy) instead of the call stack. Each cause carries its own stable audit name — `host_network_undiscoverable`, `host_addresses_exhausted`, `host_address_state_unreadable` — pinned by `every_reason_a_host_cannot_address_a_guest_has_its_own_audit_name`, and `audit::record_run_exited_with_reason*` / `ocsf_audit::refusal_event` (renamed from `broker_exit_event`) take them.

crates/lns-service/src/vm/session_client/real.rs
chenhunghan: Fixed in dc626d3b.

Red first: `the_spawned_reader_reports_the_assigned_address_into_its_own_runs_log` in `crates/lns-service/src/vm/session_client.rs`, on a new `log::testing::capture_run_frames_spawned` harness that spawns the reader *inside* the run span and polls it outside — the first harness I wrote entered the span around `block_on` and was green on arrival, which is exactly the production bug it had to model. Against the real code it captured no frames: `tokio::spawn` does not carry the span, so `FrameForwardLayer` found no `RunFrameTx` and the address line went only to the developer trace.

What changed: the spawn moved behind `spawn_server_frame_reader` in `session_client.rs`, instrumented with `tracing::Span::current()`, and `real.rs` calls the seam. `lns run` now shows `Address  guest assigned 192.168.64.x` in the run's own log.

Docs reconciled in the same commit: `docs/running-workloads.md` now describes discovery as bridge100 first and the vmnet declaration second, and states plainly that the reservation mutex covers the reservation table only — the lease and neighbour reads happen outside it, so a concurrent non-`lns` DHCP client remains a race the ARP probe narrows rather than closes.

crates/lns-service/src/vm/host_net.rs
chenhunghan, line 1: P1 — Direct host verification: the new cold-start fallback is not readable by the normal service user. On our macOS 15.7.9 host, /Library/Preferences/SystemConfiguration/com.apple.vmnet.plist is root:wheel, mode 0640; both file and plutil -p return Permission denied as the user running lns. RealHostFiles still calls std::fs::read_to_string. declared_host_network converts that read error to None, and host_network_for_boot returns to waiting for bridge100 before the VM that creates it starts. Thus the first static guest remains blocked after shared-network teardown on this actual host. Fix normal-user cold start through an accessible supported source or by establishing and retaining the network before discovery. Do not change Apple-owned file permissions, require a privileged service, or hide the failure with a guessed subnet. Add a permission-denied fallback regression and validate the production discovery path as the normal macOS user. This remains outstanding for lens-sandbox PR lensapp/lens-sandbox#395 and lens-sandbox issue lensapp/lens-sandbox#380 despite green CI.

chenhunghan: Confirmed and fixed in `252d9d6c` — thank you for reading the actual mode bits; `root:wheel` `0640` makes round 1's fallback dead code for every normal service user, and `declared_host_network` swallowing that read error is what made it look otherwise.

Red first, both failing for their own reason before any production change:

- `vm::host_net::tests::a_declaration_this_user_may_not_read_falls_back_to_the_network_this_host_was_seen_running` — `FakeHostFiles` returning `PermissionDenied` for the plist, no live bridge; failed with the round-1 "neither active nor declared" error.
- `vm::guest_addr::tests::a_first_static_boot_is_addressed_when_the_declaration_is_root_only` — the same host state through `reserve_on`, the path both boot paths take.

Both were mutation-proved afterwards: deleting the recall fallback turns exactly those two red and nothing else.

What the fix does, against your three prohibitions:

1. **The failure is named, never hidden.** `declared_host_network` now returns `io::Result<Option<HostNetwork>>`, so "unreadable" and "declares nothing" are different answers and render as different sentences (`… is not readable by this user (permission denied)` / `… does not exist` / `… declares no shared network`). Pinned by `a_declaration_this_user_may_not_read_is_named_as_that_and_not_as_an_absent_one` and `a_host_that_declares_nothing_is_told_apart_from_one_whose_declaration_is_unreadable`.
2. **A third, user-readable source: what this host was observed running.** `NetworkMemory` records the live `bridge100`'s own address and mask in `~/.lns/host-network`; discovery order is live bridge → declaration → memory → retry loop. That is an observation of this host, not a guessed subnet, and the guest still ARP-probes, announces and confirms the gateway before it commits, so a stale record refuses the run by name rather than addressing a guest wrongly.
3. **Every boot teaches the host, not just static ones.** A guest with a shared-network interface is what creates `bridge100`, so `boot_with_owner` now calls `host_net::real::learn_shared_network()` (macOS only, background). One ordinary DHCP run is enough to teach a host that has never completed a static boot — which is exactly the state your host is in, since the unreadable plist meant no static boot ever succeeded there.

No permission change to an Apple-owned file, no privileged helper, no `sudo`, no guessed subnet, and static addressing is still off by default.

Remaining gap, stated rather than papered over: a host that has never had a shared network up under this user **and** whose declaration is unreadable still has nothing readable left to read, so the run refuses naming all three sources. One ordinary run puts it right permanently. `docs/running-workloads.md` now states the discovery order, the `root:wheel 0640` reality, and this gap.

I cannot validate the production discovery path on macOS from this container, so I have not claimed it: the acceptance list in the PR description now carries "cold start as the ordinary non-root user with the plist unreadable — one DHCP run to teach `~/.lns/host-network`, then a static run booting from the remembered network after teardown". If you re-run on the 15.7.9 host, that is the sequence I would like observed.

`lns-service`: 2457 + 8 tests pass, clippy (`-D warnings` and cognitive complexity) clean, and the per-file coverage gate reports `[crates/lns-service/] 115 at 100%, 35 ignored, 0 failures` with `vm/host_net.rs` and `vm/guest_addr.rs` at 100%.

chenhunghan, line 1: P1 — Complete cold start instead of requiring a prior DHCP boot. NetworkMemory still leaves first use broken when bridge100 is absent, the vmnet plist is unreadable and no cached observation exists. It can also reuse stale subnet data. This does not satisfy the requested first static boot. The user has asked the master to complete the work.

Implementation decision for the next round: use the existing AF_VSOCK control channel to bootstrap static network configuration after VZ has started, but before ANY workload or tool-provisioner command executes. Vsock does not need guest IP networking. Currently the broker configures networking before listening, and the host reserves before starting VZ; change that ordering explicitly. The host should give the VM a stable MAC, start it with a static-bootstrap marker, observe its actual active NAT network, reserve candidates, send a typed network-configuration frame, and wait for a validated applied-address or typed refusal before opening the workload session. The broker must expose only its bootstrap control path until configuration succeeds. No DHCP fallback in the static path; default DHCP behavior can remain until coexistence is solved. This uses vsock for control only, NOT for proxying workload egress.

Remove protected-plist and stale-cache dependencies from static boot. No guessed subnet, root helper, Apple-file permission changes, macOS 26-only API or requirement to run a DHCP guest first. Keep the reservation owned through actual VMM detach even though allocation now follows VMM startup. Handle timeout, disconnect, cancellation, invalid/unoffered reports and early teardown without starting work or releasing a live address. Cover both ordinary and sessionless provisioner boots and retain Linux no-interface behavior. Existing typed exit 125 and run-log routing fixes must remain correct. If the extra NetworkMemory parsing is removed, remove its stale tests/docs too; if retained anywhere, validate prefix length BEFORE mask_bits to avoid malformed-record shifts.

The earlier design rule requiring network settings on the kernel command line was an implementation assumption, not a requirement to preserve this deadlock. The master authorizes revising that rule to a pre-workload vsock bootstrap. The completed path must work on macOS 15 as a normal user with no prior cache and no existing bridge.

Use meaningful red-first regressions for the whole ordering through injected seams, scoped tests only, CI to green, and accurate PR work-item mapping. Produce a concrete acceptance recipe for the actual built lns service; the master will review and run it on macOS. Do not claim full issue completion while external DHCP coexistence/default enablement remains unresolved. Either establish a defensible coexistence strategy with evidence or state the exact platform constraint and product decision needed. This is lens-sandbox PR lensapp/lens-sandbox#395, Part of lens-sandbox issue lensapp/lens-sandbox#380.

chenhunghan: Fixed in `349ba189`, and the shape of the round-2 answer was wrong, not just incomplete — you are right that `NetworkMemory` still left first use broken with no bridge, no readable plist and nothing cached, and could serve a stale subnet. The implementation decision you recorded is what landed: the existing guest control channel carries the plan, so nothing has to be known before the VM starts.

The ordering is inverted. The MAC is derived from the run id, so the VM boots without any reservation. The host then waits for the connector, observes the live `bridge100` that this guest's own attachment created — the first moment anything readable describes the network — reserves on it, and sends `ClientFrame::ConfigureNetwork(GuestNet)` on a dedicated vsock connection. The broker accepts that connection *before* the primary session and applies the plan before any workload or provisioner command can be opened; it answers with the address it took or a typed `Refused`. The reservation is then narrowed to that address; every failure after the reserve releases it and stops the run at 125.

Red first at each layer, each failing before its production code existed:

- `lns-session` — `the_plan_reaches_the_guest_over_the_session_channel_after_it_has_booted`, plus the `GuestNet::validate` tests that replaced the deleted cmdline codec.
- `lns-session-broker` — new `bootstrap.rs`: `the_plan_the_host_sends_is_applied_and_the_address_reported_back`, `nothing_but_a_plan_opens_this_channel`, `a_host_that_hangs_up_before_it_sends_a_plan_leaves_a_refusal_not_a_wait`, `a_plan_the_guest_cannot_use_is_refused_before_the_interface_is_touched`, `a_guest_that_cannot_take_any_offered_address_refuses_by_name`.
- `lns-service` — new `vm/net_bootstrap.rs` (8 tests over the host half of the exchange) and, for the whole sequence, `the_guest_is_told_the_plan_and_the_reservation_narrows_to_the_address_it_took`, `a_bridge_the_vmm_is_still_creating_is_waited_for_rather_than_guessed_at`, `a_guest_that_never_answers_gives_the_address_back_instead_of_holding_it`, `a_guest_that_refuses_the_plan_stops_the_run_with_its_own_reason`, `a_guest_the_host_cannot_reach_holds_no_address_and_starts_no_workload`, `a_host_that_cannot_discover_its_network_never_opens_a_control_channel`.
- Round 1's ownership property survives the reordering, pinned by `what_the_run_acquires_after_the_vmm_started_is_still_released_with_it`: the lease is handed to the already-running VMM task, which is still the last owner.

Deleted with their tests and their docs: `NetworkMemory`, `RealNetworkMemory`, `~/.lns/host-network`, `learn_shared_network`, the vmnet plist parsing and path, `host_network_for_boot`, `HostView`, `GuestNet::{to_cmdline_value, parse}`, `SessionParams::{expected_guest_addresses, address_selection}`, `AddressSelection`, and `log::testing::capture_run_frames_spawned`. The kernel command line now carries only a marker. `docs/running-workloads.md` describes the new order, and says plainly that a cold host needs no prior run.

PR #396, opened 2026-09-09
[reading only, merges via #422] feat(service): capture one guest's writes as one OCI layer

chenhunghan:

Part of #393 — slice 1, the spike. No parser, one hard-coded command.

The round trip works. A guest boots from a base image and runs one command that
creates a file and deletes a file the base image ships; the host reads what the
guest wrote and writes it as one OCI tar layer with whiteouts; a config and a
manifest are assembled over the base; the result is imported into the local layer
store; a second guest boots that image and sees the created file and not the
deleted one. Proven by the reviewer on an arm64 Mac (macOS Vz) at commit 85441e50,
and on the `e2e-microvm` workflow (x86_64 KVM):

```
Feature: one guest's writes become one OCI layer a second guest boots from
  Scenario: the second guest sees the created file and not the deleted one
   ✔  Given a clean lns cache home
   ✔  And the LNS service is running in that home with the layer-capture hook
   ✔  When the user runs a microVM command "/bin/sh -c 'echo built-by-lns > /spike-created; rm -f /etc/alpine-release'" as user "root"
   ✔  Then the exit code is 0
   ✔  And one OCI layer was captured from that run
   ✔  And the capture reports the layer size and the time from the command's exit
   ✔  When the user runs a microVM command "/bin/sh -c 'cat /spike-created; test -e /etc/alpine-release || echo release-file-gone'" over the built image
   ✔  Then the exit code is 0
   ✔  And the output contains "built-by-lns"
   ✔  And the output contains "release-file-gone"
```

(x86_64 run 34342545786, the whole scenario in 2.4 s; 22 of 88 scenarios in that
workflow already fail on `main`, so read this one by name, not the run's conclusion.)

**The scenario has not run again since the review's fixes landed.** I cannot boot a
guest where I work. The seven fixes below are host-tested, and the scenario is
unverified by me. Run it on a virt-capable host:

```
LNS_E2E_FEATURE="one guest's writes become one OCI layer" make e2e-microvm
```

### 1. How the guest's writes are captured

**Option A: the host reads the run's `upper.img` after the guest stops.** The host
takes one dependency — `ext4-view` 0.9 (Google, MIT/Apache-2.0, two deps:
`bitflags`, `crc`), a read-only ext4 reader — plus ~130 lines of adapter
(`crates/lns-service/src/containerfile/ext4_upper.rs`). The guest side is one
kernel-cmdline key (below).

**Why not option B.** The guest cannot reach its own overlay upper. `lns-init`
mounts the ext4 at `/mnt/upper` and the overlay at `/newroot`
(`crates/lns-init/src/mount.rs`), then `chdir(/newroot); move_mount(".", "/");
chroot(".")`. After that move the old root's mounts — `/mnt/upper` among them — have
no reachable path, and the broker that an exec session forks from is inside the
chroot. So option B is not "the guest tars a directory it already has": it is a new
`lns-init` capability (bind the upper into the new root, in guest PID 1) **plus** a
guest-side tar producer that can emit character devices and `trusted.overlay.*`
xattrs (a new subcommand in a static-musl guest binary, or a dependency on the base
image shipping `tar --xattrs`) **plus** the host-side conversion. Three new pieces
against option A's one dependency, one adapter and one cmdline key. I did not
implement B; the blocker above is read out of the code, not out of a failed attempt.

**What option A cost beyond the reader.** One real obstacle, found by running it:

```
⚠ the built layer was not captured: reading the run's upper volume
  …/runs/…/upper.img: incompatible filesystem: missing required journal features: 18
```

The guest never unmounts the volume — the broker calls `sync()` then
`reboot(RB_POWER_OFF)` (`crates/lns-session-broker/src/main.rs`) — so
`INCOMPAT_RECOVER` stays set, the case `crates/lns-service/src/upperfs/constants.rs`
already names. What the code does about it, exactly:

- `ext4-view` will not open a volume that says it needs recovery unless it can load
  the journal, and its journal support requires JBD2 `64BIT | CHECKSUM_V3` (0x12,
  the "18" above). `crates/lns-service/src/upperfs/journal.rs` declares **no**
  journal features, and its own test pins that. So the reader refuses every
  guest-written volume.
- The capture therefore **clears `INCOMPAT_RECOVER` on the way in**, in a 12-line
  `Ext4Read` wrapper, and reads the home blocks the guest's `sync(2)` left. It does
  not read through the journal, and no journal is replayed: with the flag masked
  `ext4-view` sees no journal at all.
- That is sound **only** because of how the guest stops: `sync(2)` commits the
  journal and then writes the block device's dirty buffers, so the home blocks are
  current. It would not be sound after a crash, and the mask costs the reader its
  one signal that a volume is unsynced. So the capture is refused unless the run's
  shutdown reports that the guest actually stopped (review item 2.4, commit
  a3ab4f56). A run that did not exit through the broker's sync gets no layer, by
  name.

**Open decision for slice 3: the journal-feature gap between the upperfs writer and
the reader.** Three ways to close it, none of them free:

1. Declare `IS_64BIT | CHECKSUM_V3` in `upperfs::journal`, so the reader can replay.
   Not sufficient on its own: `ext4-view` also requires `s_checksum_type == 4` and a
   valid crc32c over the journal superblock, and the guest kernel may clear a
   declared `CSUM_V3` at mount for a filesystem without `metadata_csum` — a real
   `dumpe2fs` of a guest-written image reports `journal features: (none)` after the
   mount, which is consistent with that. Unverified; reading `fs/ext4/super.c` for
   the pinned 6.18.15 would settle it.
2. Unmount the volume cleanly in the guest before power-off. That is a guest change:
   after `move_mount` and `chroot`, nothing in the new root can name `/mnt/upper`.
3. Keep the mask, which is what slice 1 does, with the refusal above as its guard.

Two limits of A remain, both because `ext4-view` exposes no xattrs and no `rdev`:

- **Any character device in the upper reads as a whiteout.** Overlayfs writes
  deletions as 0:0 character devices and nothing else in a `RUN` normally makes one,
  but a `mknod` in a `RUN` would be captured as a deletion.
- **`trusted.overlay.opaque` is invisible.** A directory deleted and recreated
  (`rm -rf dir && mkdir dir`) is captured as a plain merge, not as `.wh..wh..opq`, so
  the built image keeps every base file under it. Slice 3 closes this by reading the
  xattr block.

The metacopy half of that limit is gone: see question 2.

### 2. Whether `lns-init` or guest-tools need a new call

**One kernel-cmdline key, no new call.** `upper.capture=1` (commit caba2afc). When
the service sets it, `lns-init` mounts the workload overlay with
`metacopy=off,redirect_dir=off` instead of `metacopy=on,redirect_dir=on`.

This is the review's items 2.2 and 2.6, and it is the option the review asked me to
prefer. It stayed small: one bool on `CmdlineParams`, one branch in
`overlay_options`, one token in `build_kernel_cmdline`, one field on `VmSpec`. No new
session frame, no new guest binary, no new guest-tools payload.

Why it is needed. With `metacopy=on`, a mode or owner change copies up an inode with
**no data**; the reader reads zero bytes and the layer would carry a 0-byte file that
shadows the content the base image shipped. That is reachable without any workload
`chmod`: `lns-init` lchowns every `owner: workload` fileset path on every run that has
a run-as user. With `redirect_dir=on`, a moved directory leaves its old path in an
xattr the reader cannot see. `metacopy=off` costs a full copy on a metadata-only
change, which is why only a captured run pays it.

The service reads the capture hook **once** and that one answer decides both the
guest's mount options and whether the capture runs
(`crates/lns-service/src/run/orchestrator.rs`), so a captured upper can hold no
metacopy stub.

### 3. Whether the layer imports through the existing path unchanged

**Unchanged.** `crates/lns-service/src/composefs/oci.rs` and
`crates/lns-service/src/composefs/changeset.rs` are untouched. The layer is written
as `application/vnd.oci.image.layer.v1.tar` (uncompressed), a media type
`image/mod.rs` already accepts and `apply_layer_bytes` already handles via its
gzip-magic check, and the whiteouts are the `.wh.NAME` spelling `classify_path`
already reads.

Pinned twice. Host-side, by two tests in
`crates/lns-service/src/containerfile/mod.rs` that feed the layer this code produced
back through `build_filesystem_from_layer_bytes`:
`the_layer_reimports_through_the_composefs_path_and_the_deletion_holds` (the created
file is there, the deleted one is not, nothing else the base shipped disappears) and
`a_long_path_and_a_long_symlink_target_survive_the_layer_and_the_reimport`. End to
end, by the scenario above.

The import needed no new store either. The layer goes into the layer cache, the
manifest and config into the manifest cache under a digest-pinned reference, and an
`ImageRecord` into the image index — so `lns run` of the built reference resolves
from cache with no registry and no network:

```
lns-build.local/built@sha256:761743bbdb153e1972e441f6c22c7d20711a6c218b84fbfc4026f1419e7c4438
```

## What the review asked for, and where it landed

| Review item | Commit | What changed |
|---|---|---|
| 2.1 long paths abort the capture | 21cb2ae3 | `builder.append_data` / `append_link` emit the GNU long-name entries; the manual `set_path` and `set_cksum` are gone. Tested with a 150-byte path and a 150-byte symlink target, and the layer re-imports through `build_filesystem_from_layer_bytes`. |
| 2.2 metacopy stub captured as an empty file, 2.6 opaque directory | caba2afc | `upper.capture=1` → `metacopy=off,redirect_dir=off` for a captured run. Opaque directories stay invisible; that half is slice 3's xattr read. |
| 2.3 no exclusion list | 31933d03 | `containerfile/exclude.rs` filters the change set before the tar: `/.lens`, `/etc/resolv.conf`, `/etc/hosts`, `/etc/ssl`, `/etc/passwd`, `/etc/group`, `/home`, and every path a fileset landed on. The capture logs how many entries went and which rule took them. `a_fileset_seeded_credential_never_reaches_the_layer` is the test that pins it. |
| 2.4 capture runs whether or not the guest stopped, 2.5 the mask replaces a refusal with a stale read | a3ab4f56 | `shutdown_after_session` returns `GuestStop::Stopped` or `GuestStop::GraceExpired`; the run carries that to the capture, which refuses `GraceExpired` by name before it opens the volume. The mask stays, on this path only, and the reader's comment says the refusal is what makes it sound. |
| 2.9 the PR body's reader claim | this body | The "reads through the journal instead of refusing" paragraph is gone. Question 1 now says what the code does, and the journal-feature gap is recorded as an open decision for slice 3. |
| 2.10 one global pointer, and the built image is user-visible | a4c46d8e, 91d49049 | The pointer is `runs/<id>/built-image`, so two runs ending together cannot overwrite each other and `lns rm` of the run takes it. The scenario reads it by run id. The "never user-visible" claim is dropped: see "What runs it". |
| 2.11 a tag-referenced base is refused | 1aef22c6 | `IngestedImage` carries `manifest_reference`, the digest-pinned spelling `CachingRegistry` keys the cache under, and the capture looks the parent up by that instead of by the reference the document wrote. |

## Measurements

From the reviewer's arm64 Mac (macOS Vz), commit 85441e50, alpine 3.20 base,
command `echo built-by-lns > /spike-created; rm -f /etc/alpine-release`:

```
Built  lns-build.local/built@sha256:7c516468… (9 entries, 225792 bytes, 10.08ms from exit)
```

- **Layer size: 225 792 bytes** (220.5 KiB), **9 entries**.
- **Time from the command's exit to the imported layer: 10.08 ms.** The clock starts
  when the workload's session ends, so it covers the guest's power-off and quiesce as
  well as the capture itself.
- The capture finished before the CLI returned (`✓ built …` printed before
  `✓ finished in 8.82s`, and the pointer was present at CLI return).

From the `e2e-microvm` workflow run 34342545786 (x86_64 KVM), same command, on the
branch after eab806a6 split the timing in two — which is why this line carries two
durations and the arm64 one above carries one:

```
[slice-1 measurement] ✓ built lns-build.local/built@sha256:761743bb… \
  (9 entries, 225792 bytes, 47.33ms from exit, 2.28ms to read and import)
```

- 47.33 ms from exit, of which **2.28 ms is the host's own work**: reading the ext4
  upper, writing the tar, assembling the config and manifest, and importing all
  three. The rest is the guest stopping. The whole run — resolve, boot, command,
  capture, import — reported `finished in 1.51s`.

**Nine entries for a two-file change was the honest cost of having no exclusion
list.** The bulk of those 225 KiB was not the workload's writes: the trust store the
supervisor appends the proxy CA to (~220 KiB of it), the seeded `/etc/hosts` and
`/etc/resolv.conf`, and `/.lens`. Booting the built image showed them. Commit
31933d03 filters exactly those, so the layer this branch now builds for the same
command is the workload's change and the `etc` directory it lives in. **The new
numbers are not measured yet** — they come from the next `make e2e-microvm` run, and
the capture prints them in the same line, now preceded by one that says what it
dropped:

```
Excluded  <n> entries this boot wrote for the run, <m> a fileset seeded
```

## What runs it

Internal surface, no CLI verb, no IPC request and no document field: the environment
variable `LNS_SPIKE_CONTAINERFILE_LAYER`, read by the service
(`crates/lns-service/src/containerfile/real.rs`). When it is set, every run whose
guest powers off has its overlay mounted without `metacopy`, its upper captured, and
the result imported; the built reference is written to `runs/<id>/built-image` so a
test can boot it. The id in that path is the full 32-character run id, and the CLI
prints the id short, so the scenario reads the pointer from the one directory under
`runs/` that the printed id prefixes (commit baef065b). **Slice 3 replaces this hook with the executor**, which decides per
instruction instead of per run.

The built image is not invisible: the import writes an `ImageRecord` that
`Request::ListImages` serves, so every capture adds a row `lns artifact ls` shows and
`lns artifact prune` can sweep. That is intended for the spike and is another reason the hook stays
test-only.

The scenario, on a virt-capable host:

```
LNS_E2E_FEATURE="one guest's writes become one OCI layer" make e2e-microvm
```

It is `crates/e2e-tests/features/microvm_containerfile_layer.feature`, scenario "the
second guest sees the created file and not the deleted one". It boots
`public.ecr.aws/docker/library/alpine:3.20`, pinned to its platform digest — the base
every `@microvm` scenario uses. `/etc/alpine-release` is the deleted file because that
image really ships it, as `microvm_image_pull.feature` already asserts. The command
runs as root: the identity a run drops to owns nothing in the base rootfs, so an
unprivileged one can neither create a file at `/` nor delete one from `/etc` (the
first attempt failed exactly that way, and busybox `rm` then blocked on its
overwrite prompt until the harness killed the run).

## Verification

Host-side, all of it Layer 3 in `crates/lns-service/src/containerfile/` and
`crates/lns-init/`:

```
cargo test -p lns-service --lib containerfile   →  66 passed; 0 failed
cargo test -p lns-service --lib                 →  2451 passed; 0 failed
cargo test -p lns-init                          →  217 passed; 0 failed
cargo test -p lns-service --test behaviours     →  259 scenarios, 1032 steps, all passed
cargo test -p e2e-tests --no-run                →  builds
```

The 66 cover the walk of the upper tree (through a fake, so every error path is
reachable), the exclusion rules, the layer tar with whiteouts and long names, the
config and manifest assembly, the import, the composefs re-import above, the
stopped-guest refusal, and the ext4 reader against a volume this repo's writer
formats and marks unclean. Also run: `cargo fmt --check`, `cargo clippy -p lns-service
-p lns-init -p e2e-tests --all-targets` with `-D warnings`.

Each fix in this round was written test-first, and the red failure was read before
the fix:

| Claim | The red test that failed first |
|---|---|
| a path over 100 bytes fails the whole capture | `a_path_past_the_tar_name_field_is_carried_by_a_gnu_long_name_entry`, `a_symlink_target_past_the_link_field_keeps_the_target_it_pointed_at` — `provided value is too long` |
| a captured run must not use metacopy | `a_captured_run_mounts_the_overlay_without_metacopy_or_redirect_dir`, `the_capture_key_reaches_the_overlay_mount`, `build_kernel_cmdline_asks_for_whole_copy_ups_only_for_a_captured_run` |
| a fileset-seeded credential must not reach a layer | `a_fileset_seeded_credential_never_reaches_the_layer` and four siblings in `exclude.rs` |
| a run the grace period outlived must not be captured | `a_run_the_grace_period_outlived_is_refused_by_name`, plus the two `shutdown.rs` tests that now assert the outcome |
| a tag-referenced base must resolve | `a_tag_pull_reports_the_digest_pinned_reference_the_manifest_cache_holds` |

The original spike's pure modules were written together with their tests rather than
strictly test-first, so each of those claims was proved to bite by mutation instead:

| Mutation | Result |
|---|---|
| the reader stops clearing the recovery flag | 4 fail, with the CI message verbatim: `missing required journal features: 18` |
| `whiteout_path` stops adding `.wh.` | 2 fail (the whiteout round trip, and the composefs re-import) |
| the upper walk stops recursing into directories | 5 fail |
| `assemble` stops appending the new `diff_id` | 2 fail |

## Deviations, and what is left out

- **The `@microvm` scenario is unverified by me.** No `/dev/kvm`, no Vz, no
  cloud-hypervisor/virtiofsd where I work, and `lns-service`'s GUI dependencies do
  not link there either without stub libraries. Every claim above about the guest is
  read out of the code and pinned by host tests.
- **Left for slice 3, as the review allowed:** 2.7 streaming (the layer and every
  captured file's bytes are still held in memory twice, so a run that writes 3 GiB
  makes the service allocate about 6 GiB at exit), 2.8 logging the entry kinds the
  capture drops or mistranslates (block devices, fifos, sockets, `mknod` character
  devices, hardlinks, non-UTF8 names), and 2.12 the coverage fixture for
  `ext4_upper.rs` (a committed compressed image, the way `ext4-view` tests its own
  read paths).
- **Opaque directories are still captured as a merge.** `metacopy=off` fixes the
  stub; the opaque xattr needs the xattr read, which is slice 3.
- **A restarted run captures every boot since the run was created.** `upperfs::provision`
  reuses an existing `upper.img`, and the hook is per service, not per run. Worse: a
  run first booted without the hook can hold metacopy stubs that a later capture-armed
  boot mounts with `metacopy=off`, which the kernel refuses to follow per file. Slice 3
  owns this, because the executor gets one upper per instruction.
- **`/home` is dropped whole.** The boot creates the run-as identity's home there and
  slice 1 cannot tell that write from the workload's, so nothing under `/home` reaches
  a layer. A Dockerfile that installs into `$HOME` needs slice 3's per-instruction
  bookkeeping.
- **Two IGNORES entries, both with reasons.** `containerfile/real.rs` is the capture
  composition root. `containerfile/ext4_upper.rs` sits just under 100%: the host tests
  reach the recovery-flag clear, the path mapping, every file-type mapping and both
  open failures, but this repo's ext4 writer creates an *empty* volume and nothing
  in-process can put a file, a symlink or a whiteout device into an ext4 image, so the
  successful `read_file`/`read_link` lines need a volume a guest wrote. Everything
  those two files drive is at 100%.
- **Auto-merge is not armed.** The PR is a draft (as a spike should be) and the
  repository does not allow squash merging, so `gh pr merge --auto` is refused on both
  counts.
- **No parser, no cache key, no `lns sandbox build`, no push, no egress-at-build** —
  slices 2 to 5. One layer per run, the "instruction" is whatever the run itself runs,
  and nothing decides whether a rebuild is needed.
- **`lns-artifact`, the grammar, the spec and `docs/` are untouched**, as the brief
  asked. Slice 0 (#394) owns the document layer and this branch does not depend on it.


chenhunghan (commented):
Review of slice 1 of lens-sandbox issue #393. Two sources: this branch was run on an arm64 Mac with real guests (section 1), and a reading review of the diff at eab806a6 against the locked dependency sources (section 2). The verdict is at the end.

# 1. What happened when the branch ran

Not a reading review: this section is what happened when the branch ran here, first at 17c07e2a and then at 85441e50.

**At 17c07e2a and 22a73e81 the round trip did not complete.**

1. `LNS_E2E_FEATURE="one guest's writes become one OCI layer" make e2e-microvm` hung on the first run until the harness timeout (exit 124). The guest printed `rm: remove '/etc/alpine-release'?`: the run-as user is the unprivileged default and busybox `rm` prompts on a write-protected file. Fixed by 22a73e81.
2. With the run as root, the capture was refused for every guest-written image: `incompatible filesystem: missing required journal features: 18`. `ext4-view` 0.9.3 `src/journal/superblock.rs` requires `IS_64BIT | CHECKSUM_V3` on any journal it has to load; `crates/lns-service/src/upperfs/journal.rs` declares no journal features (its test `journal_superblock_declares_no_features_so_any_kernel_can_recover_it` pins that), and the guest is powered off with `sync` then `reboot`, never unmounted, so `needs_recovery` is always set. `dumpe2fs` of the guest-written image: `has_journal ext_attr resize_inode filetype needs_recovery extent sparse_super large_file`, journal features `(none)`. Fixed by 85441e50 by clearing the recovery flag in the reader.
3. A tag reference is refused by design (`the base image ... is not in the local manifest cache`); the manifest cache is keyed by the sha256 of the normalized reference and the pull stores the digest form. The harness pins by digest, so it is unaffected.

**At 85441e50 the round trip passes.**

```
Scenario: the second guest sees the created file and not the deleted one
 ✔ ... as user "root"
 ✔ Then the exit code is 0
 ✔ And one OCI layer was captured from that run
 ✔ When the user runs a microVM command "..." over the built image
 ✔ And the output contains "built-by-lns"
 ✔ And the output contains "release-file-gone"
1 scenario (1 passed), 9 steps (9 passed)
```

Measurements from the service log, alpine 3.20 base, command `echo built-by-lns > /spike-created; rm -f /etc/alpine-release`:

```
Built  lns-build.local/built@sha256:7c516468… (9 entries, 225792 bytes, 10.08ms from exit)
```

The capture finished before the CLI returned (`✓ built …` printed before `✓ finished in 8.82s`, and `built-image` was present at CLI return), so the ordering of `publish_exit_after_quiesce` and `capture_after_run` did not race here.

**What the built image carries besides the command's writes.** Booting it and listing shows `/etc/resolv.conf` dated at build time and a `/.lens` directory: lns's own boot writes are in the layer, as the PR body says. 9 entries for a 2-file change is that noise. Slice 3 has to exclude lns's paths or keep them out of the upper.

**On the 85441e50 fix.** Clearing `needs_recovery` in the reader is right for the path the broker takes (`sync` checkpoints the journal, so the home blocks are current). It is wrong for a guest that died without the sync, where the journal holds metadata the home blocks do not; the reader would then see a stale tree and silently capture it. For slice 3, declare `IS_64BIT | CHECKSUM_V3` in the upperfs journal superblock so the guest kernel writes a journal the reader can replay, or unmount cleanly before power-off, or both; and refuse to capture when the run did not exit through the sync path.

# 2. What reading found

Ranked. Each item names the file, the defect, a concrete failure, and the fix.

**2.1 The layer tar cannot hold a path longer than 100 bytes, so the whole capture aborts.** `crates/lns-service/src/containerfile/tar_layer.rs:70-72` builds a GNU header and calls `header.set_path(path)` then `builder.append(&header, …)`. In `tar` 0.4.46 that path must fit the 100-byte `name` field; `set_path` returns `provided value is too long` instead of emitting a long-name entry, and `set_link_name` has the same limit for symlink targets. Failure: `RUN npm install -g @anthropic-ai/claude-code`, the motivating example of #393, leaves paths like `/usr/local/lib/node_modules/@anthropic-ai/claude-code/node_modules/@img/sharp-libvips-linuxmusl-arm64/lib/…` in the upper, and the capture fails for the whole run. Fix: `builder.append_data(&mut header, path, data)` and `builder.append_link(&mut header, path, target)`, which emit `././@LongLink` entries; drop the manual `set_path` and `set_cksum`.

**2.2 A metacopy stub is captured as an empty file, so the built image silently truncates a file the base shipped.** `ext4_upper.rs:136` maps `FileType::Regular` to a regular entry; with `metacopy=on` (`crates/lns-init/src/mount.rs:270`) a mode or owner change copies up an inode with no data, the reader reads zero bytes, and the layer carries a 0-byte file that shadows the base's content. This is reachable without any workload `chmod`: `lns-init/src/mount.rs:444-459` lchowns every `owner: workload` fileset path on every run that has a run-as user. Fix: mount the captured run's overlay with `metacopy=off,redirect_dir=off`, or refuse the capture while `overlay_options()` contains `metacopy=on`.

**2.3 No exclusion list, so the run's own state, including staged fileset content, lands in a layer that outlives the run.** `real.rs:39-44` walks the whole upper. Beyond the boot noise the PR body names, `crates/lns-service/src/artifact/fileset.rs:31-32` stages fileset writes that a bind mask would cover into the rootfs. Failure: the mahler-shaped document seeds `$HOME/.claude/.credentials.json` through a fileset under a bind exclude; with the hook armed, that credential is written uncompressed into `~/.lns/layers/<sha256>` and referenced by an `ImageRecord`, and `lns rm` of the run leaves it behind. Fix: filter the change set against what this boot wrote, at minimum `/.lens`, `/etc/resolv.conf`, `/etc/hosts`, `/etc/ssl`, `/etc/passwd`, `/etc/group`, and the landed fileset paths, and log how many entries were dropped.

**2.4 The capture runs whether or not the guest actually stopped.** `orchestrator.rs:704-705` assumes the guest is powered off, but `shutdown.rs:20-22` waits at most two seconds for the VM task and proceeds without aborting it, and the stop guard fires only after the capture. Failure: a large dirty page cache makes the broker's `sync()` take longer than two seconds, and the host reads a live filesystem with the recovery flag masked, producing a layer from half-written metadata reported as success. Fix: await the VM task before capturing, or pass the shutdown outcome in and refuse the capture by name when the grace expired.

**2.5 Masking `INCOMPAT_RECOVER` replaces a refusal with a silent stale read.** The mask in `ext4_upper.rs:62-63` works for the reason its comment gives (in `ext4-view` the journal inode is `Some` only when `HAS_JOURNAL` and `RECOVERY` are both set, so the journal becomes empty and the mapping identity). The cost is that the reader can no longer tell a synced volume from an unsynced one, which is exactly the case of 2.4 and of a run killed without a guest-side sync. Of the fix options, declaring `IS_64BIT | CHECKSUM_V3` in the upperfs journal is not sufficient on its own: `ext4-view` also requires `s_checksum_type == 4` and a valid crc32c over the journal superblock, and the kernel may clear a declared `CSUM_V3` at mount for a filesystem without `metadata_csum` (not verified; the `dumpe2fs` above shows `journal features: (none)` after a real mount, which is consistent with that). A clean unmount is a guest change: after `lns-init` does `move_mount` and `chroot`, nothing in the new root can name `/mnt/upper`. The cheapest correct move for the spike is to keep the mask and make it conditional on the guest having stopped, which is the plumbing 2.4 asks for anyway.

**2.6 An opaque directory is captured as a merge, so delete-then-recreate loses the deletion.** `upper.rs:86-94` emits a plain directory for any directory; `rm -rf dir && mkdir dir` leaves `trusted.overlay.opaque`, which the reader cannot see, so `.wh..wh..opq` is never written and the built image keeps every base file under that directory. The PR body discloses this; the code neither warns nor refuses. Fix: as 2.2, mount without `redirect_dir` and `metacopy`, which leaves opaque as the only xattr case, or read the xattr block.

**2.7 The whole layer and every captured file's bytes are held in memory twice.** `upper.rs:99-105` collects a `ChangeSet` owning every file's content and `tar_layer.rs:26-28` copies it all into a second `Vec<u8>`. A run that writes 3 GiB into its rootfs makes the service allocate about 6 GiB at exit. Fix: stream; `ext4_view::File` implements `Read` and the tar builder can write to a file in the layer cache's tmp path, hashing on the way.

**2.8 Entry kinds the capture cannot represent are dropped or mistranslated without a log line.** `upper.rs:118` drops block devices, fifos and sockets; `ext4_upper.rs:138` maps every character device to a whiteout, so `RUN mknod /dev/loop-control c 10 237` produces `dev/.wh.loop-control`; hardlinks become independent copies; a non-UTF8 name aborts the capture. Fix: count and log each dropped or ambiguous entry with its path, so the answer to question 1 is measured rather than assumed.

**2.9 The PR body's central claim about the reader contradicts the code it ships.** The body says option A works because `ext4-view` "reads through the journal instead of refusing" and every feature the writer sets is supported. Both halves are false, and `ext4_upper.rs:47-51`'s own doc comment says so. The body is slice 1's deliverable, so rewrite that paragraph to say what the code does and record the journal-feature gap as the open decision it is.

**2.10 The `built-image` pointer is one global file, and the built image is user-visible.** `real.rs:91-93` writes `$LNS_HOME/built-image`; two runs ending close together overwrite each other's pointer (the hand run above captured the second guest too, 11 entries). The import writes an `ImageRecord` that `Request::ListImages` serves, so every capture adds a row a user can list and prune; drop the "never user-visible" claim. Fix: write the pointer under `runs/<id>/`.

**2.11 A tag-referenced base is refused although the run already resolved its digest.** `real.rs:76-84` looks up `Reference::whole()` of the document's reference, while `CachingRegistry::pull_manifest_and_config` stores a tag pull under the digest-pinned spelling, so a tag lookup misses by construction. Fix: pass the resolved manifest digest from the ingest result into `capture_after_run`.

**2.12 The coverage ignore for `ext4_upper.rs` argues impossibility that is not real.** `ext4-view` covers exactly these paths in its own suite with a committed compressed image. Either name a small committed fixture upper as the design move, or reclassify the entry as platform-only and say why a fixture is refused.

Also confirmed by reading: whiteout spelling round-trips through `composefs/changeset.rs`; `diff_ids` order and `rootfs.type` are preserved and the uncompressed layer's blob digest correctly doubles as its `diff_id`; the import resolves offline and writes nothing outside `$LNS_HOME`; the exit frame the CLI waits for is sent after `orchestrate` returns, so the capture does not race the CLI (matches the host run).

# 3. Questions for the author

- Does the guest kernel clear a declared `CSUM_V3` at mount for a filesystem without `metadata_csum`? Reading `fs/ext4/super.c` for the pinned 6.18.15 would settle it.
- Does `sync(2)` from PID 1 guarantee that every committed transaction's metadata reached its home blocks? 2.5 rests on it.
- What does "one run's writes" mean for a restarted run? `upperfs::provision` reuses an existing `upper.img`, so `lns start` of a stopped run captures every boot's writes since the run was created.
- Should the hook be per run rather than per service? Once the variable is in `lns service start`'s environment, every run with an image is captured until the service stops, and nothing prints the variable.
- The body quotes 46 tests and a `log::info!` line; at eab806a6 there are 54 tests and the line differs.

# 4. Verdict

The spike did its job: the round trip works on real hardware, and it surfaced two facts the plan did not have (the journal-feature gap between the upperfs writer and the reader, and how much of lns's own boot ends up in the upper). Those are now in #393.

Requesting changes before this merges, because as written the hook can put a fileset-seeded credential into a layer that outlives the run (2.3), truncate files the base shipped (2.2), and fail on any real `RUN` with long paths (2.1). Must fix: 2.1, 2.2 with 2.6, 2.3, 2.4 with the conditional mask of 2.5, 2.9, 2.10, 2.11. Can wait for slice 3: 2.7, 2.8, 2.12. Verified here at 85441e50 with `LNS_E2E_FEATURE="one guest's writes become one OCI layer" make e2e-microvm`; re-run it after the fixes, since none of them is covered by the PR gate.

chenhunghan:

PR #407, opened 2026-09-09
[reading only, merges via #422] feat(artifact): parse a Containerfile and hold it to the v1 instruction subset

chenhunghan:

Part of #393. Slice 2 of its plan: the Containerfile parser and the instruction subset gate. **Stacked on #394** (slice 0) — this PR targets `feat/image-containerfile-path` and GitHub retargets it to `main` when that merges. The diff below is this slice only.

## The problem, and what changes

#394 taught the document layer to tell a Containerfile path from a registry reference and to say *which file* a build would use. Nothing reads that file. So a document naming a Containerfile with a second stage, a `RUN --mount` or a `HEALTHCHECK` is "valid" today, and the author finds out what lns cannot build from a builder — at push or run time, off the machine where the file can still be fixed. Inspect had the same gap from the approver's side: it disclosed the file's *name*, and an approver decides on what a build would **run**.

After this PR, `validate` parses the file and refuses every instruction outside the v1 subset with the instruction, its line and the alternative to write instead; `inspect` prints the Containerfile and lists the context a build would send, with sizes. Offline, no guest, no network. `run` and `push` still refuse the path form exactly as #394 left them — the builder is slice 3.

## Acceptance criteria → code and tests

Slice 2's row in #393's plan table: *"`dockerfile-parser` in `lns-artifact`. `validate` parses the Containerfile and refuses every unsupported instruction with its line and an alternative, and a `COPY` whose source leaves the context. `inspect` prints the Dockerfile and lists the context's files with sizes. Offline, no guest."* Done when: *"Every refused instruction in the subset list has a test that names the line and the alternative. Both files, `Containerfile` and `Dockerfile`, parse."*

| Criterion | Code | Test |
|---|---|---|
| A parser in `lns-artifact`, parsing pure | `crates/lns-artifact/src/containerfile.rs` `parse` — text in, instructions out; no `Path`, no I/O. Reading stays behind the CLI's `Fs` seam (`image_build.rs` `built_from`), as resolution does in #394 | 33 Layer 3 tests in `containerfile.rs` |
| Refused: a second `FROM` | `accept_from` | `a_second_stage_is_refused_with_its_line_and_the_one_stage_alternative` (line 3 + "name it in FROM"); Layer 2 `containerfile_subset.feature` |
| Refused: `COPY --from` | `accept_copy` | `copy_from_another_stage_is_refused_…` |
| Refused: `RUN --mount` | `accept_run` | `a_run_mount_is_refused_with_its_line_and_the_copy_alternative` |
| Refused: `ADD` from a URL | `accept_add` / `is_remote` | `an_add_from_a_url_is_refused_…_run_curl_alternative` |
| Refused: `ADD` that would unpack an archive | `accept_add` / `unpacks_an_archive` | `an_add_that_would_unpack_an_archive_is_refused_…_tar_alternative` |
| Refused: `ONBUILD`, `HEALTHCHECK`, `STOPSIGNAL`, naming the issue | `unsupported` / `no_alternative` | `onbuild_healthcheck_and_stopsignal_are_refused_with_their_line_and_the_issue_that_decides` |
| Refused: a `COPY`/`ADD` source leaving the context, naming the field | `leaves_the_context` | `a_copy_source_that_leaves_the_context_is_refused_naming_the_source_it_read` (`../secret`, `./a/../../secret`, `/etc/shadow`), `an_add_source_that_leaves_the_context_is_refused_the_same_way` |
| Refused: a parse error, with its line | `parse`'s error arm | `a_parse_error_is_refused_naming_the_line_it_sits_on` |
| Accepted: `RUN` shell, exec **and heredoc** | `accept_run`, `upstream::command` | `a_run_written_as_a_here_document_is_accepted_with_its_body`, `run_is_accepted_in_shell_and_exec_form` |
| Accepted: the whole subset, each with its line | `accept` | `every_instruction_of_the_v1_subset_is_accepted_with_the_line_it_was_written_on` (14 instructions, lines 1–14), plus continuation/comment line-number tests |
| `ARG` defaults recorded | `accept_arg`, `arg_defaults` | `an_arg_default_is_recorded_so_the_build_and_the_cache_key_can_read_it` |
| Both `Containerfile` and `Dockerfile` parse | resolution is #394's `CONTAINERFILE_NAMES`; this slice reads whichever it picks | `containerfile_subset.feature`: "a Containerfile of subset instructions only is valid" / "a Dockerfile is read under that name too" |
| `validate` runs the gate for the path form | `crates/lns-cli/src/artifact/image_build.rs` `image_problems`, wired at `author.rs` `validate` (#394's call site) | 6 scenarios in `crates/lns-cli/tests/behaviours/containerfile_subset.feature` |
| `inspect` prints the Dockerfile | `author.rs` `render_containerfile` | `image_containerfile.feature`: "inspect prints the Containerfile itself…" |
| `inspect` lists the context's files with sizes | `image_build.rs` `context_files` (recursive, through `SnapshotFs`), rendered as `context:      app/main.js (23 B)` | `image_containerfile.feature`: "inspect lists the context a build would send, with each file's size" |
| Summary line | `BuiltImage::summary` → `built from ./image/Dockerfile (2 lines, context 2 files)` | same feature, asserted verbatim |
| `run` and `push` unchanged | `refuse_an_unbuilt_image` untouched | #394's two scenarios still green |

## The parser: not `dockerfile-parser`

#393 named `dockerfile-parser`. It is unfit, on two counts checked against crates.io and GitHub on 2026-09-09:

- **Deprecated and archived upstream.** `HewlettPackard/dockerfile-parser-rs` was archived on 2026-07-23, last commit "Deprecate repo (#49)". The last release is 0.9.0, October 2024.
- **It rejects the subset.** A heredoc `RUN` — which #393's subset accepts explicitly — is a parse error: `expected any_breakable`. `RUN --mount`, `ADD`, `USER`, `WORKDIR`, `SHELL`, `EXPOSE`, `VOLUME`, `HEALTHCHECK` and `STOPSIGNAL` all fall into an untyped `Misc` variant, so the gate would re-parse most of the subset by hand.

Chosen instead: **`parse-dockerfile` 0.1.8** (taiki-e), Apache-2.0 OR MIT, released 2026-07-10, ~4k recent downloads. It parses heredocs, types `RUN --mount` and `COPY --from` as flags, types `ONBUILD`/`HEALTHCHECK`/`STOPSIGNAL`, and its errors carry `line()`. Taken with `default-features = false`, which drops its CLI's `lexopt`/`serde`/`serde_json` and leaves one new transitive dependency, `smallvec`. The two runners-up were rejected: `dockerfile-parser-rs` 3.3.0 (a different, MIT, actively-maintained crate under the same name) is built for *generating and modifying* Dockerfiles and exposes no spans, so a refusal could not name a line; `tree-sitter-dockerfile` would bring a C grammar and an untyped CST.

## Specification

`docs/sandbox-spec.md` is normative and the code does not reach all of it (AGENTS.md, Transitional mode), so the spec change is its own commit (cd030640). §3.1.1 gains **the instruction subset**: the accepted forms, and for each refusal the alternative a refusal MUST name, with "a build MUST NOT silently skip an instruction it does not implement". §5's offline-validation summary says the gate runs there. This is the target stated whole — the divergence this PR leaves is slice 3's executor, named below. No user guide was touched: the path form still does not build, and a guide answers "how do I use this today".

## Implementation decisions

**The parsed form is public, for slice 3.** `lns_artifact::containerfile::{Containerfile, Instruction, InstructionKind, Command, Transfer, Flag}` — one stage's instructions in order, each with the line it was written on, with a doc comment on each type. `parse` returns *either* that form or the refusals, so a value of type `Containerfile` can only hold subset instructions: the executor cannot be handed an instruction the gate would have refused.

**Every refusal of one file is reported at once**, as `Vec<String>`, because the CLI's validate prints a list of problems and an author fixes a file, not a line.

**Line numbers, not byte offsets.** The parser reports spans; `LineIndex` converts once per file. Continuations, comments and blank lines are covered by their own tests, because a refusal that names the wrong line is worse than none.

**One ignored file, and the reason it exists.** `parse_dockerfile`'s `Instruction` and `Command` are `#[non_exhaustive]`, so an exhaustive match must carry a wildcard arm that no test can reach — a variant upstream adds later cannot be constructed. Rather than exempt the whole gate, the three such arms are isolated in `containerfile/upstream.rs` (span, keyword, command conversions — one line each) and that file alone carries an IGNORES entry with its reason. `containerfile.rs` itself is at 100% (313/313), and so is every CLI file this touches.

**`MAINTAINER` is refused too** — it is outside the subset, and its alternative is `LABEL org.opencontainers.image.authors`. A `COPY`/`ADD` from a here-document is refused for the same reason: the subset says "from the context".

## Tests, red first

- **Layer 3** (`crates/lns-artifact/src/containerfile.rs`, 33 tests): written and run before any implementation existed — 19 compile errors, `cannot find function 'parse' in this scope`, `use of undeclared type 'InstructionKind'`. Green after.
- **Layer 2** (`crates/lns-cli/tests/behaviours/containerfile_subset.feature`, 6 scenarios; `image_containerfile.feature`, +2): watched fail against the un-wired CLI — the four refusal scenarios red with `expected exit code 1, got 0 (output: "lns.yaml is valid.")`, the two inspect scenarios red on the missing text and context lines. All 553 Layer 2 scenarios green after. Everything runs through the in-memory `Fs`/`SnapshotFs`; no file is touched.

| Command | Result |
|---|---|
| `cargo test -p lns-artifact` | 414 passed |
| `cargo test -p lns-cli` | 978 unit tests passed — cucumber 554 scenarios green |
| `cargo clippy -p lns-artifact -p lns-cli --all-targets -- -D warnings` | clean |
| `cargo clippy -p <each> -- -D clippy::cognitive_complexity` (`target/complexity`) | clean |
| coverage, narrowed to the two crates | `containerfile.rs` 100% (406/406), `walk.rs` 100%, `image_build.rs` 100%, `author.rs` 100%; `crates/lns-cli/` 42 at 100%, **0 failures**; `crates/lns-artifact/` 14 at 100% with the two pre-existing FAILs below |
| `make shell-tests` | 7 harnesses green; `affected-crates.test.sh` fails 4 cases **identically with this branch stashed**, so it is environmental (as #394 recorded) |

**Limitations of the local run, stated plainly:** `lns-service` does not compile in this container (no GTK development libraries for `gdk-sys`/`pango-sys`), so `make lint`'s workspace clippy, `make test` and full-workspace coverage were not run here. The narrowed coverage run reports two pre-existing `lns-artifact` FAILs (`connector.rs` 477/478, `tools/mod.rs` 203/212) whose uncovered lines are `pub` API exercised by `lns-service`'s suite — the same two #394 recorded, neither touched here. **CI is the oracle**, and touching `scripts/coverage-floor.sh` makes `affected-crates.sh` return `__FULL__`, so CI runs the full workspace gate on this branch.

## Review response ([review 5155187842](https://github.com/lensapp/lens-sandbox/pull/407#pullrequestreview-5155187842))

One commit per item, each red first — the test was written and watched fail before the fix.

| # | Finding | Commit | What changed |
|---|---|---|---|
| 1 | A symlink in the context aborted the whole walk, so `inspect` and `validate` failed on any `node_modules` | `55083c65` | The rule is now stated in `docs/sandbox-spec.md` §3.1.1: **a build context carries regular files only; a symlink is listed and not sent, never fatal**. `walk::DirEntry` gains `symlink`, the fileset refusal moves from `real_dir_entries` into `walk_into` (a fileset still refuses one), and `collect_context` lists it as `node_modules/.bin/tsc (symlink, not sent)`. Layer 3 through the in-memory `MapFs`, Layer 2 scenario, and a real-FS listing test in `walk.rs` — `RealFs` had none |
| 2 | `ARG A B C` produced one `Arg` holding `"A B C"` | `def00316` | `accept_arg` splits with `split_words` and emits one `Arg` per name, each with the line; a default binds to its own name only |
| 3 | The heredoc's `expand` flag was read and dropped, so slice 3 could not tell `<<EOF` from `<<'EOF'` | `5a002f7d` | New `pub struct HereDoc { expand, body }`; `Command::Run` carries `Vec<HereDoc>`. **Additive to the public type** — the field replaced a `Vec<String>` of bodies, and no other field moved |
| 4 | `split_words` ignored `\`-escapes; an `ENV` word with no `=` was silently dropped | `efdc34f5` | Backslash escapes are honoured (literal inside single quotes), and `key_values` refuses a non-legacy `ENV`/`LABEL` word naming no value, with its line |
| 5 | The spec's table said `SHELL` was accepted without saying in which form | `599f4c12` | The Accepted table splits the `SHELL` row (exec form only), the Refused table gains "A `SHELL` in shell form", and the refusal now names the exec-form alternative |
| 6 | A refusal for an error carrying no line said "line 0" | `bf703922` | `parse_refusal` omits the prefix when the parser reports no line, so an empty file refuses "no FROM" rather than inventing a line |
| 7 | `validate` walked and read the whole context to run a gate that only needs the Containerfile | `359defee` | `image_problems` reads the Containerfile alone and walks nothing; `inspect` sizes each context file through a new `Fs::size` instead of reading its bytes. Pinned by a test that asserts the fake recorded no read |
| 8 | An `ADD` was judged by its file extension, so `ADD ./toolchain /opt` slipped past | `00e011e0` | `unpacks_an_archive` sniffs the first 262 bytes for gzip/bzip2/xz/zstd/tar magic (not zip — `docker build` does not unpack one), falling back to the suffix when the source cannot be read. The spec's row now says content decides |
| — | Follow-up | `38953ad` | The fixes' own coverage: dead `unquote` deleted, the single-use Fs fakes replaced by the shared `MapFs` (which now reports a size and records every listing, sizing and read), and the paths they left unexercised pinned |

### The review's questions

**Does `summary()` count lines or instructions?** Lines — physical lines of the Containerfile, including comments, blanks and continuations. It reads `built from ./image/Dockerfile (14 lines, context 2 files)` and sits directly above the file `inspect` then prints, so it measures *how much there is to read*; the instructions are visible in the print itself. Counting instructions would also make the summary depend on a successful parse, and `inspect` must still describe a file the gate refuses.

**Where does the context stop?** At the directory the `image` path names, or — when the path names a file — that file's parent. Everything below it is listed, recursively, and nothing outside it is: a `COPY ../secret` is refused for leaving the context. Symlinks are listed and not sent (item 1). What is *not* here: `.dockerignore`, exclusion rules, and any cap on entries or bytes. Those belong to slice 5, which decides what a build may send.

**Which flags will slice 3 honour?** Stated so the executor's branch can build against it:

| Flag | Slice 3 |
|---|---|
| `COPY`/`ADD --chown` | **Honoured** — the placement rule already carries ownership |
| `COPY`/`ADD --chmod` | **Honoured** — applied to the written file |
| `COPY --link` | **Accepted, no-op** — it names a BuildKit layer optimisation, and lns's composefs layering has no equivalent to switch on; behaviour is identical either way |
| `FROM --platform` | **Not honoured** — a build produces the host's architecture; refusing it is slice 3's decision to take, with its own spec row |
| `RUN --network` | **Not honoured, and to be refused when slice 3 lands** — the network a `RUN` sees is the run's decisions file, not a flag inside the file being built |
| `RUN --security` | **Not honoured, and to be refused when slice 3 lands** — `--security=insecure` asks for the sandbox to be turned off |
| `RUN --mount`, `COPY --from` | Already refused here, by this slice's gate |

The two "to be refused" rows are the only ones that need a spec change; they are deliberately not made here, because refusing a flag #393 did not list is a decision, and it belongs in the commit that implements the behaviour.

## Deviations and omissions

- **`crates/lns-service` is untouched**, as asked: slice 3's executor is another agent's branch and will consume `containerfile::parse`'s output.
- **`FROM` is not held to "a digest or a tag".** #393 lists the accepted form as "one stage, a digest or a tag" but names only *a second* `FROM` as refused; refusing a bare `FROM alpine` is not in the refusal list, so it parses and the image string is recorded verbatim. Resolving and pinning it is slice 3's.
- **Flags outside the refusal list are recorded, not judged** — `--platform` on `FROM`, `--chown`/`--chmod`/`--link` on `COPY`, `--network` and `--security` on `RUN`. The subset "grows by decision, never by accident", and refusing a flag #393 did not list would be a decision made here. Which of them slice 3 honours is answered under **Review response** below.
- **The context listing has no entry cap.** `inspect` lists every file a build would send, however many there are. The limits on what a build may send — an entry count, a total size, `.dockerignore` — are slice 5's packing, and a cap invented here would refuse documents slice 5 has not decided about yet.
- **`ARG` substitution is not performed.** Defaults are recorded (`arg_defaults`); expanding `${VAR}` inside a `RUN` is the executor's, where the build environment exists.
- **Nothing is built, cached or published.** Slices 3 to 7.

**On the checks reported here:** `ci.yml` triggers only on pull requests whose base is `main`, so a stacked PR gets no checks at all. To get a real verdict rather than an assertion, this PR was pointed at `main` for one run and then pointed back at `feat/image-containerfile-path`. [Run 34364195372](https://github.com/lensapp/lens-sandbox/actions/runs/34364195372) is that run, on this branch's tree with every review fix in it, and it is green throughout: `lint` (6m34s), `test` (7m30s), `coverage` (5m31s, full workspace — `scripts/coverage-floor.sh` is touched, so `affected-crates.sh` returns `__FULL__`), `coverage-as-root` (5m05s), `audit`, `shell-tests-macos` and both `check-release-build` cross-builds (linux 12m45s, darwin 16m33s).


chenhunghan (commented):
Reading review of slice 2 of lens-sandbox issue #393, against `feat/image-containerfile-path` so #394's own changes are not re-reviewed. Nothing was run; every claim below was verified by reading this branch and the pinned `parse-dockerfile` v0.1.8 source. Verdict at the end.

# Findings, ranked

**1. A symlink anywhere in the build context makes `validate` fail and `inspect` print nothing.** `crates/lns-cli/src/artifact/image_build.rs:106` walks the context through `RealFs::dir_entries`, which delegates to `lns_artifact::walk::real_dir_entries` and returns `InvalidData` for any symlink ("symlink {name} — filesets carry only regular files", `walk.rs:133`). Failure: `spec.image: ./image` with `image/node_modules/.bin/tsc` a symlink, which any `npm install` in the context produces; `validate` exits 1 with a message about filesets and an absolute host path, and `inspect` returns the same error from `built_from` before `render_effective`, so the approver sees no image, no egress, nothing. Fix: decide the rule, state it in §3.1.1, and refuse naming `spec.image` and the offending path, or skip symlinks in `collect_context` and leave the packing rule to slice 5.

**2. `ARG` with more than one name is parsed as one name with a garbage default.** `containerfile.rs:198` splits the argument line on the first `=`, but Docker's grammar is `ARG <name>[=<default>] [<name>[=<default>]...]` and upstream hands the whole line through as one `ArgInstruction` (`src/lib.rs:1394`). Failure: `ARG NODE_VERSION=24 CLAUDE_CODE_VERSION=2.1.263` becomes one `Arg { name: "NODE_VERSION", default: Some("24 CLAUDE_CODE_VERSION=2.1.263") }`, so slice 3 expands `${NODE_VERSION}` wrongly and `${CLAUDE_CODE_VERSION}` to nothing, and slice 4's cache key records the same. Fix: split with `split_words` first and emit one `Arg` per name, or refuse a multi-name `ARG` with its line.

**3. A heredoc's `expand` flag is dropped.** `containerfile.rs:39` keeps `here_docs: Vec<String>`; upstream's `HereDoc` carries `expand: bool` (false when the delimiter is quoted, `src/lib.rs:960`) and `accept_run` at `:223` keeps only the body. Failure: `RUN <<'EOF'` and `RUN <<EOF` are indistinguishable in the parsed form, so slice 3 must pick one and the other builds an image that differs from `docker build`, which the PRD's acceptance criterion forbids. This is the public type slice 3 consumes, so it is cheap now and expensive later. Fix: `here_docs: Vec<HereDoc { expand: bool, body: String }>`.

**4. `ENV` and `LABEL` values lose backslash escapes, and a word without `=` is dropped in silence.** `containerfile.rs:353` uses `filter_map(|word| word.split_once('='))` over `split_words`, which handles no escape. Failure: `ENV GREETING="say \"hi\""` records `say \hi\`; `ENV PATH=/opt/a\ b` records `/opt/a\` and drops `b`. Both write a wrong value into the image config at slice 3 with no refusal. Fix: handle `\` escapes in `split_words`, and refuse a word without `=` in the non-legacy form.

**5. The spec's Accepted table says `SHELL` takes shell form, which nothing accepts.** `docs/sandbox-spec.md:393` groups `SHELL` with `ENTRYPOINT`, `CMD` as "Shell and exec form"; `SHELL` is JSON-array only in Docker and upstream's `parse_shell` returns `expected JSON array` otherwise (`src/lib.rs:1912`). The spec is normative, so it states as accepted a form the gate refuses, and refuses through the parse-error arm without the alternative §3.1.1 requires. Fix: split the row.

**6. A Containerfile with no `FROM` is refused at "line 0".** `containerfile.rs:83` prepends `line {}` to every parse error; upstream's `NoStage` is built with line 0 (`src/error.rs:189`). Failure: an empty Containerfile prints `line 0: expected at least one FROM instruction`; the test asserts only `contains("FROM")`. Fix: omit the prefix when `e.line()` is 0.

**7. Context file sizes are computed by reading every byte, and cap at 256 MiB.** `image_build.rs:115` uses `read_limited(..., MAX_FILESET_BYTES)` and takes `data.len()`, so a 500 MiB file prints `(256.0 MiB)` after reading 256 MiB into memory, and `image_problems` at `:56` pays for the whole walk although it uses only `built.text`. There is no entry cap, unlike `walk`. Fix: add `size(path)` to the `Fs` seam, and walk the context only for `inspect`.

**8. The `ADD` archive gate matches suffixes while Docker decides by content.** `containerfile.rs:262` checks `ARCHIVE_SUFFIXES`; `ADD ./toolchain /opt/toolchain` with an extension-less gzip tar passes the gate, `docker build` unpacks it and lns's executor would copy one file, silently, which the spec sentence this PR adds forbids. Fix: sniff the leading bytes at validate, or refuse `ADD` from the context and point at `COPY`.

Confirmed and not defects: parsing is pure (`containerfile.rs` takes `&str`, imports only `BTreeMap`), every read stays behind the CLI's `Fs` seam where #394 put it, the parsed form is public and `lns-service` already depends on `lns-artifact`, and every refusal in the issue's list is produced with instruction, line and alternative.

# Questions for the author

- `BuiltImage::summary` counts physical lines; the plan's example says `4 lines` for a four-instruction file. Instructions or lines?
- Where does the context stop? `spec.image: ./Containerfile` at the project root makes the whole project the context, `.git` included, with no cap on `inspect`'s listing. Is `.dockerignore` or an exclusion list slice 2's or slice 5's?
- Flags outside the refusal list are recorded and not judged: `RUN --network`, `RUN --security`, `COPY --chown`, `--chmod`, `--link`. Which will slice 3 honour? An ignored flag changes the image, which is what "MUST NOT silently skip" is aimed at.
- The `# syntax=` directive is parsed upstream and discarded. Refuse or record?
- `FROM node:24 AS build` drops the alias. Harmless for one stage, or refuse a lone `AS` as a multi-stage leftover?
- Should `inspect` report the refusals `validate` knows? Today it prints a Containerfile that cannot build and exits 0.
- `is_remote` treats `git@` as remote (`containerfile.rs:274`), which is `ADD` from a git repository, not in the issue's list and untested.

# Verdict

Good slice: pure parser in the grammar crate, refusals with line and alternative, the seam respected. Fix before merge: 1 (a symlink in the context is the common case), 2, 3 and 4 (they shape the public type slice 3 consumes and would silently mis-build), 5 and 6 (the spec and the message). 7 and 8 can follow, but say so in the body. Answer the first three questions in the PR body, since slice 3 is being written against this type now.

chenhunghan:

PR #408, opened 2026-09-09
[reading only, merges via #422] feat(service): lns builds a Containerfile spec.image names, and the run boots it

chenhunghan:

Part of #393 — slice 3, the executor and a local run that builds. Stacked on
[#396](https://github.com/lensapp/lens-sandbox/pull/396) (slice 1, the spike) and
opened against `spike/containerfile-layer-roundtrip`.

`lns run` of a local document whose `spec.image` is a path builds the Containerfile
once and boots from the result:

```
$ cat lns.yaml
apiVersion: lns.run/v1
kind: sandbox
name: agent
spec:
  image: ./image
  egress:
    http:
      - match: registry.npmjs.org
        verdict: allow

$ cat image/Dockerfile
FROM public.ecr.aws/docker/library/node:24-alpine@sha256:…
ARG CLAUDE_CODE_VERSION=2.1.263
ENV AGENT_MODE=research
RUN npm install -g @anthropic-ai/claude-code@${CLAUDE_CODE_VERSION}
COPY skills /opt/agent-skills
USER node
WORKDIR /srv
CMD ["claude", "--help"]

$ lns run
Building   ./image/Dockerfile (8 instructions)
Running    instruction 1 on line 4: /bin/sh -c npm install -g @anthropic-ai/claude-code@2.1.263
Booted     microVM   (0.94s)
Finished   in 39.71s
Built      lns-build.local/built@sha256:761743bb… in 41.20s (2 layers)
Image      built from ./image/Dockerfile, arm64, sha256:761743bb…
Booted     microVM   (0.91s)
…
```

Only a `RUN` boots a guest: it is the document's own run with one instruction in place
of the workload, so it prints the run path's own `Booted` and `Finished` lines (and the
transient `Booting` / `Connecting` progress lines, which do not survive in a log). `COPY`,
`ADD` and every config instruction are host-side and print nothing of their own. The
per-step boot report stays for this slice — a build that boots a guest per `RUN` should
say so, and hiding it would hide where a build spends its time. What that output should
look like once a build is a command of its own belongs to `lns sandbox build` in slice 4,
where there is a build to report on rather than a run that happens to build first.

A second `lns run` of a document whose context holds nothing but the Containerfile
prints the `Image` line and no `Building` line: that build is remembered for this
document on this machine. A context that holds any other file — the `skills/` above —
is built again every time, because the pointer's key is the Containerfile's text and
path and cannot notice an edit to anything beside it. Slice 4's key ends that.

## The slice 3 row of the Plan table, mapped to code and tests

|---|---|---|
| The instruction loop in `lns-service` | `containerfile/executor.rs` — one `BuildHost` port for the registry, the build guest, the context and the store | `containerfile/executor.rs` 43 tests: instruction order, each instruction standing on the image the last produced, which instruction makes a layer, which writes config |
| `FROM` pulls through the existing image path and is the parent | `RealBuildHost::resolve_base` in `containerfile/real.rs` calls `ingest::run` with `image::pull` and takes its `manifest_reference` — the digest-pinned spelling the manifest cache keys, as review item 2.11 of #396 established | `a_run_runs_in_a_guest_booted_from_the_image_the_build_has_so_far`, `a_base_the_build_cannot_resolve_stops_it_before_any_instruction_runs` |
| `RUN` goes through guest exec, as root unless `USER` changed it, with the accumulated `ENV` and `ARG`, in `WORKDIR`, under the document's egress and credentials | `containerfile/step.rs` builds the `RunImageArgs` for one step; `run::run_build_step` (in `run/orchestrator.rs`) boots it through the ordinary run path | `containerfile/step.rs` 11 tests; `a_user_and_a_workdir_apply_to_every_run_after_them`, `a_shell_instruction_decides_how_the_next_run_in_shell_form_is_run`, `a_run_in_exec_form_is_run_without_any_shell`, `a_here_document_run_is_run_as_the_script_it_holds`; @microvm `microvm_containerfile_egress.feature` and `microvm_containerfile_credentials.feature` |
| After a `RUN`, slice 1's capture produces one layer, and the next instruction runs in a guest booted from the image so far | `run/orchestrator.rs` arms `capture_upper` for a build step and hands the change set to the executor through `run::BuildStep`; the loop commits it | `what_the_capture_recorded_is_handed_over_once` (`run/mod.rs`), `the_instructions_run_in_the_order_they_were_written_and_each_stands_on_the_last` |
| `COPY` and `ADD` from the context produce one layer each | `containerfile/context.rs` reads the context through a `ContextFs` port and returns the change set the layer is written from; `containerfile/tree.rs` answers whether the parent image already holds a directory at the destination | `containerfile/context.rs` 23 tests: a file, a directory whose contents land at the destination, a symlink's target, several sources, `--chown`, `--chmod`, and every read that can fail; `containerfile/tree.rs` 10 tests |
| `ENV`, `ARG`, `USER`, `WORKDIR`, `LABEL`, `ENTRYPOINT`, `CMD`, `SHELL`, `EXPOSE` and `VOLUME` write the image config and produce no layer | `containerfile/executor.rs` accumulates a `ConfigDraft`; `containerfile/image.rs` writes it over the base's config | `only_the_filesystem_instructions_produce_a_layer_and_the_rest_write_config`, `every_config_instruction_writes_the_field_it_owns`, and 30 tests in `containerfile/image.rs` including `an_entrypoint_with_no_cmd_of_its_own_drops_the_base_s_cmd` and `a_config_only_commit_adds_a_history_entry_and_no_layer` |
| `ARG` and `ENV` scoping as Docker defines it | `executor::scope` / `declare_arg` | `an_arg_reaches_the_next_run_and_never_the_image_config`, `an_env_reaches_the_next_run_as_well_as_the_image_config`, `an_env_outranks_an_arg_of_the_same_name_the_way_docker_defines_it`, `an_arg_declared_before_from_belongs_to_no_stage_until_it_is_declared_again`, `a_stage_that_declares_a_global_arg_again_inherits_its_value` |
| The build guest mounts its overlay without metacopy and redirect_dir, set only for a build guest | unchanged from slice 1 (`upper.capture=1` on the kernel cmdline, read by `lns-init`); the flag is now set by `PreparedRun.build`, per instruction, instead of by a service-wide hook | `a_captured_run_mounts_the_overlay_without_metacopy_or_redirect_dir` (`lns-init`, from slice 1) |
| The exclusion list keeps lns's own boot writes and any staged fileset out of every layer | the loop applies `exclude::only_the_workloads_writes` to every captured change set, so no caller can forget it | `what_this_boot_wrote_for_the_run_is_kept_out_of_every_captured_layer` asserts it for **every** commit of a two-`RUN` build |
| A refused instruction stops the build with its line | `containerfile/parse` refuses the file; the loop wraps every step in `line <n>: <instruction>` | `a_run_that_fails_stops_the_build_naming_its_line_and_what_it_was`, `a_copy_the_context_cannot_answer_stops_the_build_naming_its_line`, `a_commit_the_store_refuses_stops_the_build_naming_the_instruction`, and 27 tests in `containerfile/parse/mod.rs` |
| The finished image is imported into the local layer store and the run boots from it | `containerfile::commit_step` (slice 1's import, now with an optional layer and the config draft) | `containerfile/mod.rs`, `containerfile/import.rs`; @microvm `microvm_containerfile_build.feature` |
| The run summary prints `image: built from ./image/Dockerfile, arm64, sha256:…` | `step::built_line`, printed by the service as its `Image` line | `the_summary_line_names_the_file_the_architecture_and_the_digest`; the @microvm step `the run reports the image it built from "./image/Dockerfile"` |
| The slice 1 hook is removed; the executor decides per instruction | `LNS_SPIKE_CONTAINERFILE_LAYER` and `capture_after_run` are gone | the hook's e2e feature is replaced by the three below |

## The @microvm scenarios, and how to run them

cloud-hypervisor/virtiofsd. The three feature files are written to be right by
reasoning and **are unverified by me**. Each is one command on an arm64 Mac:

```
LNS_E2E_FEATURE="a run whose spec.image is a Containerfile builds it and boots the result" make e2e-microvm
LNS_E2E_FEATURE="a build reaches only what the document's egress allows" make e2e-microvm
LNS_E2E_FEATURE="a build carries no credential the document did not declare" make e2e-microvm
```

- `crates/e2e-tests/features/microvm_containerfile_build.feature` — a Containerfile
  with `FROM`, `ENV`, `RUN`, `COPY`, `USER`, `WORKDIR` and `CMD`. The booted guest
  reads the file the `RUN` created, the file the `COPY` brought, the variable the
  `ENV` set, the directory the `WORKDIR` made the workload's own, and then a second
  run with no command of its own runs the `CMD` and prints no `Building` line.
- `crates/e2e-tests/features/microvm_containerfile_egress.feature` — a `RUN` that
  reaches `example.com` under a deny-all policy fails the build, and the output
  names the host and the line; the same Containerfile with `example.com` in the
  document's egress builds and boots. The deny-all half exists because an
  *undecided* destination raises a consent card, which no headless scenario can
  answer; the allow half really reaches the internet, the way `microvm_tools.feature`
  really fetches from crates.io.
- `crates/e2e-tests/features/microvm_containerfile_credentials.feature` — the service
  process is started carrying a variable the document never declares, a `RUN` writes
  what it can see of that variable, of the document's `ENV` and of its `ARG` into the
  image, and the booted guest reads the verdict out of the image. It is read out of the
  image rather than echoed by the boot command because the run's output repeats the
  command it was given, which is what made the old version fail on its own assertion.

## Slice 2, and the parser this branch reads

Slice 2's branch `feat/containerfile-parser` exists (5 commits, tip `51f5f509`) but
**no pull request is open for it**, and it is stacked on `feat/image-containerfile-path`
(slice 0, #394) — neither is in this branch's base. Merging them in would put slice 0
and slice 2 in this diff.

So this branch reads the same upstream crate (`parse-dockerfile`) through one small
adapter in `crates/lns-service/src/containerfile/parse/`, with **the same type shape
slice 2 published** (`Containerfile`, `Instruction`, `InstructionKind`, `Command`,
`Transfer`). When slice 2 lands, the swap is: delete `containerfile/parse/`, and
`use lns_artifact::containerfile::{…}` in `executor.rs` and `real.rs`. The adapter's
refusals name the instruction and its line but not the alternative for each one —
that catalogue is slice 2's, at `validate`, and this executor only has to stop.

`containerfile/locate.rs` is the same arrangement for slice 0's `lns_artifact::image`:
the same predicate and the same path rules, deleted when #394 lands.

`crates/lns-artifact` and `crates/lns-cli` are untouched. A path-form `spec.image`
already reaches the service on this base, and `definition_dir` already travels with
it, so the service needed no new wire field and no CLI change.

## How the build guest boots, and why

A build step is the document's own run with one instruction in place of the workload.
`containerfile/step.rs` rewrites `spec.image` to the parent digest and `spec.user` to
the `USER` in force, drops `command`, `workdir`, `scripts`, `volumes` and `ports`,
and clears the flags a boot carries (`tty`, `stdin`, `detached`, ports, mounts, name).
`egress`, `credentials`, `filesets`, `mixins` and `definition_dir` stay, which is what
puts the document's own rules in force during the build. `run::run_build_step` then
boots it through `run::orchestrate` — the same path a workload takes, so a `RUN` uses
the guest exec every workload uses — and removes the step's run directory afterwards,
so a build leaves no run behind.

## Deviations

- **`COPY` and `ADD` are made on the host, not staged through a guest.** The issue
  says they "go through the fileset machinery". A staged fileset is *excluded from
  every capture* by slice 1's exclusion list (review item 2.3), so staging a `COPY`
  into a build guest and capturing it would produce an empty layer. This branch reuses
  the fileset *rules* — the path rules, the destination resolution, `--chown` — on the
  host and writes the layer directly. It is also stricter and cheaper: no guest boots
  for a `COPY`, and the copied bytes never enter a guest at all.
- **A non-numeric `--chown` is refused**, naming `uid:gid` as the spelling to use. Only
  the image can resolve `--chown=node:node`, and the layer tar carries numbers.
- **Every config instruction commits its own image** (a manifest and a config, no layer,
  `empty_layer: true` in the history), which is what the classic builder records. It is
  cheap — no guest, no layer blob — and it makes the commit sequence exactly the
  instruction sequence.
- **The `image:` line is printed by the service, not by the CLI's pre-run summary.**
  The CLI prints its summary before the run resolves anything, and the built digest can
  only be known after the build. The summary still names the document; the service prints
  `Image      built from ./image/Dockerfile, arm64, sha256:…` when the build ends.
- **The cache is one pointer**, keyed by the Containerfile's path, its text and the
  architecture, kept in its own directory (not in `builds/`, which `lns prune` sweeps),
  honoured only while the manifest it names is still in the manifest cache, and refused
  outright when the context holds any file the key cannot see. Slice 4 owns the real key
  (`FROM` digest, context content hash) and `lns sandbox build`.
- **Slice 1's `@microvm` feature is deleted** with the hook it drove. Its assertions live
  on in `microvm_containerfile_build.feature`, which boots a built image the same way.
- `context.rs`, `locate.rs`, `step.rs` and `parse/` were written together with their
  tests rather than strictly test-first; the executor and the run-side handover went red
  first (`test(service): the instruction loop, before there is one`, `2857c8da`, where
  every type and the loop are missing).

## The review, item by item, with the commit that answers it

Every fix is its own commit, in the order the review ranks them, red first wherever a
test could exist for the defect.

| Item | Commit | What changed |
|---|---|---|
| §1 + 2.15 the credential scenario | `b650f686` | The negative step matched the echoed command line, so it failed whatever the guest saw. The scenario now sets a variable on the **service** process that the document never declares, and the verdict is written into the image and read out of it, not echoed by the boot command. |
| 2.1 base image env | `5656ae8f` | `resolve_base` returns the base's `Env` beside its reference; the scope is base env → `ARG` → `ENV`, and the draft starts from the base's, so `ENV PATH=/opt/agent/bin:$PATH` no longer truncates `PATH`. |
| 2.2 `COPY` onto an existing directory | `a9dd6626` | New `containerfile/tree.rs` reads the parent image's layer tars (whiteouts included) and answers whether a directory stands at the destination; the executor asks the host, so `COPY entrypoint.sh /usr/local/bin` lands inside it instead of shadowing it. |
| 2.3 here-document `RUN` | `898b5ab0` | The rest of the line is the command and the body is fed on its standard input, with a delimiter the body cannot spell; a quoted delimiter still keeps the shell out of the body. The parse adapter now carries `HereDoc { expand, body }`, slice 2's shape. |
| 2.4 directory source | `9fe9cb3f` | A directory never carries its own name across: its entries land at the destination, and two directory sources merge. Only a file or a symlink keeps its name. |
| 2.5 multi-name `ARG` | `f2171acb` | #407's `def00316` ported: the line is split into words first and each name is its own `Arg` on the line it was written on. |
| 2.6 transfer and run flags | `340d5a26` | `--chmod` decides the mode of every file and directory a transfer writes; `COPY --link`, `ADD --link`, `RUN --network`, `RUN --security` and a non-octal `--chmod` are refused by name with the line. |
| 2.7 `${VAR:-default}` / `${VAR:+alt}` | `8cc40c27` | Both modifiers expand, and their operand is itself expanded; a brace that names no variable, or carries a modifier lns does not expand, stops the build naming what it read. |
| 2.8 fixed `created` | `3d9b3ec4` | The stamp is a constant the assembler owns, not the wall clock, so an unchanged step on an unchanged parent is byte-identical and slice 4 can key on the digest. The composition root no longer decides it. |
| 2.9 the reuse pointer | `bf32f612` | Out of `builds/` (which is `lns_ipc::build_cache_root()`, swept by `lns prune`) into its own directory, and reuse is refused outright while the context holds any file the key cannot see. |
| 2.10 `/home` | `9d555ecc` | `lns-init` records the paths it creates — the seeded `/home/<user>`, the mount points it had to make — in `/.lens/boot-written`, and a capture drops exactly those paths. What the workload writes under them is the workload's, so `USER node` + `RUN pip install --user` commits a real layer. |
| 2.16 the transcript | this body | Corrected above, with what a build step prints and why it still prints it. |
| — | `713473aa` | The tests the coverage floor needs for the lines the above added, and the death of `unquote`, which the new word splitter made unreachable. |

2.11, 2.12 and 2.13 are left for the next round and are named under
[what is left out](#what-is-left-out). 2.14 asks for nothing on this branch.

## Answers to the review's two questions

**Should a build step carry the document's `env`, `tools`, `filesets` and `mixins`?**
No — only `egress`, `credentials` and the `definition_dir` should travel, and the rest
should be dropped. The reason is the one the question names: an image must be a function
of its base, its Containerfile and its context, because that is what slice 4's key
measures. If `spec.env` reaches every `RUN`, two documents naming the same Containerfile
build two different images and a key that hashes the Containerfile and the context is a
key that lies. `egress` and `credentials` are a different kind of thing: they decide what
a `RUN` may *reach*, not what the document *tells* it, and a build that could reach
anything a document forbids would be a hole in the product's own claim. `tools` and
`filesets` fall on the `env` side — a `RUN` that needs a toolchain installs it, which is
what a Containerfile is for — and `mixins` split by the same rule: the egress and
credential rules a mixin merges in stay, the `env`, `tools` and `filesets` it merges in
go. The code today carries all of them (`containerfile/step.rs`), so this is a divergence
with a decision behind it, and it lands with the key in slice 4 rather than here, where
changing it would move the images this slice's scenarios boot without a key to notice.

**How does `lns sandbox build` build without a run?** By splitting `build_for_run`, not by
allocating a throwaway run. It takes a `RunImageArgs` only to shape each step's args and a
frame sender only to carry the step's output; neither needs an outer run to exist. Slice 4
splits it into a `build(document, definition_dir, progress) -> Built` that shapes the step
args from the document itself, and `build_for_run` becomes a caller of it that passes the
run's own frame sender. Each *step* still allocates its own run id, because a `RUN` is a
run — that is how it gets a guest — but nothing allocates a run for the build. A throwaway
run would also be wrong for a second reason: it would register, audit and reclaim a run the
user never asked for, which is exactly what item 2.12 already objects to per step.

## What is left out

- **No `lns sandbox build`, no cache key, no push, no pull, no architecture index** —
  slices 4 to 6.
- **A `COPY` whose source leaves the context is refused by the executor**, but the
  `.dockerignore` file is not read at all. Nobody has asked for it; it is a decision for
  the subset, like any other.
- **Opaque directories are still captured as a merge.** `rm -rf dir && mkdir dir` in a
  `RUN` needs the `trusted.overlay.opaque` xattr, which `ext4-view` does not expose. The
  gap is slice 1's question 1 and it stays open; it needs a reader that reads the xattr
  block, not an executor change.
- **Review item 2.11 — a failed build leaves its committed intermediates behind,
  invisible.** Each step is imported as `RecordKind::Image`, which `ListImages` hides and
  `remove_with` refuses, so a build that fails at instruction 6 leaves five images no
  command can see or remove. Naming intermediates so `prune` and `ListImages` reach them
  (or rolling back on failure) is a change to the image store's record model, not to the
  executor, and it is the next round's.
- **Review item 2.12 — each build step leaves an audit chain, ledger rows and a registry
  entry behind.** The nested `orchestrate` writes `~/.lns/audit/<step-id>/` and never
  deregisters; only the run directory is reclaimed. A build's connections belong on the
  run's own timeline, which means closing each step's chain and attaching it — again the
  next round's, and it touches the audit writer rather than this loop.
- **Review item 2.13 — `real.rs` is exempt from coverage and still holds two decisions.**
  This round moved two out of it (the `created` stamp now belongs to `image.rs`, the reuse
  refusal to `locate.rs`, both at 100%), but `already_built` and the pointer's key remain.
  They go when slice 4 replaces the key, which is the same change; the exemption goes with
  them.
- **Streaming.** A captured layer and every captured file are still held in memory twice
  (slice 1's item 2.7), so a `RUN` that writes 3 GiB makes the service allocate about
  6 GiB. Every instruction pays it separately now, which is better than once per run but
  is not a fix.
- **A build step's consent card lands in the step's own decisions file**, which goes with
  the step. An `ask` answered during a build is not remembered for the next instruction.
  Under a document that decides its destinations — which the egress scenario uses — this
  never arises.
- **`docs/` is untouched.** `docs/sandbox-spec.md` already states the target (slice 0 and
  slice 2 wrote it), and the user guides describe what ships; they describe the path form
  once slice 0's refusals are replaced, which is that branch's call, not this one's.
- **The positive half of the credential story** — a document declaring a credential a
  connector serves, and a build starting with that connector's placeholder — has no
  @microvm step, because the harness cannot install a connector yet. It is asserted at
  Layer 2/3 in `lns-service` and named in the feature's own description.

## Verification

Host-side, all of it Layer 2/3 in `lns-service` and `lns-init`:

```
cargo test -p lns-service --lib containerfile   →  215 passed; 0 failed
cargo test -p lns-service --lib                 →  2601 passed; 0 failed
cargo test -p lns-service --test behaviours     →  259 scenarios, 1032 steps, all passed
cargo test -p lns-init                          →  221 passed; 0 failed
cargo build -p e2e-tests --tests                →  builds (the @microvm scenarios do not run here)
cargo clippy -p lns-service -p lns-init -p e2e-tests --all-targets -- -D warnings  →  clean
cargo clippy -p lns-service -p lns-init -p e2e-tests -- -D clippy::cognitive_complexity  →  clean
cargo fmt --all -- --check                      →  clean
```

Coverage, measured with the gate's own tools (`cargo llvm-cov` + `coverage-strip-ast` +
`scripts/coverage-floor.sh`):

```
[crates/lns-service/src/containerfile/] 12 at 100%, 3 ignored, 0 failures
[crates/lns-service/]                  125 at 100%, 36 ignored, 0 failures
[crates/lns-init/]                       5 at 100%,  2 ignored, 0 failures
```

The new file this round adds — `containerfile/tree.rs`, which reads the parent image's
layers to answer "is there a directory at this path" — goes in at 100% and is not in
IGNORES. `containerfile/real.rs` and `containerfile/parse/upstream.rs` keep the two
entries this branch added, with the reasons they were added with; `real.rs` holds two
fewer decisions than it did (see item 2.13 under what is left out).

CI: this PR was pointed at `main` for one run, since the workflow only triggers for pull
requests based on `main`, and pointed back at `spike/containerfile-layer-roundtrip`
afterwards. The run is
[actions/runs/34376841995](https://github.com/lensapp/lens-sandbox/actions/runs/34376841995),
`shell-tests-macos` and both `check-release-build`s.

**I still cannot boot a guest where I work.** The three @microvm features are unverified
by me this round as they were last round; the credential one is the scenario that failed
on its own assertion and is rewritten above.

Auto-merge is not armed: this repository forbids squash merging, and the base branch is
itself an open PR.

## Slice 4, merged into this branch, and the review round that followed

[#410](https://github.com/lensapp/lens-sandbox/pull/410) (slice 4 — the build key, the
build cache and `lns sandbox build`) auto-merged into this branch before it was reviewed.
The review of #410 —
[pullrequestreview-5158647858](https://github.com/lensapp/lens-sandbox/pull/410#pullrequestreview-5158647858)
— is therefore answered here, in seven commits on `feat/containerfile-executor`, one per
item, red test first wherever a test could exist:

| Item | Commit | What changed |
| ---- | ------ | ------------ |
| 2.1 | `dbda688d` | The sweep identified a built image by `RecordKind::Image`, which every pull writes, so `lns sandbox prune` deleted every pulled artifact. A build now records `RecordKind::Built`, and the sweep partitions on it. Red test: `a_pulled_image_survives_the_built_image_sweep`. |
| 2.2 | `49365ebe` | `lns sandbox build` sent the document verbatim, so any document with `mixins` refused at the first `RUN` — the failure reproduced on the Mac. The verb now takes the same preflight `run_image` takes: `Request::ResolveDefinition`, then the resolved definition, its authored egress and its packed filesets. Red: two Layer 2 scenarios in `sandbox_build.feature`. |
| 2.3 | `d3981489` | A fingerprint of the resolved policy the step runs under — `spec.egress` and `spec.credentials` — enters every step key. Red test: `two_documents_that_declare_different_policy_do_not_share_a_step`. |
| 2.4 | `3ba5cc1d` | The sweep takes `cache_lock().write()` and passes `recorded_run_pins()`, as `prune()` does. No test: the change is three lines in the composition root (`containerfile/real.rs`, an IGNORES entry), and nothing below it could observe the lock. |
| 2.5 | `09a58b09` | The interactive path returned on "No stopped sandboxes." and never swept. It now asks the service what the sweep would drop (`Request::ListPrunableBuiltImages`, answered without removing anything), lists it beside the stopped sandboxes, and sends `PruneRuns` when either list is non-empty. The prompt names the images too. Red: two Layer 2 scenarios in `prune_lists.feature`. |
| 2.6 | `1714c487` | `Entry.source` becomes `Entry.sources`: a hit records the asking document, a remembered build merges, and the sweep keeps an entry while any of its sources is still here. Red test: `a_second_document_that_hits_a_key_is_recorded_as_answering_for_it`. |
| 2.7 | `4c3f3ce3` | `docs/cli-reference.md` gains the `build` verb and the amended `prune` row. |

Two of these are breaking, unversioned format changes, with no shim: an image record
written as `image` by an earlier build is no longer swept, and a cache entry written with
`source` no longer parses, so it is swept and the next build is a miss.

**For `feat/containerfile-push`, which branched from here:** the additive surfaces to
pull are `RecordKind::Built` and `image_store::unreferenced_builds_with`; `lns-ipc` gains
`Request::ListPrunableBuiltImages` and `Response::PrunableBuiltImages`, and
`Request::BuildSandbox` gains `authored_egress` and `packed_filesets` (both
`#[serde(default)]`); `ipc::BuiltImageSweep` gains a required `candidates` method (three
test fakes implement it here); `containerfile::key`'s four free functions become
`key::Steps`, carrying the policy fingerprint; and `cache::Entry.source` is now
`sources: Vec<String>`, with `Entry::built_from`.

### How the binary under test was built

This sandbox has no GTK development libraries and no package that installs them, so
`lns-service` was compiled against stub `pkg-config` packages and stub `.so` files with
`-Wl,--unresolved-symbols=ignore-all`. That is sound for what the runs below measure —
Layer 2 and Layer 3 run in-process against `lns_service::ipc::handle_request` and the
library modules, none of which touch the tray — but **none of the runs listed for slice 4
or for this round exercised a real `lns-service` binary**. `cargo build -p e2e-tests
--tests` compiles the Layer 1 harness without running it; neither `make e2e` nor `make
e2e-microvm` was run here, and I cannot boot a guest. Everything end-to-end — the four
@microvm features included — is unverified by me and re-run on the Mac.

### Answers to the review's questions

**Where does the context stop?** Nowhere, today: `context_hash` walks every file under
the context directory, so a `spec.image: ./Containerfile` at a project root does hash
`.git` and `node_modules` on every build. That is a real defect of the shipped verb, not
a deliberate boundary, and it is not fixed in this round. `.dockerignore` is not slice
4's and is not implemented anywhere on this stack; I would put it in slice 5 with the
push work, or in a slice of its own, together with a refusal above a context size — a
walk that silently takes a minute is worse than a refusal that names the directory. I
have not added the size refusal here because the threshold is a product decision and the
review round was scoped to the seven items.

**The spec's "MUST NOT reuse a build on ... the path a document sits at" versus
`cache.rs` keying an entry's lifetime on that path.** Both are intended, and they are
about two different things — but the code was wrong about the second, which is item 2.6.
No path reaches any key: reuse is decided by the base reference, the Containerfile text,
the context hash and the architecture (and, since 2.3, the policy fingerprint), so two
copies of one document at two paths hit the same key, which is what the spec's sentence
requires. The `source` path is only a liveness question — "does any document that this
key answers for still exist on this machine, or may the sweep drop the image?" — and it
must never decide reuse. Before 2.6 the entry recorded exactly one path, so deleting one
copy dropped the image the other copy still needed; the entry now records every source it
has answered for, and the sweep keeps it while any of them is still here.

**An interrupted `lns sandbox build` leaves the build guest running.** Correct, and it is
not fixed here: a build step is not in the run registry and `BuildSandbox` has no cancel,
so a Ctrl-C between steps orphans a guest. I would not fold it into the push slice — it
is a lifecycle question (register the build's guest so `lns ps`, `lns stop` and the boot
sweep can all see it) that touches the run registry rather than the build cache. It
wants its own slice, and I have not opened one.

**Two things this round does not settle, flagged rather than silently left:**

- The *image* key still takes no document field, which the review verified as correct.
  With 2.3 the step keys carry the policy, but a whole-image hit — same `FROM`, same
  Containerfile, same context, same architecture — is still shared across two documents
  with different egress and credentials. The step-level fix closes the failure the review
  described; the image-level analogue is open.
- The spec's prune row reads "Lists the sandboxes and asks". 2.5 lists the built images
  it would drop as well, on the ground that the question is only answerable when what
  goes is on screen. If the spec means the prompt to name sandboxes only, this is a
  divergence to correct in the spec's direction, not in the code's.

### Verification for this round

Scoped to what changed, all Layer 2/3, with the caveat above about the binary:

```
cargo test -p lns-service                       →  2657 passed; 259 scenarios
cargo test -p lns-cli                           →  975 passed; 546 scenarios
cargo test -p lns-ipc                           →  179 passed
cargo test -p e2e-tests --no-run                →  builds (the @microvm scenarios do not run here)
cargo clippy -p lns-service -p lns-cli -p lns-ipc --all-targets -- -D warnings  →  clean
cargo clippy -p lns-service -p lns-cli -p lns-ipc -- -D clippy::cognitive_complexity  →  clean
cargo fmt --all -- --check                      →  clean
make coverage COVERAGE_CRATES="lns-cli lns-service"  →  [crates/lns-cli/] 42 at 100%, 19 ignored, 0 failures
                                                        [crates/lns-service/] 127 at 100%, 36 ignored, 0 failures
```

No IGNORES entry was added or removed this round. `5cc480cc` is the commit that took
the four CLI error arms and the five sweep-fake methods the gate found uncovered to 100%.

CI: this PR was pointed at `main` for one run, since the workflow only triggers for pull
requests based on `main`, and pointed back at `spike/containerfile-layer-roundtrip`
afterwards. The run is
[actions/runs/34410124705](https://github.com/lensapp/lens-sandbox/actions/runs/34410124705),
`shell-tests-macos`, both `check-release-build`s and `changes`. (The run before it,
[34407735514](https://github.com/lensapp/lens-sandbox/actions/runs/34407735514), failed
`coverage` on exactly the arms `5cc480cc` then covered.)


chenhunghan (commented):
Review of slice 3 of lens-sandbox issue #393. Two sources: the three @microvm features were run on an arm64 Mac at cacead4b (section 1), and a reading review of the diff against `spike/containerfile-layer-roundtrip` with the pinned `parse-dockerfile` 0.1.8 source (section 2). Verdict at the end.

# 1. What happened when the branch ran

```
LNS_E2E_FEATURE="a run whose spec.image is a Containerfile builds it and boots the result"  → passed
LNS_E2E_FEATURE="a build reaches only what the document's egress allows"                    → 2 scenarios, 2 passed
LNS_E2E_FEATURE="a build carries no credential the document did not declare"                → 1 failed
```

The credential failure is the scenario's, not the code's. The step `And the output does not contain "credential=leaked"` fails because the run's output includes the agent's own start line, which echoes the command, and the command text contains the literal `credential=leaked`. The positive step `output contains "credential=none"` passed, so the guest saw no placeholder. Fix the assertion (assert on a value the command line does not spell out, or drop the negative half since the two branches are exclusive), and see 2.15 for what the scenario should actually claim.

# 2. What reading found

Ranked. Each names the file, the defect, a failure and the fix.

**2.1 Expansion cannot see the base image's environment, so `ENV PATH=…:$PATH` truncates PATH.** `executor.rs:309-311` builds the scope from this Containerfile's `ARG` and `ENV` only; `resolve_base` returns a reference and no config. `FROM node:24-alpine` then `ENV PATH=/opt/agent/bin:$PATH` yields `/opt/agent/bin:`. Fix: read the parent config's `Env` into the initial scope and into the config draft.

**2.2 A `COPY` whose destination is an existing directory writes a regular file at that path and shadows the directory.** `context.rs:54-65` decides "into directory" from the string alone. `COPY entrypoint.sh /usr/local/bin` emits a regular file at `usr/local/bin` and the built image loses every binary there, silently. Fix: consult the parent image's tree (the composefs tree exists on the host) or require a trailing slash and refuse otherwise, by name.

**2.3 A here-document `RUN` drops the rest of the instruction line.** `executor.rs:273-277` uses `here_docs.concat()` and discards `Command::Shell`, so `RUN <<EOF > /usr/local/bin/entry … EOF` runs the body as the script and writes no file, and the build succeeds. Fix: run the shell text with the here-doc fed as stdin, as Docker does.

**2.4 A directory source lands one level too deep.** `context.rs:61-62` appends the base name for every source; Docker copies a directory's contents. `COPY skills/ /opt/agent-skills/` lands at `/opt/agent-skills/skills/…`. Fix: for a directory source, copy its entries.

**2.5 `ARG` with more than one name becomes one name with a garbage default.** `parse/mod.rs:190-193`, the same defect the #407 review found and #407 fixed in def00316. Consume #407's parser or port that commit.

**2.6 `COPY --chmod`, `--link`, `RUN --network`, `--security` are parsed and ignored.** `parse/mod.rs:285-289` reads only `chown`. `COPY --chmod=755 entrypoint.sh …` produces a non-executable entrypoint. Fix: honour `--chmod`, refuse the others by name with the line.

**2.7 `${VAR:-default}` and `${VAR:+…}` expand to nothing.** `executor.rs:357` takes everything up to `}` as the name. Fix: implement the two modifiers or refuse a name that is not `[A-Za-z_][A-Za-z0-9_]*`.

**2.8 Every commit is stamped with the current time, so one instruction on one parent never produces the same digest twice.** `real.rs:259` uses `rfc3339_now()` for `created`; it flows into the config, the manifest and the image digest. Slice 4's instruction key cannot match anything a previous build left in the store. Fix: a fixed timestamp (the epoch, or the parent's `created`) so an unchanged step is byte-identical.

**2.9 The reuse pointer ignores the context and lives in a directory another sweeper owns.** `real.rs:162-171` keys on path, text and arch, so an edited `COPY` source boots the previous image; `real.rs:30` writes into `builds/`, which `lns_ipc::build_cache_root()` defines and `build_cache::sweep_with` empties on `lns prune`. Fix: a distinct directory, and either add the context hash now or refuse reuse when the context has any file, until slice 4 lands the real key.

**2.10 `/home` is dropped from every capture.** `exclude.rs:5-12` excludes the prefix because lns cannot tell the boot's `/home/<user>` from the workload's writes, so `USER node` then `RUN pip install --user …` commits an empty layer. Fix: exclude what the boot wrote by record (lns-init knows which paths it created), not by prefix.

**2.11 A failed build leaves every committed intermediate image behind, invisible.** `import.rs:47-50` records each step as `RecordKind::Image`, which `ListImages` hides and `remove_with` refuses. Fix: name intermediates so `prune` and `ListImages` can see them, or roll back on failure.

**2.12 Each build step leaves an audit chain, ledger rows and a registry entry behind.** `orchestrator.rs:59-62` reclaims only the run dir; the nested `orchestrate` writes `~/.lns/audit/<step-id>/` and never deregisters. Fix: close each step's chain and deregister, and attach the build's connections to the run's own timeline.

**2.13 The composition root now holds decisions and is exempt from coverage.** `real.rs:154-158` (`already_built`) and `:227-232` are rules, untested under the `scripts/coverage-floor.sh:32` exemption. Fix: move them out or drop the exemption.

**2.14 `lns push` publishes a path-form `spec.image` as written on this branch.** The refusal lives in #394, which this stack does not include yet. Nothing to change here; merge order matters (#394, then #396, then this).

**2.15 The three @microvm scenarios: one is load-bearing, two assert less than they claim.** The credential scenario declares no credential and installs no connector, so nothing could leak whatever the code does, and its grep for `LNSPLACEHOLDER` is narrower than the placeholder rule in `lns-spec/src/credential.rs:79-82`. Make it assert on the actual service environment not reaching the guest (a variable set on the service process), and note the positive half is Layer 2 only.

**2.16 The PR body's transcript cannot be what this code prints.** Each step runs the full `orchestrate` and logs `Booting`, `Booted`, `Connecting`, `Finished` per instruction. Correct the transcript, and decide whether a build step should print a boot report at all.

Confirmed correct by reading: instruction order, one layer for `RUN`/`COPY`/`ADD` and config-only commits with `empty_layer` for the rest, `USER` and `WORKDIR` reaching the next `RUN`, the build guest's `upper.capture=1` flag set only for build steps, the exclusion list applied to every capture, and the slice 1 hook removed.

# 3. Questions for the author

- A build step keeps the document's `env`, `tools`, `filesets`, `credentials` and `mixins`, so `spec.env` reaches every `RUN` and every step provisions the document's tools. Intended? It makes the image a function of the document, not of the Containerfile and context, which is not what slice 4's key measures.
- `lns sandbox build` needs a build without a run; `build_for_run` takes `RunImageArgs` and a frame sender. Split it in slice 4, or allocate a throwaway run?
- `COPY package*.json ./` refuses with "the build context holds no package*.json". Wildcards are in neither list. Add or refuse by name?
- `ENV A=1 B` drops `B` where Docker errors. Refuse?
- `context::stage` and the Containerfile read are synchronous `std::fs` inside async, unlike the ext4 read. Intended for a large context?
- `lns prune` during a run booted from a built image: the run's registered image stays `./image`, so the built record and manifest are prunable mid-run.

# 4. Verdict

The executor's skeleton is right and the egress and build-and-boot features pass on real hardware. Before merge: 2.1, 2.2, 2.3, 2.4, 2.5, 2.6, 2.7 (each mis-builds an image that `docker build` builds correctly, which the PRD's parity criterion forbids), 2.8 and 2.9 (slice 4 is blocked on them), 2.10 (the mahler use case installs into `$HOME`), the scenario fix in section 1 with 2.15, and the body fix of 2.16. Can follow: 2.11, 2.12, 2.13. 2.14 is merge order. After the round, the three features are re-run here.

chenhunghan:

PR #410, opened 2026-09-09
[reading only, merges via #422] feat(service): a build is keyed by its inputs, and lns sandbox build fills the cache

chenhunghan:

Part of #393 — slice 4, the build cache key and `lns sandbox build`. Stacked on
[#408](https://github.com/lensapp/lens-sandbox/pull/408) (slice 3, the executor) and
opened against `feat/containerfile-executor`.

A build is a function of four inputs, and a build whose key this machine already
answers runs nothing:

```
$ lns sandbox build
Building   ./image/Containerfile (4 instructions)
Running    instruction 1 on line 2: /bin/sh -c npm install -g @anthropic-ai/claude-code@2.1.263
Built      lns-build.local/built@sha256:761743bb… in 41.20s (2 layers)
key sha256:5f0b1a2c3d4e5f60718293a4b5c6d7e8f90a1b2c3d4e5f60718293a4b5c6d7e8
built ./image/Containerfile as lns-build.local/built@sha256:761743bb… (2 layers)

$ lns sandbox build
key sha256:5f0b1a2c3d4e5f60718293a4b5c6d7e8f90a1b2c3d4e5f60718293a4b5c6d7e8
nothing to build: ./image/Containerfile is unchanged, and lns-build.local/built@sha256:761743bb… is what it builds to

$ echo "one more skill" >> image/skills/prompt.md
$ lns sandbox build
key sha256:c41d0f9b…                      # a context file changed, so the image key did
built ./image/Containerfile as lns-build.local/built@sha256:9ab3c02e… (2 layers)
                                          # the RUN before the COPY was reused; nothing booted for it
```

The two `key`/`built` lines are the CLI's. The `Building` / `Running` / `Built` lines
are the service's own, and reach a terminal only on the `lns run` path — see
[deviations](#deviations).

## The slice 4 row of the Plan table, mapped to code and tests

|---|---|---|
| The image key: the `FROM` digest, the Containerfile text, the context content hash, the architecture | `containerfile/key.rs` — `image_key`, and `context_hash` over the `ContextFs` port | `the_from_digest_the_text_the_context_and_the_architecture_each_change_the_image_key`, `the_same_containerfile_with_different_whitespace_in_a_comment_is_a_different_text_and_so_a_different_key`, `a_touched_but_identical_context_file_is_the_same_key`, `the_content_the_name_and_the_mode_of_a_context_file_each_change_the_context_hash`, `a_file_added_beside_the_containerfile_changes_the_context_hash`, `a_context_symlink_hashes_the_target_it_names_and_not_what_it_points_at` |
| Same key, same image, no rebuild | `executor::build` asks the key before the first instruction and returns what it answers | `a_second_build_of_an_unchanged_containerfile_runs_nothing_at_all` |
| The instruction key: the parent layer digest, the instruction, and for `COPY`/`ADD` the content hash of the copied files | `key::run_key`, `key::transfer_key`, `key::config_key`; `executor::reuse` | `a_run_is_keyed_by_the_command_the_env_the_user_and_the_workdir_it_resolved_to`, `a_transfer_is_keyed_by_the_instruction_and_by_what_it_copies`, `a_config_instruction_is_keyed_by_the_config_the_image_then_carries`, `the_line_a_run_was_written_on_is_not_part_of_its_key` |
| A build reuses every leading step whose key matches and rebuilds from the first that differs | `executor::step` asks the step key before it boots a guest or commits | `a_build_reuses_every_leading_step_and_rebuilds_from_the_first_that_differs`, `a_containerfile_whose_instructions_are_unchanged_reuses_every_step_of_them`, `a_copy_whose_file_changed_is_a_different_step_although_the_instruction_is_the_same`, `an_arg_the_run_reads_makes_it_a_different_step_although_the_line_is_unchanged` |
| Where a key is remembered | `containerfile/cache.rs` — one JSON entry per key, under `~/.lns/containerfile-builds/{images,steps}/`, through a `CacheFs` port | `containerfile/cache.rs` 9 tests: what a key answers, what it stops answering when the image is gone, an entry that does not parse, a cache that cannot be written |
| `lns sandbox build [-f FILE] [--rebuild]` builds now, fills the cache, publishes nothing | `lns-cli` `SandboxCommand::Build` → `Request::BuildSandbox` → `containerfile::real::build_sandbox`, which builds with no run around it | Layer 2 `crates/lns-cli/tests/behaviours/sandbox_build.feature`, 6 scenarios; `build_surfaces_the_services_refusal_and_rejects_an_unrelated_answer`, `build_names_the_document_it_could_not_read`, `build_names_the_document_it_could_not_parse`; `handle_request_build_sandbox_refuses_a_document_that_names_no_containerfile` |
| It prints the key and the built digest, and an unchanged build is a no-op that says so | `sandbox::build` prints `key …` and either `built … as …` or `nothing to build: …` | the first three scenarios of `sandbox_build.feature` |
| `--rebuild` ignores the cache for this build | `BuildPlan.rebuild`, read by `executor::build` and `executor::reuse` | `a_rebuild_reads_no_key_and_still_writes_every_one_it_produces`; the `--rebuild` scenario |
| `lns sandbox prune` also drops built images nothing references (a document on this machine, or a run) | `cache::still_referenced` (entries whose Containerfile is gone go, and with them the images nothing else names) + `image_store::remove_unreferenced_builds_with` + the `BuiltImageSweep` port on `prune_runs_with` | `a_build_whose_document_left_this_machine_is_dropped_and_its_image_is_named_by_nothing`, `a_run_booted_from_a_built_image_names_it_even_when_no_document_does`, `nothing_survives_a_sweep_once_the_manifests_are_gone`, `a_built_image_nothing_names_is_dropped_and_one_something_names_is_kept`, `a_cached_sandbox_is_never_dropped_by_the_built_image_sweep`, `a_prune_answers_with_the_built_images_it_dropped`, `a_prune_whose_built_image_sweep_fails_still_answers_for_the_runs` |
| `lns run` of a path document uses the same cache, so the inner loop needs no verb | `real::build_for_run` is now a caller of `real::build`; slice 3's one-pointer cache and its "refuse reuse when the context holds anything" rule are deleted | `crates/e2e-tests/features/microvm_containerfile_cache.feature`; slice 3's build-and-boot scenario still asserts a second `lns run` prints no `Building` line |
| No push, no `spec.image` rewrite, no `imageSource`, no architecture index | untouched | — |

cloud-hypervisor/virtiofsd. The feature is written to be right by reasoning and
**is unverified by me**. One command on an arm64 Mac:

```
LNS_E2E_FEATURE="a build is remembered by its key, and only what changed is built again" make e2e-microvm
```

`crates/e2e-tests/features/microvm_containerfile_cache.feature` builds a Containerfile
of `FROM`, `RUN`, `COPY` and a second `RUN` that reads what the `COPY` brought; builds
it again unchanged and reads the no-op line and the same key; then edits the context
file, builds again, and asserts a different key and **one** instruction booted a guest —
the `RUN` after the `COPY`, not the one before it. What ran is counted from the
service's own log, which names each instruction it boots a guest for on one line. It
ends by running the sandbox and reading the edited bytes out of the booted guest.

The slice 3 scenarios are unchanged and should still pass; the `Building` line now
comes from the executor rather than from `real.rs`, which is what keeps
`microvm_containerfile_build.feature`'s "a second run prints no `Building` line"
true — a build the key answers outright now prints nothing at all.

## Deviations

- **The instruction key is over the instruction as it resolved, not as it was
  written.** The Plan row says "the instruction text". An `ARG` commits nothing, so two
  builds can reach the same parent image with the same written `RUN` line and still run
  two different commands; keying on the text would hand the second build the first
  build's image. A `RUN` is therefore keyed by its argv, its environment, its user and
  its working directory; a `COPY`/`ADD` by its label and the content hash of what it
  copies (which is where its sources, destination, `--chown` and `--chmod` all end up);
  a config instruction by its label and the config the image then carries. The line
  number is deliberately **not** part of the key: moving an instruction down the file
  rebuilds nothing.
- **The parent is the parent image's digest-pinned reference, not "the parent layer
  digest".** It identifies the same thing more exactly: a config-only instruction
  produces no layer, and the image it produced is what the next instruction stands on.
- **`lns sandbox build` prints no progress.** It is a one-shot request, so the
  `Building` / `Running` / `Built` lines the build logs go to the service's log rather
  than to the caller's terminal. The verb's own two lines (`key`, `built`/`nothing to
  build`) are the answer. Streaming them the way `lns run` does needs the frame-forward
  plumbing `serve_prepared_run` owns, which is a change to the request dispatcher rather
  than to the build, and it is not what this row asks for.
- **The CLI reads the document through the `SandboxService` port**, beside
  `write_document` and `load_policy`, rather than through a second filesystem parameter
  on every sandbox verb. It resolves `-f` against this machine's working directory and
  reads the file; the fake reads the behaviours' in-memory documents.
- **`lns sandbox prune`, not `lns artifact prune`, is where the built-image sweep
  landed**, as the issue's CLI surface says. `lns artifact prune` already reaches a
  built image, through `prune_with`'s run-pinning partition; the new sweep is the one
  that knows about *documents*, and it is what answers review item 2.11 of #408 for the
  intermediates a build leaves behind.
- **A build cache entry records the Containerfile it was built from.** That is the only
  way a sweep can ask "does a document on this machine still name this image": the key
  itself is a hash and names nothing. An entry whose Containerfile has left the machine
  goes, and the images no surviving entry and no run names go with it.
- **`image_store::real` and `RealCaches` became `pub(crate)`** so the sweep's
  composition root could stay in `containerfile/real.rs`, which is the file the coverage
  table already exempts.

## What is left out

- **No push, no pull, no `spec.imageSource`, no architecture index** — slices 5 and 6.
  `lns push` refuses the path form as #394 left it.
- **`lns push --dry-run` does not print the key yet.** It is slice 5's row; the key it
  will print is `key::image_key`, which this branch lands.
- **A build cache entry is never invalidated by age or size.** Nothing sweeps a build
  whose document is still on this machine, however old, until the user runs
  `lns sandbox prune`.
- **Review items 2.11, 2.12 and 2.13 of #408 are only half-answered.** The built-image
  sweep now reaches the intermediates a failed build left behind (2.11) once the
  document goes, and `real.rs` lost the reuse pointer and its key to `cache.rs` and
  `key.rs` (2.13) — but `real.rs` keeps its coverage exemption for the composition root
  it still is, and each build step still leaves its own audit chain and ledger rows
  (2.12), which is the audit writer's change, not this one's.
- **The key does not read `.dockerignore`**, because nothing reads it yet (#408).
- **Streaming.** A captured layer is still held in memory twice. Unchanged from slice 3.

## What happened to this pull request

**It merged the moment it was opened, and that was not the intent.** `gh pr merge 410
--auto --rebase` was run right after `gh pr create`, as the standing workflow for this
repository asks. `feat/containerfile-executor` carries no required checks and no
required review, so GitHub had nothing to wait for and merged at once. Slice 4's eleven
commits are therefore on `feat/containerfile-executor` — [#408](https://github.com/lensapp/lens-sandbox/pull/408)
now shows slices 3 and 4 together, and #408's own description does not describe slice 4.
This description does.

Nothing was force-pushed to undo it: `feat/containerfile-executor` is another pull
request's head branch, and rewriting it is the dispatcher's call, not mine. Two ways
back, if slice 4 should be reviewed on its own: reset `feat/containerfile-executor` to
`713473aa` and reopen this branch against it, or review slice 4 as the last eleven
commits of #408 with this description beside it.

## Verification

Host-side, all of it Layer 2/3:

```
cargo test -p lns-service --lib containerfile   →  283 passed
cargo test -p lns-service                       →  2653 passed; 259 scenarios
cargo test -p lns-cli                           →  1052 passed; 542 scenarios
cargo test -p lns-ipc                           →  179 passed
cargo build -p e2e-tests --tests                →  builds (the @microvm scenario does not run here)
cargo clippy -p lns-service -p lns-cli -p lns-ipc -p e2e-tests --all-targets -- -D warnings  →  clean
cargo clippy -p lns-service -p lns-cli -p lns-ipc -- -D clippy::cognitive_complexity          →  clean
cargo fmt --all -- --check                      →  clean
```

**How that binary was built.** This sandbox has no GTK development libraries and no
package that installs them, so `lns-service` was compiled against stub `pkg-config`
packages and stub `.so` files with `-Wl,--unresolved-symbols=ignore-all`. Every run
above therefore links a binary whose tray/GUI symbols are stubs. That is sound for what
these runs measure — Layer 2 and Layer 3 tests exercise `lns_service::ipc::handle_request`
and the library modules in-process, none of which touch the tray — but **none of the runs
listed above exercised a real `lns-service` binary**: `cargo build -p e2e-tests --tests`
compiles the Layer 1 harness without running it, and neither `make e2e` nor
`make e2e-microvm` was run here. Everything that boots a guest or spawns the real
service — the @microvm containerfile scenarios included — is unverified by me.

Coverage, measured with the gate's own tools (`cargo llvm-cov` + `coverage-strip-ast` +
`scripts/coverage-floor.sh`):

```
[crates/lns-service/] 127 at 100%, 36 ignored, 0 failures
[crates/lns-cli/]      42 at 100%, 19 ignored, 0 failures
[crates/lns-ipc/]       7 at 100%,  0 ignored, 0 failures
```

`containerfile/key.rs` and `containerfile/cache.rs` go in at 100% and are not in
IGNORES. No entry was added to the table, and none was removed.

CI: this branch was pushed as `ci-oracle/containerfile-cache` and opened as a draft
against `main` for one run, since the workflow only triggers for pull requests based on
`main`; that draft is closed and its branch deleted. The run is
[actions/runs/34387861418](https://github.com/lensapp/lens-sandbox/actions/runs/34387861418),
`shell-tests-macos`, both `check-release-build`s and `labels`.

## Red first

`test(service): the key a build is remembered by, before there is one` (`787d01ef`) and
`test(cli): what lns sandbox build prints, before the verb exists` (`5126c7fb`) are each
a failing test ahead of the code that answers it. `cache.rs` and the prune sweep were
written with their tests rather than strictly test-first.

### DIFF
--- a/crates/e2e-tests/features/microvm_containerfile_cache.feature
+++ b/crates/e2e-tests/features/microvm_containerfile_cache.feature
@@ -1,0 +1,48 @@
+@microvm
+Feature: a build is remembered by its key, and only what changed is built again
+  Slice 4 of lensapp/lens-sandbox#393, the build cache and `lns sandbox build`. The
+  verb builds what `spec.image` names, fills the cache and publishes nothing. The image
+  is keyed by the `FROM` digest, the Containerfile's text, the context's content hash
+  and the architecture, so a second build of an untouched document runs nothing and
+  says so. One instruction is keyed by the image it stands on and by what it resolved
+  to, so an edited context file rebuilds the `COPY` that carries it and everything
+  after it, and nothing before it.
+
+  What ran is read two ways: from what the verb printed, and from the instructions the
+  service booted a guest for, which it names one line each in its own log. A `RUN` no
+  key answered is a boot; a `RUN` a key answers is not.
+
+  Like every @microvm scenario this boots real guests and reaches a real registry for
+  the base image, so it runs only via `make e2e-microvm`, never in CI's PR gate.
+
+  Scenario: an unchanged build runs nothing, and an edited context file rebuilds from the copy that carries it
+    Given a clean lns cache home
+    And the LNS service is running in that home
+    And the build context holds "app/index.js" containing "copied-by-lns"
+    And the project builds its image from this Containerfile
+      """
+      FROM {base}
+      RUN /bin/sh -c 'echo built-by-lns > /built-marker'
+      COPY app /srv/app
+      RUN /bin/sh -c 'cat /srv/app/index.js > /srv/derived'
+      """
+    When the user builds the sandbox definition
+    Then the exit code is 0
+    And the build reports the key it is remembered under
+    And the build reports the image it produced from "./image/Dockerfile"
+    And the service booted a guest for 2 instructions of that build
+    When the user builds the sandbox definition
+    Then the exit code is 0
+    And the output contains "nothing to build"
+    And the build reports the same key as the build before it
+    And the service booted a guest for 0 instructions of that build
+    When the build context file "app/index.js" is changed to "edited-by-hand"
+    And the user builds the sandbox definition
+    Then the exit code is 0
+    And the build reports a different key from the build before it
+    And the service booted a guest for 1 instruction of that build
+    When the user runs a microVM command "/bin/sh -c 'cat /built-marker; cat /srv/derived'"
+    Then the exit code is 0
+    And the output contains "built-by-lns"
+    And the output contains "edited-by-hand"
+    And the output does not contain "Building"
--- a/crates/e2e-tests/tests/e2e.rs
+++ b/crates/e2e-tests/tests/e2e.rs
@@ -43,6 +43,10 @@ pub struct E2eWorld {
     pub run_budget: Option<std::time::Duration>,
     /// Destinations the definition's own policy allows, so a scenario whose workload really fetches something is not left at an approval prompt.
     pub project_egress: Vec<String>,
+    /// The build cache key each `lns sandbox build` printed, in the order the scenario asked for them.
+    pub build_keys: Vec<String>,
+    /// How many instructions the service had booted a guest for when the last build started.
+    pub instructions_before_build: usize,
 }
 
 impl E2eWorld {
--- a/crates/e2e-tests/tests/steps/microvm.rs
+++ b/crates/e2e-tests/tests/steps/microvm.rs
@@ -1883,3 +1883,96 @@ fn built_reference_of_last_run(world: &mut E2eWorld) -> Result<String, String> {
             )
         })
 }
+
+/// Slice 4 of lensapp/lens-sandbox#393: `lns sandbox build` builds and publishes nothing, so what
+/// it did is read off its own two lines and off the instructions the service booted a guest for.
+#[when("the user builds the sandbox definition")]
+fn build_definition(world: &mut E2eWorld) {
+    world.instructions_before_build = instructions_run(world);
+    run_lns_microvm_as(world, &["sandbox", "build"], Vec::new());
+    if let Some(key) = printed_key(world) {
+        world.build_keys.push(key);
+    }
+}
+
+#[when(regex = r#"^the build context file "([^"]+)" is changed to "([^"]*)"$"#)]
+fn context_file_is_changed(world: &mut E2eWorld, path: String, content: String) {
+    let content = format!("{content}\n");
+    match world
+        .project_context_files
+        .iter_mut()
+        .find(|(existing, _)| *existing == path)
+    {
+        Some(entry) => entry.1 = content,
+        None => world.project_context_files.push((path, content)),
+    }
+}
+
+#[then("the build reports the key it is remembered under")]
+fn the_build_reports_a_key(world: &mut E2eWorld) -> Result<(), String> {
+    match world.build_keys.last() {
+        Some(key) if key.len() == "sha256:".len() + 64 => Ok(()),
+        other => Err(format!("the build printed no usable key: {other:?}")),
+    }
+}
+
+#[then(regex = r#"^the build reports the image it produced from "([^"]+)"$"#)]
+fn the_build_reports_the_image(world: &mut E2eWorld, label: String) -> Result<(), String> {
+    let run = world.result.as_ref().ok_or("no CLI run captured")?;
+    let line = run
+        .stdout
+        .lines()
+        .find(|line| line.starts_with(&format!("built {label} as ")))
+        .ok_or_else(|| format!("the build must name what it built:\n{}", run.stdout))?;
+    match line.contains("@sha256:") {
+        true => Ok(()),
+        false => Err(format!("the built image must be named by digest: {line}")),
+    }
+}
+
+#[then("the build reports the same key as the build before it")]
+fn the_same_key(world: &mut E2eWorld) -> Result<(), String> {
+    match world.build_keys.as_slice() {
+        [.., before, now] if before == now => Ok(()),
+        keys => Err(format!("expected the last two keys to agree, got {keys:?}")),
+    }
+}
+
+#[then("the build reports a different key from the build before it")]
+fn a_different_key(world: &mut E2eWorld) -> Result<(), String> {
+    match world.build_keys.as_slice() {
+        [.., before, now] if before != now => Ok(()),
+        keys => Err(format!(
+            "expected the last two keys to differ, got {keys:?}"
+        )),
+    }
+}
+
+#[then(regex = r"^the service booted a guest for (\d+) instructions? of that build$")]
+fn instructions_of_that_build(world: &mut E2eWorld, expected: usize) -> Result<(), String> {
+    let ran = instructions_run(world) - world.instructions_before_build;
+    match ran == expected {
+        true => Ok(()),
+        false => Err(format!(
+            "expected {expected} instruction(s) to run, {ran} did\n--- service.log ---\n{}",
+            crate::steps::service::read_service_log(world),
+        )),
+    }
+}
+
+/// Every `RUN` the service reaches a guest for says so once, so counting the lines counts the steps no key answered.
+fn instructions_run(world: &E2eWorld) -> usize {
+    crate::steps::service::read_service_log(world)
+        .lines()
+        .filter(|line| line.contains("Running") && line.contains("instruction"))
+        .count()
+}
+
+fn printed_key(world: &E2eWorld) -> Option<String> {
+    world
+        .result
+        .as_ref()?
+        .stdout
+        .lines()
+        .find_map(|line| line.strip_prefix("key ").map(|key| key.trim().to_string()))
+}
--- a/crates/lns-cli/src/sandbox/mod.rs
+++ b/crates/lns-cli/src/sandbox/mod.rs
@@ -1514,6 +1514,61 @@ mod tests {
         assert!(format!("{err:#}").contains("unexpected response"));
     }
 
+    const A_DOCUMENT: &str =
+        "apiVersion: lns.run/v1\nkind: sandbox\nname: agent\nspec:\n  image: ./image\n";
+
+    #[tokio::test]
+    async fn build_surfaces_the_services_refusal_and_rejects_an_unrelated_answer() {
+        let svc = CannedService::holding_a_document(
+            Response::Error {
+                message: "there is nothing to build".into(),
+            },
+            A_DOCUMENT,
+        );
+        let err = build(&svc, &build_args(), &mut Vec::new())
+            .await
+            .unwrap_err();
+        assert!(format!("{err:#}").contains("there is nothing to build"));
+
+        let svc = CannedService::holding_a_document(Response::Pong, A_DOCUMENT);
+        let err = build(&svc, &build_args(), &mut Vec::new())
+            .await
+            .unwrap_err();
+        assert!(format!("{err:#}").contains("unexpected response"));
+    }
+
+    /// A document nobody wrote is named, and no build reaches the service.
+    #[tokio::test]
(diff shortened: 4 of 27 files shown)


chenhunghan (commented):
Review of slice 4 of lens-sandbox issue #393. This PR auto-merged into `feat/containerfile-executor` minutes after it opened, before any review, because a stacked PR's base has no required checks; the review is posted here anyway and the fixes go to that branch. Two sources: the combined branch was rebuilt on an arm64 Mac and every Containerfile feature run (section 1), and a reading review of `713473aa..35ccea77` against the pinned sources (section 2). Verdict at the end.

# 1. What happened when the combined branch ran

```
a run whose spec.image is a Containerfile builds it and boots from the result   → 1 passed
a build carries no credential the document did not declare                     → 1 passed
a build reaches only what the document's egress allows                         → 2 passed
a build is remembered by its key, and only what changed is built again         → FAILED
```

Slice 3 has no regression. Slice 4's own scenario fails at `When the user builds the sandbox definition`, exit 1:

```
Error: building ./image/Dockerfile: line 2: RUN /bin/sh -c 'echo built-by-lns > /built-marker':
this definition reached the plan without being resolved; it still declares mixins (./project-egress.yaml)
```

That is finding 2.2 below, reproduced on real hardware.

# 2. What reading found

**2.1 `lns sandbox prune` drops every pulled image and sweeps the whole layer cache.** `image_store/mod.rs:362` partitions on `record.kind == RecordKind::Image && !named.contains(…)`, but `RecordKind::Image` is the kind every `lns pull` and every auto-pull writes (`image_store/mod.rs:112`), and `named` only ever holds `lns-build.local/built@…` references. Failure: on a machine with no Containerfile, `lns run docker.io/library/alpine:3.20`, stop it, `lns sandbox prune`; the alpine record, its manifest and its layers are gone. Fix: a `RecordKind::Built`, or a filter on the built repository prefix, and partition on that.

**2.2 `lns sandbox build` never resolves the document's mixins, so it refuses at the first `RUN`.** `real.rs:198` sends `mixins: Vec::new()` while the CLI's `build` sends the document verbatim and never `Request::ResolveDefinition`; `artifact/mod.rs:309` then refuses. Every document the e2e harness writes declares a mixin, and so does the mahler kit. Fix: the same preflight `run_image` takes (resolve, pass the resolved definition, its authored egress and packed filesets), or refuse a document with mixins by name until it does.

**2.3 A step is reused across documents, so a build inherits another document's egress and credentials.** `key.rs:44` keys a `RUN` on parent, argv, env, user and workdir; nothing about the policy it ran under enters it, and the step cache is machine-global (`real.rs:355`). Failure: document A with a credential and an egress rule builds `RUN npm install -g @acme/private`; document B on the same machine, with neither, and the same `FROM` and `RUN`, boots no guest and gets A's image. Fix: a fingerprint of the resolved policy (egress, declared credentials) in the step key, or scope the step cache to the document.

**2.4 The built-image sweep holds no cache lock and pins no run's layers.** `real.rs:249` passes an empty pin set and takes no lock, while `prune()` at `image_store/mod.rs:824` takes the exclusive lock and passes `recorded_run_pins()`. Failure: a build committing step 3 while another terminal prunes loses the blob it just installed. Fix: `cache_lock().write()` and `recorded_run_pins()`, as `prune()` does.

**2.5 Without `-f`, prune returns before the sweep can run.** `sandbox/mod.rs:799` returns on "No stopped sandboxes." and never sends `PruneRuns`, so orphaned built images stay forever on the interactive path, and the prompt names only sandboxes while the command now deletes images. Fix: ask the service what the sweep would drop, list it, and send `PruneRuns` when either list is non-empty.

**2.6 Two copies of one document share one entry, and deleting either drops the image the other needs.** `cache.rs:36,90` keys an entry's lifetime on one `source` path while the image key excludes the path by decision, and a hit does not refresh the entry. Fix: record every source a key answered for, or refresh on hit.

**2.7 `docs/cli-reference.md` describes neither the verb nor the wider prune**, and its prune row still says stopped sandboxes only. Fix: add the verb and amend the row.

Verified correct by reading: the image key takes exactly the base reference, the Containerfile text, the context hash and the architecture, and no document field reaches it; the length-prefixed digest keeps the three step-key shapes apart; the prefix-reuse rule is sound when a middle step changes because the parent reference is in every key.

# 3. Questions for the author

- The run log says the sandbox compiled `lns-service` against stub GTK libraries with unresolved symbols ignored; the PR body does not. Which of the listed test runs exercised the real binary? None of the listed commands is `make e2e` or `make e2e-microvm`.
- Where does the context stop? `spec.image: ./Containerfile` at a project root hashes `.git` and `node_modules` on every build. Is `.dockerignore` slice 5's, and until then should the key refuse a context above a size?
- The spec says an implementation MUST NOT reuse a build on the path a document sits at, while `cache.rs` keys an entry's lifetime on exactly that path. Which is intended?
- An interrupted `lns sandbox build` leaves the build guest running: build steps are not in the run registry and `BuildSandbox` has no cancel. Whose slice?
- The spec now reads "the instruction as it runs" where the Plan row reads "instruction text". The Plan row will be updated to match.

# 4. Verdict

The key, the prefix reuse and the verb's shape are right. Before this stack merges: 2.1 (a prune that deletes every pulled image), 2.2 (the verb does not work on any document with a mixin, reproduced here), 2.3 (cross-document reuse leaks policy), 2.4 and 2.5 (the sweep's lock and path), and the PR body must state the stub-build caveat. 2.6 and 2.7 can follow. After the round, the four features are re-run here.

PR #412, opened 2026-09-09
[reading only, merges via #422] feat(cli): lns push publishes the image a Containerfile builds, and lns pull fetches it

chenhunghan:

Part of #393 — slice 5, push and pull of a built image. Stacked on
[#408](https://github.com/lensapp/lens-sandbox/pull/408) (slices 3 and 4) and opened
against `feat/containerfile-executor`.

A document whose `spec.image` is a path publishes the image lns built, into the same
repository as the artifact, and the document a consumer receives names it by digest:

```
$ cat lns.yaml
apiVersion: lns.run/v1
kind: sandbox
name: agent
spec:
  image: ./image

$ lns push ghcr.io/team/agent:1.4.0
key sha256:5f0b1a2c3d4e5f60718293a4b5c6d7e8f90a1b2c3d4e5f60718293a4b5c6d7e8
built ./image/Containerfile as ghcr.io/team/agent@sha256:761743bb… (2 layers)
packed ./image/Containerfile -> sha256:c41d0f9b… (612 bytes)
built and pushed ghcr.io/team/agent:1.4.0@sha256:9ab3c02e…

$ lns push --dry-run ghcr.io/team/agent:1.4.0
key sha256:5f0b1a2c3d4e5f60718293a4b5c6d7e8f90a1b2c3d4e5f60718293a4b5c6d7e8
would publish the image ./image/Containerfile builds to, sha256:761743bb…
would pack ./image/Containerfile -> sha256:c41d0f9b… (612 bytes)
would push ghcr.io/team/agent:1.4.0@sha256:9ab3c02e… (1704 bytes)
dry run — built and validated; nothing uploaded
```

A dry run on a machine that has never built it says so rather than promising a digest:

```
$ lns push --dry-run ghcr.io/team/agent:1.4.0
key sha256:5f0b1a2c3d4e5f60718293a4b5c6d7e8f90a1b2c3d4e5f60718293a4b5c6d7e8
note: ./image/Containerfile has not been built on this machine, so its image digest can
only be known by building; the published digest will differ from this preview
```

And a machine that pulls it reads the build off the artifact, without building or
running anything:

```
kind: sandbox
reference: ghcr.io/team/agent:1.4.0
digest: sha256:9ab3c02e…
image: ghcr.io/team/agent@sha256:761743bb…
imageSource: built from ./image/Containerfile (2 lines, context 2 files)
context: Containerfile (37 B)
context: app/main.js (15 B)

./image/Containerfile:
  FROM docker.io/library/node:24-bookworm
  RUN npm install -g @anthropic-ai/claude-code
```

## The slice 5 row of the Plan table, mapped to code and tests

|---|---|---|
| `push` builds when the cache is cold | `distribute::publish_the_image` asks the `ImageBuilder` port; `ServiceImageBuilder` (`artifact/real.rs`) sends `Request::BuildImageForPush`, which the service answers through slice 4's `executor::build` | Layer 2 `image_push.feature` "a push of a path document builds the image and publishes it beside the artifact" and "a push whose image this machine already built runs nothing and says so"; Layer 3 `distribute.rs` `a_build_the_service_refuses_stops_the_push_with_what_it_said` |
| publishes the image manifest, its config and its layers into the same repository as the artifact, through the OCI client lns already uses, so one grant covers both | `build::push::push_image` — the same `oci_client` client, the same `lns login` credential and the same auth/error mapping `push_artifact` uses; the repository is `mixin_plan::repository_of(<REF>)` | Layer 2 "the image was published into `ghcr.io/team/hermes`"; Layer 3 `a_registry_that_refuses_the_image_names_the_repository_it_was_publishing_into` |
| the published document's `spec.image` is rewritten to the image's digest reference and `spec.imageSource` keeps the path the author wrote | `lns_artifact::image::rewrite_to_built`, and `image::validate_source` holds the pair to the grammar: a path form, and never beside a path-form `image` | Layer 3 `image.rs` 6 tests; `sandbox.rs` `a_published_document_carries_the_digest_it_built_to_and_the_path_it_was_built_from`, `a_document_that_says_both_build_this_and_built_from_that_refuses`; Layer 2 "the published document names the image by digest and keeps the path the author wrote" |
| the Containerfile and its context are packed as a layer of the artifact, the way a `README.md` beside the document is packed today, so `lns inspect` and the hub render them | new `lns_artifact::build_source` (`pack`/`read`, media type `application/vnd.lns.image.source.v1.tar+gzip`, title annotation), `build_artifact_with`, and `walk::walk_context` — a fileset walk with §3.1.1's one difference, a symlink is left behind rather than refusing the push | `build_source.rs` 7 tests; `build.rs` `a_built_image_s_containerfile_and_context_become_a_layer_of_the_same_artifact`; `walk.rs` `a_symlink_in_a_build_context_is_left_behind_rather_than_refusing_the_push`, `a_build_context_still_refuses_a_secret_shaped_file`; Layer 2 "the Containerfile and its context ship as a layer of the same artifact" |
| `push --dry-run` prints the build key and, when cached, the image digest, and builds nothing; it says when a digest can only be known by building | `executor::plan` (new, sharing `executor::open` with `build`) → `real::build_image_for_push`'s `plan_only` arm → `distribute::preview_the_image` | `executor.rs` `a_plan_answers_with_the_key_and_builds_nothing`, `a_plan_of_a_build_this_machine_already_made_names_its_image`, `a_plan_that_ignores_the_cache_names_no_image_although_one_is_held`, `a_plan_of_a_file_with_no_from_is_refused_the_way_a_build_is`; Layer 2 two `--dry-run` scenarios |
| `push --rebuild` ignores the cache for this push | `PushArgs.rebuild` → `PushPorts.rebuild` → `ImageRequest.rebuild` → `BuildPlan.rebuild` | Layer 2 "push --rebuild ignores the cache for this push" |
| a built image over a configurable limit, default 4 GiB, fails the push and names the layer that grew it, the way a fileset over its limit fails validate | `lns_artifact::image::refuse_an_image_over` names the heaviest layer and reads the instruction that wrote it out of the image config's history; the limit is `push.imageLimit` (MiB) via `config::load_image_limit` | `image.rs` 5 tests; `config/mod.rs` 4 tests; Layer 2 "a built image over the limit fails the push and names the layer that grew it" |
| `pull` fetches the document and the built image for this host's architecture | **no new code** — see [deviations](#deviations) | Layer 1 `microvm_containerfile_push.feature`; Layer 2 "the published document's `spec.image` is the digest of the published image", which is the property the existing dependency pull turns into a fetch |
| a run from the pulled document finds the layers without any pre-start script | the built image is an ordinary digest reference by the time a consumer has it, so the ordinary boot path finds it | Layer 1 `microvm_containerfile_push.feature`, last three steps |
| `inspect` on a pulled artifact prints the Dockerfile and lists the context's files with sizes, then the built digest, without running anything | `artifact/real.rs::fetch_build_source` pulls the one layer; `inspect::build_source_layer` / `read_build_source` decode it; `SandboxView.image_source` carries it; the CLI's `render_build_source` prints it | `inspect.rs` 7 tests; `build_source.rs` `read`; Layer 2 `image_pulled_disclosure.feature`, 2 scenarios; Layer 1 the inspect half of the @microvm scenario |
| the invariant: `lns push` never publishes a document whose image it does not have | the image is built and uploaded before the artifact is built, and a build that answers with no image stops the push | Layer 3 `a_build_that_answers_with_no_image_stops_the_push_before_it_publishes_anything`, which also asserts nothing was uploaded |

## The review round

[The reading review](https://github.com/lensapp/lens-sandbox/pull/412#pullrequestreview-5159540041)
raised thirteen items. Twelve are fixed here, one commit each; item 8 is left
out and named below. The base branch moved first — the slice 4 round landed on
`feat/containerfile-executor` — so this branch merges it before anything else.

| # | What the review found | Commit |
|---|---|---|
| — | the base branch moved: a build is now keyed by the policy it runs under | `merge: feat/containerfile-executor into feat/containerfile-push` |
| 1 | `lns push` handed the service an unresolved document, so any document with `spec.mixins` failed at the first `RUN` | `fix(cli): a push resolves the document before it builds the image` |
| 2 | `lns inspect` pulled the build source layer with no pre-download size limit | `fix(service): inspect refuses a build source layer by its declared size` |
| 3 | `push --dry-run` pulled the entire base image | `fix(service): a dry run reads its base rather than downloading it` |
| 4 | the @microvm push step omitted `--yes` on a document that always declares a mixin | `test(e2e): the push scenario publishes a document a push can answer` |
| 5 | `lns validate` and the build gated different subsets, because the merge kept two parsers | `refactor: one Containerfile parser, and the flag policy it holds both callers to` |
| 6 | a build source layer that does not decode failed the whole `lns inspect`, and the fallback name was wrong for half the projects | `fix: a build source an approver cannot read costs them the layer, not the verb` (the name, in item 9's commit) |
| 7 | `push --dry-run` did not apply `push.imageLimit` | `fix(cli): a dry run is held to the size limit the push is` |
| 8 | the build context is packed under fileset rules and refused in fileset words | **left out** — see below |
| 9 | `spec.imageSource` and the layer annotation recorded different strings | `fix(cli): one record for where an image was built from`, and the spec decision behind it in `docs(spec): the title annotation is the recorded path, and the tar names the file` |
| 10 | a pulled artifact's inspect printed the Containerfile in the middle of the render | `fix: a build source an approver cannot read costs them the layer, not the verb` |
| 11 | `docs/cli-reference.md` described neither `--rebuild`, nor `push.imageLimit`, nor that `--dry-run` is no longer offline | `docs(cli): the reference carries what push does with a Containerfile` |
| 12 | the coverage exemptions grew new decisions their reasons did not name | `refactor: the decisions the exempt files carried move where they are tested` |
| 13 | a fourteen-space gap in the manifest-mismatch refusal, and doc comments running to two lines | `style: the manifest refusal reads as one sentence, and each new note is one line` |

Two commits after those are the round's own repairs, both found by CI rather
than by reading: `test(service): a peeked base is one this host would run,
whichever host that is` (the peek fixture declared arm64, so the architecture
refusal fired on an x86_64 runner and the test inverted on the host), and
`test: every line the review round added is exercised`.

### What each fix does

**1 — the push takes the preflight a run takes.** The `ImageBuilder` port gains
`resolve`, and `publish_the_image` / `preview_the_image` call it before the
build when the document declares a mixin. The seam the slice 4 round added for
`lns sandbox build` moves to `crate::resolve`, so both verbs read one function,
and `Request::BuildImageForPush` carries `authored_egress` and
`packed_filesets` exactly as `BuildSandbox` does. A document declaring no mixin
is built with no round trip, as a run of one boots with none.

```
Scenario: a push of a document that declares a mixin resolves it before the build
Scenario: a push of a document that declares no mixin is built without a resolution
```

**2 — a read-only verb fetches no unbounded bytes.** `build_source_layer`
refuses a declared size over `MAX_FILESET_BYTES` and the `pull_blob` never
starts, the way `validate_packed_layer_size` refuses a fileset layer. The
decision sits in `artifact/inspect.rs`, which the gate holds at 100%, not in
the exempt wiring that calls it.

**3 — a dry run reads its base.** `executor::open` takes a `BaseNeed`: a build
still ingests, a plan peeks. `image::peek_base_with` pulls one manifest and one
config, verifies the digest pin and applies the same architecture refusal a pull
applies, and fetches no layer. The key is taken over the same digest-pinned
spelling an ingest resolves to, so a plan and the build after it agree.

```
a plan reads its base without ingesting it            → calls: ["peek alpine"]
a plan of a base no registry answers for is refused where a build would be
```

**5 — one parser, one flag policy.** `crates/lns-service/src/containerfile/parse/`
is deleted and the executor consumes `lns_artifact::containerfile`. The flag
policy — `RUN --network`, `RUN --security`, `COPY --link`, `ADD --link`, and a
`--chmod` that is not octal — is now a table in `lns_artifact::containerfile`
that the parse itself holds every caller to, so what `lns validate` refuses is
what a build refuses. `Transfer` keeps its flags as written and reads back the
two a build honours:

```rust
transfer.owner()   // Some("node:node")
transfer.mode()    // Some(0o750) — the parse refused every non-octal spelling
```

The IGNORES entry the deleted `upstream.rs` carried goes with it, as does the
service's `parse-dockerfile` dependency.

**9 — one record for the source path.** The layer's title annotation is now
exactly `spec.imageSource`, which is the path the author wrote. Which entry of
the tar holds the instructions is read out of the tar: the name the recorded
path ends in where the author named a file, and otherwise `Containerfile`, then
`Dockerfile`, in the order a build searches a context. That is a change to
§7.3, which showed a title naming a file while `imageSource` records a path
that may be a directory — the two could not both hold, so the section is
changed as a decision, in its own commit.

The record also goes with the layer it describes: a pulled document re-pushed
as it stands ships no build source, so it publishes no `imageSource`.

```
Scenario: a pulled document re-pushed as it stands records no source it does not carry
```

### What was left out

**Item 8 — the context's own limits and wording.** The build context is packed
under `walk_context`, which inherits `MAX_FILESET_BYTES`, `MAX_FILESET_ENTRIES`
and the secret-name refusal in fileset words, while the service's own context
walk (`key.rs`) has no limit at all — so `lns run` builds what `lns push`
refuses. The review is right that these are two different questions and the
context deserves its own answer. Deciding it means deciding what a build
context may weigh, how many entries it may hold, whether the build path gets
the same ceiling, and what `.dockerignore` does to all three — the slice 4
review asks the last of those as an open question too. That is a decision, not
a repair, and it is not made here. The divergence is unchanged by this round:
nothing in it widened or narrowed either walk.

## Answers to the review's questions

**The PR base.** It is `feat/containerfile-executor`, and it was pointed at
`main` for one run in this session and pointed straight back; the timeline now
carries both events. It is confirmed above and in the CI section below.

**Where slice 6 finds the other architecture's manifest.** Today a push writes
one image manifest to `{repository}:sha256-<digest>` — `published_tag` — and the
document names it by digest. Nothing writes an index and nothing reads the
registry back, so a second push from another architecture replaces what the
first published and slice 6 has nowhere to look. That gap is real and this round
does not close it, because closing it changes what a push publishes.

What it should be, so slice 6 has an answer: **a push writes a per-architecture
tag it can read back, and the document names an index.**

1. Each host's push writes its own manifest under
   `{repository}:sha256-<digest>-<os>-<arch>` — a stable name derived from the
   two things the manifest already declares, so a second push from the same
   architecture overwrites only its own entry.
2. The push then reads the repository's existing entries for the key it built,
   assembles an OCI image index over every architecture it finds plus its own,
   and pushes that index under `{repository}:sha256-<index-digest>`.
3. `spec.image` names the index's digest, not the manifest's. A consumer's
   ordinary pull already selects by platform, so nothing on the run side
   changes — which is the same property that made `pull` need no code in this
   slice.

That makes the registry the place slice 6 looks, rather than a second local
store, and it keeps the invariant this slice states: a push never publishes a
document whose image it does not have, because the index it names is assembled
from manifests it has just confirmed are there. `PushableImage` needs the
platform to do it, and does not carry one today; that field is slice 6's first
commit.

**What the 4 GiB limit measures.** The review is right that the layers a build
commits are uncompressed tars, so "the compressed bytes a push uploads" is
wrong. The limit is measured over the manifest's declared layer sizes, which
for a built image are the bytes on this machine and the bytes the registry
receives — an accurate count of what the push costs and what a consumer
downloads, and not a count of what the image occupies unpacked in a guest. The
wording in the deviations below is corrected to say so. Compressing a committed
layer is worth doing and is not this slice's; streaming it rather than reading
it whole is item 2.7 of slice 1, still open.

**Which registries this has been pushed to.** The e2e registry
(`registry:2` under `make e2e-microvm`) and nothing else. The Docker-schema2
case the review names — appending an OCI layer type under a base's Docker
`mediaType` — is untested against a real Docker Hub or GHCR repository, and the
master's Mac run is the first place it will be. If it is refused there, the fix
is to convert the base's media types to their OCI equivalents at assembly, in
`containerfile::image::assemble`.

**Whether the retry hint should name the published image tag.** Yes. The image
is uploaded before the artifact, so a `push_built` failure leaves the image in
the repository and the hint names only the mixins. That is a real gap and it is
not fixed here — it is one sentence in `distribute::push`'s error context, and
it belongs with the garbage-collection question the "what is left out" section
already carries (a published image under a digest-derived tag stays in the
repository when the document that named it is re-pushed). Both are the same
decision: who removes a published image nothing names.

**Whether `--rebuild` on a reference-form `spec.image` should be refused.** Yes
— a flag that silently does nothing is worse than a flag that says it does not
apply. It is not refused here, because the same is true of `lns sandbox build
--rebuild`, and refusing in one verb and not the other is a third behaviour. It
is one check in `PushArgs` and one in `SandboxBuildArgs`, and it needs a
decision on whether `lns push --rebuild` on a document that declares no
Containerfile is an error or a warning.

## The merge of the parser branch, and what it resolved

The first commit of this branch is `git merge origin/feat/containerfile-parser`, which
brings slice 0 ([#394](https://github.com/lensapp/lens-sandbox/pull/394), the grammar's
path form) and slice 2 ([#407](https://github.com/lensapp/lens-sandbox/pull/407), the
parser and the subset gate in `lns-artifact`) onto the executor's branch.

**One conflict, in `docs/sandbox-spec.md` §3.1.1.** Slice 2 wrote the two forms of
`image`, the build context and the instruction subset; slice 4 wrote what a build is
keyed by. Both are kept: the form first, the key after it. Nothing was dropped.

**The executor's private parser copy is gone.** #408 said the swap would be "delete
`containerfile/parse/`, and `use lns_artifact::containerfile::{…}`". The first round of
this branch deferred it, because the two types had diverged: `lns_artifact`'s
`Transfer` carried `flags: Vec<Flag>` as written while the service's carried a parsed
`owner` and octal `mode`, and swapping meant deciding where the subset's flag policy
lives. The review made that decision necessary — the two parsers gated two different
subsets — so the swap is done in this round and the policy lives in
`lns_artifact::containerfile`. See item 5 above.

**Two refusals this branch lifts, both from #394.** `lns run` no longer refuses a
path-form `spec.image` (`run/target.rs`) — slice 3 builds it, and the merge would
otherwise have re-introduced a refusal for behaviour that already ships on this base.
`lns push` no longer refuses it either, which is this slice. The two scenarios in
`image_containerfile.feature` that asserted those refusals are deleted; what replaces
them is `image_push.feature` here and the @microvm scenarios of slices 3 and 5.

cloud-hypervisor/virtiofsd. The feature is written to be right by reasoning and **is
unverified by me**. One command on an arm64 Mac:

```
LNS_E2E_FEATURE="a document built from a Containerfile publishes its image and pulls back whole" make e2e-microvm
```

`crates/e2e-tests/features/microvm_containerfile_push.feature` pushes a path-form
document to the e2e registry, then **stops the service that built it and starts a
second one in a home that has never built anything** — so what the pull, the inspect
and the run then find can only have come off the registry. It asserts the image landed
in the artifact's own repository, that the published document names it by digest, that
inspect prints the Containerfile and the context and boots nothing, and that the run
reads the file the `RUN` wrote and the file the `COPY` brought out of a document that
declares no `pre-start` script.

The push step of that scenario changed in this round (review item 4): it now writes a
mixin-free publisher document beside a copy of the build context and pushes it with
`--yes`, as the declarative and tools publisher steps beside it already do. The project
document always declares a mixin, and with no terminal the push bailed before it built
anything — so the feature could not reach its first assertion as written.

The slice 3 and slice 4 scenarios are unchanged.

## Deviations

- **The image is built by the service and uploaded by the CLI.** The service owns the
  executor, the build guest and the layer store; the CLI owns the `lns login`
  credential and every existing registry upload. So `Response::ImageBuiltForPush`
  carries the manifest and the config as text and each layer as **the path this machine
  holds it at**, and the CLI reads those files and pushes them. An image is gigabytes
  and a socket is not the place for them; the alternative — teaching the service to
  push — would put the push credential in a second place.
- **The manifest is re-serialized and held to the digest the build recorded.** The
  manifest cache stores the parsed manifest, and a registry addresses bytes, so
  `image::manifest_bytes` re-serializes and refuses if the result no longer hashes to
  the digest the build named. That is a real failure mode of storing a parse rather
  than the bytes, and it stops the push rather than publishing a manifest under a
  digest that is not its own.
- **`pull` needed no code at all.** The published `spec.image` is an ordinary
  digest reference in the artifact's repository, and `image_store::pull` already
  fetches `spec.image` as the sandbox's dependency. That is the point of the
  rewrite: a consumer's machine cannot tell a built image from any other. The row is
  therefore covered by the @microvm scenario and by the Layer 2 assertion that the
  published document carries a digest, not by new pull code.
- **The size limit is `push.imageLimit`, in MiB, a new `lns config` key.** #393 says
  "a configurable limit (default 4 GiB)" and names no surface. `lns config` is where
  this machine's defaults live, and `run.mem` already takes MiB, so the key follows it.
  `docs/cli-spec.md` §3.5 gains the row.
- **The limit is measured over the manifest's declared layer sizes**, which for an
  image lns built are uncompressed tars — the bytes this machine holds, the bytes the
  push uploads and the bytes a consumer downloads. It is not a count of what the image
  occupies unpacked in a guest. The refusal names the heaviest layer, not the one that
  crossed the line: on a sorted walk those differ, and the heaviest is the one an
  author can act on.
- **The build source layer is packed by the CLI from the working directory**, not from
  what the service built. The context the build read and the context the push packs are
  the same directory, read twice; keying them together would need the service to hand
  the bytes back, which is the streaming problem slice 3 already names.
- **A symlink in the context is listed by `inspect` on a local document and absent from
  a pulled one.** §3.1.1 says a symlink is not sent, so it is not in the layer, so the
  artifact cannot disclose it. The local render still names it.
- **`inspect` of a published artifact fetches one more blob.** Reading the
  Containerfile off the artifact means pulling the build source layer. It is refused by
  its declared size before the fetch starts and bounded again by `MAX_FILESET_BYTES` on
  the way out of the tar (review item 2), and it is a blob fetch, not a boot.
- **`containerfile/real.rs` keeps its coverage exemption** and grew:
  `build_image_for_push`, `pushable`, `plan` and `peek_base` are wiring over the
  registry, the manifest cache and the layer cache. Every decision inside them that is
  not wiring now lives where the gate holds it at 100% — the manifest-bytes check and
  the projection of a cached manifest into a pushable image in `containerfile/image.rs`,
  the plan and the `BaseNeed` split in `containerfile/executor.rs`, the peek itself in
  `image/mod.rs` — and the IGNORES reason names them (review item 12).

## What is left out

- **No architecture index** — slice 6. A second push from another architecture replaces
  the image rather than adding an entry to an index, and `inspect` prints one digest.
- **No host-Docker switch** — slice 7.
- **The hub does not render the image family yet** (lensapp/lens-artifact-hub#135). The
  layer and its media type are what it will read; the push does not wait for it.
- **A push does not garbage-collect what it published**, and a failed artifact upload
  does not name the image it already published. An image under a digest-derived tag
  stays in the repository when the document that named it is re-pushed, and the retry
  hint on a `push_built` failure names only the mixins. `lns sandbox prune` sweeps the
  local side (slice 4); the registry side is one decision — who removes a published
  image nothing names — and it is answered in the questions section above.

- **`--rebuild` on a reference-form `spec.image` is still a silent no-op**, in `push`
  and in `lns sandbox build` alike. It should be refused by name; see the questions
  section.

- **The build context is still walked under fileset rules on the push path and under
  no limit on the run path** — review item 8, left out with its reason above.
- **`--dry-run` still previews an artifact digest that the real push may not produce**
  when the image has not been built. It says so, in the same shape the tools note uses.
- **Streaming.** A captured layer is still held in memory twice (slice 1's item 2.7),
  and now a published layer is read into memory once more on its way to the registry.

`lns-service` does not link in this container (no libgtk-3/libatk/libxdo on the image),
so the harness was made to compile and link against stub pkg-config files and stub
shared objects with `--unresolved-symbols=ignore-all`. **Nothing in the repository
changed for it**, and no listed run exercises the real `lns-service` binary — every
test below is in-process (Layer 2 and Layer 3). `make e2e` and `make e2e-microvm` are
not run here; CI's `test` job and the master's Mac are the oracles for those.

| Command | Result |
|---|---|
| `cargo test -p lns-artifact` | 454 passed |
| `cargo test -p lns-ipc` | 181 passed |
| `cargo test -p lns-cli` | 999 unit tests; cucumber 578 scenarios / 2901 steps, all green |
| `cargo test -p lns-service` | 2655 unit tests; cucumber 259 scenarios / 1032 steps, all green |
| `cargo build -p e2e-tests --tests` | builds (the @microvm scenario does not run here) |
| `cargo clippy --all-targets -- -D warnings` on every crate touched | clean |
| `cargo clippy -- -D clippy::cognitive_complexity`, per crate, on every crate touched | clean |
| `make coverage COVERAGE_CRATES="lns-artifact lns-service lns-cli lns-ipc"` | `lns-artifact` 17 at 100%, 2 ignored; `lns-service` 126 at 100%, 35 ignored; `lns-cli` 44 at 100%, 19 ignored; `lns-ipc` 7 at 100% — **0 failures** |

`lns-artifact/src/build_source.rs` and `lns-cli/src/resolve.rs` go in at 100% and are
not in IGNORES. One entry left the table this round —
`lns-service/src/containerfile/parse/upstream.rs`, whose file is deleted — and none was
added.

## Red first

Every fix in the round that a test can express was watched fail first:

- item 1 — `image_push.feature`'s two new scenarios, red on
  `a build step is the document's own run, so it never reaches the plan unresolved`
  with the resolve call removed;
- item 2 — `build_source_layer`'s two new tests, red on `no method named expect_err`
  before the signature returned a `Result`;
- item 3 — `a plan reads its base without ingesting it`, red on
  `left: [Base("alpine")] right: [Peek("alpine")]`;
- item 5 — five refusal tests in `lns_artifact::containerfile`, red before the flag
  table existed, and two accessor tests red on `no method named owner`;
- item 7 — `a dry run of an image over the limit refuses it, as the push would`, red on
  `expected a non-zero exit code`;
- item 9 — the pack/read tests against a recorded directory, and
  `a republished document that ships no build source records none`, red on
  `cannot find function forget_source`;
- item 10 — `the output prints "RUN …" after "mount: bind . -> /workspace"`, red while
  the render printed the instructions before the mounts.

Items 11, 12 and 13 are documentation, a move of already-tested code, and whitespace;
each carries its tests where the code now lives.

## CI

**The base is `feat/containerfile-executor`.** It was pointed at `main` for one run in
this session and pointed straight back; both events are on the timeline. As the review
predicted, the run produced nothing — the workflow triggers only for pull requests based
on `main`, and a pull request that conflicts with `main` gets no run at all.

So the oracle is the same shape the first round used: this branch **merged with
`main`** on `ci-oracle/containerfile-push-2`, opened as a draft against `main` for one
run, now closed with its branch deleted. The run is
[actions/runs/34419131843](https://github.com/lensapp/lens-sandbox/actions/runs/34419131843),
`shell-tests-macos`, both `check-release-build`s, `changes`, `labels` and `check`.

Two earlier attempts on that branch were red, and both are fixed on this branch:
`cargo fmt` on the merge's own export list, and
`peeking_a_base_reads_its_manifest_and_config_and_fetches_no_layer`, whose fixture
declared arm64 and inverted on the x86_64 runner — the environment-parity rule the
repository states, caught by CI rather than by reading.

The merge with `main` was three conflicts, all in files `main` also moved: the
`lns-ipc` export list (a union), one `lns config` test `main` made serial, and
`handle_request`, which `main` split into `answered_whole`. None of those resolutions
are on this branch; they are the merge `feat/containerfile-executor` will have to do
when it catches up.

Auto-merge is not armed: the base branch is itself an open pull request and carries no
required checks.

**The @microvm scenario is still unverified by me** — no `/dev/kvm`, no Vz, no
cloud-hypervisor/virtiofsd where I work. The master runs it on the Mac after this round.


chenhunghan (commented):
Reading review of slice 5 of lens-sandbox issue #393, against `feat/containerfile-executor` (the branch also merges `feat/containerfile-parser`, judged separately). Nothing was run: the host verification of this branch waits on disk and follows once the slice 4 round on the base branch lands; every claim below was verified by reading this branch and the pinned sources. Verdict at the end.

# Findings, ranked

**1. `lns push` hands the service an unresolved document, so any document that declares a mixin fails at the first `RUN`.** `distribute.rs:511-519` passes `doc` as read off disk, before `pin_local_mixins` at `:526`; `real.rs:177` sends it verbatim as `BuildImageForPush`; `containerfile/real.rs:334` sends `mixins: Vec::new()` and never resolves. This is the same defect the slice 4 review found in `lns sandbox build` (2.2 there), reproduced on the Mac for the build verb: `this definition reached the plan without being resolved; it still declares mixins`. Failure: `lns push --yes ghcr.io/team/agent:1.4.0` on any document with `spec.mixins`, which is every document the e2e harness and the mahler kit write. Fix: resolve before the build, the same preflight `run_image` takes; the slice 4 round on the base branch is fixing the verb, so share that seam.

**2. `lns inspect` downloads the build source layer with no pre-download size limit.** `artifact/real.rs:448-451` calls `pull_blob` on whatever the manifest declares into a `Vec<u8>`; the `MAX_FILESET_BYTES` ceiling in `build_source::read_within` applies only after the blob is in memory, while every other layer this codebase fetches off a stranger's artifact is refused by declared size first (`fileset.rs:449-455`). Failure: `lns inspect ghcr.io/attacker/agent:1`, the read-only verb, downloads a declared 20 GiB layer into the service. Fix: `validate_packed_layer_size` before `pull_blob`.

**3. `push --dry-run` pulls the entire base image.** `executor::plan` calls `open`, which resolves `FROM` through `resolve_base` at `containerfile/real.rs:433-438`, a full ingest that downloads every base layer. The body and §6 say the dry run builds nothing and stays offline. Failure: a first `--dry-run` against `FROM node:24-bookworm` downloads about 400 MiB and then prints that nothing has been built. Fix: a resolve-only path that peeks the base manifest for its digest and config.

**4. The @microvm scenario's push step omits `--yes` on a document that always declares a mixin.** `steps/microvm.rs:1990` pushes `microvm_project`, which writes `spec.mixins`; with no terminal `mixin_plan.rs:256-259` bails, and interactively it blocks on `read_answer()`. The neighbouring push steps at `:745` and `:1064` write a mixin-free publisher document. Failure: the feature fails at `When the user pushes …` before finding 1 is even reached. Fix: `--yes` and a mixin-free publisher document.

**5. `lns validate` and the build gate different Containerfile subsets, because the merge kept two parsers.** The service copy at `parse/mod.rs:296` refuses `COPY --link`, `ADD --link`, `RUN --network`, `RUN --security` and a non-octal `--chmod`; `lns_artifact::containerfile::flags` at `:384-388` carries flags as data and refuses none, and `image_problems` never looks at them. §3.1.1 as merged says offline validation MUST refuse what the build refuses. Failure: `COPY --link app /srv/app` validates and then fails inside the build. Fix: the parser swap this PR deferred, with the flag policy in `lns_artifact::containerfile` read by both.

**6. A build source layer that does not decode fails the whole `lns inspect`, and the fallback name is wrong for half the projects.** `artifact/real.rs:452` propagates `read_build_source`'s error out of `inspect`; `inspect.rs:105` falls back to `Containerfile` while this PR's own scenario writes `Dockerfile`. Failure: an artifact whose descriptor lost its title annotation makes `inspect` exit non-zero and print nothing, so the consumer cannot approve the run. Fix: warn and return `Ok(None)`, and try both names.

**7. `push --dry-run` does not apply `push.imageLimit`.** `distribute.rs:406` enforces it on `push` only; `push_dry_run` at `:453-461` takes no limit. Failure: the preview promises a publish the push then refuses. Fix: pass the limit and call `refuse_an_image_over` in the preview.

**8. The build context is packed under fileset rules and refused in fileset words.** `image_build.rs:198` uses `walk_context`, which inherits `MAX_FILESET_BYTES`, `MAX_FILESET_ENTRIES` and the secret-name refusal with "fileset" wording, while the service's own context walk (`key.rs:134`) has no limit, so `lns run` builds what `lns push` refuses. Fix: the context's own limits and wording, decided as a context question.

**9. `spec.imageSource` is written and validated but never read, and the layer annotation records a different string.** `rewrite_to_built` writes `./image`; `build.rs:155` annotates `./image/Containerfile`; nothing reads `image_source`, and §7.3 says the annotation is the path `imageSource` records. Failure: a re-push of a pulled document keeps `imageSource` and packs no source layer. Fix: pick one record.

**10. A pulled artifact's inspect prints the Containerfile in the middle of the render.** `artifact/mod.rs:690-693` calls `render_build_source` before mounts, credentials and scripts, while the local render calls it last. Fix: last, as locally.

**11. `docs/cli-reference.md` does not describe `--rebuild`, `push.imageLimit`, or that `--dry-run` is no longer offline for a path image.** Fix per AGENTS.md rule 3.

**12. The coverage exemptions grew new decisions** (the publish tag in `build/push.rs:57`, blob order, `pushable`'s defaults at `containerfile/real.rs:246-282`) without their reasons naming them. Fix: update the reasons or move the decisions into tested modules.

**13. The manifest-mismatch refusal at `containerfile/image.rs:50` carries a fourteen-space gap** from a dropped continuation, and several new doc comments run to two lines.

Confirmed by reading: the merge of the parser branch resolved `docs/sandbox-spec.md` and `scripts/coverage-floor.sh` by union and dropped nothing; the rewrite to the digest reference and `imageSource` are written where §6 says; the size limit is enforced on the push path.

# Questions for the author

- **The PR base is `main` and GitHub reports it conflicting.** The body says it was pointed at main for one run and back; the timeline has no base change event. As it stands, merging would land #394, #407, #396, #408, #410 and this slice into main in one PR. The master will point it back at `feat/containerfile-executor` if the run has not.
- What does slice 6 read the other architecture's manifest from? The publish pushes one manifest under `{repository}:sha256-<digest>` and the document names it by digest; `PushableImage` carries no platform and nothing reads the registry back. Slice 6 needs a place to find the first host's manifest.
- Layers are uncompressed tars, so "the compressed bytes a push uploads" is not what the 4 GiB default measures, and `push_image` reads each layer whole into a `Vec<u8>`. Stream, or say what the limit measures?
- For a Docker-schema2 base, `assemble` appends an OCI layer type under the base's `mediaType`. Which registries has this been pushed to?
- The image is uploaded before the artifact; if `push_built` fails, the retry hint names only mixins. Should it name the published image tag?
- `--rebuild` on a reference-form `spec.image` is a silent no-op. Refuse by name?

# Verdict

The shape is right: build when cold through slice 4's key, publish beside the artifact, rewrite to the digest, keep the source. Before merge: 1 (shared with the slice 4 round), 2 (a read-only verb must not fetch unbounded bytes), 3, 4 (the scenario cannot pass as written), 5 (the spec promise), 7 and 9. Can follow: 6, 8, 10 to 13. The base must return to `feat/containerfile-executor` before anything else. The @microvm scenario is run on the Mac after the round.

chenhunghan:

PR #415, opened 2026-09-10
[reading only, merges via #422] feat: a push from a second architecture adds an entry to the image index

chenhunghan:

Part of #393 — slice 6, the per-architecture image index. Stacked on
[#412](https://github.com/lensapp/lens-sandbox/pull/412) (slice 5, push and pull) and
opened against `feat/containerfile-push`.

A build runs on the architecture of the host that pushes, so a team on mixed
hardware pushes one document from two hosts. Slice 5 let the second push replace
what the first published. Now it adds to it:

```
$ lns push ghcr.io/team/agent:1.4.0                 # on an arm64 Mac
key sha256:5f0b1a2c…
built ./image/Containerfile as ghcr.io/team/agent@sha256:761743bb… (2 layers)
index ghcr.io/team/agent@sha256:aa41f0c9… holds linux/arm64 sha256:761743bb…

$ lns push ghcr.io/team/agent:1.4.0                 # then from amd64 CI
key sha256:9c2ab61e…
built ./image/Containerfile as ghcr.io/team/agent@sha256:9ab3c02e… (2 layers)
index ghcr.io/team/agent@sha256:b17d3f42… holds linux/amd64 sha256:9ab3c02e…, linux/arm64 sha256:761743bb…

$ lns push ghcr.io/team/agent:1.4.0                 # from the Mac again, nothing changed
key sha256:5f0b1a2c…
nothing to publish: the index already holds linux/arm64 sha256:761743bb… for ./image/Containerfile
index ghcr.io/team/agent@sha256:b17d3f42… holds linux/amd64 sha256:9ab3c02e…, linux/arm64 sha256:761743bb…
```

A dry run says what the index holds and what this push would add:

```
$ lns push --dry-run ghcr.io/team/agent:1.4.0
key sha256:5f0b1a2c…
the index holds linux/amd64 sha256:9ab3c02e…
would publish the image ./image/Containerfile builds to, sha256:761743bb…
this push would add linux/arm64 sha256:761743bb…
…
dry run — built and validated; nothing uploaded
```

`lns inspect` prints one built digest per architecture, and a host nobody built
for is told so:

```
kind: sandbox
reference: ghcr.io/team/agent:1.4.0
digest: sha256:c0ffee…
image: built from ./image/Containerfile (2 lines, context 2 files), amd64 sha256:9ab3c02e…, arm64 sha256:761743bb…
…

$ lns run ghcr.io/team/agent:1.4.0                  # on an amd64 host, before CI pushed
error: the image index ghcr.io/team/agent@sha256:aa41f0c9… holds linux/arm64,
and this host is linux/amd64; nobody has built this document for linux/amd64 —
run `lns push` for it on an amd64 host to add it
```

## Fix round — the review of 2026-09, and the host run at `a74de7d4`

The host ran every Containerfile feature on an arm64 Mac. Four passed. Two
failed, both with the same message shape:

```
Error: fetching the sandbox's base image 127.0.0.1:59961/e2e-built-sandbox@sha256:9d4905c6…
  — manifest digest mismatch — expected sha256:9d4905c6…, received sha256:ea0…
```

That is review item 1, reproduced. **Both failures are that one defect.** The
message is written in exactly one place, `image_store::pull` (`crates/lns-service/src/image_store/mod.rs:706`),
which only the `PullImage` request reaches, which only `lns pull` sends — a push
never pulls a document. The two runs differ only in what the pin was: the slice 5
feature failed against a one-entry index, the slice 6 feature against the
two-entry index its second push had just published, which is why the expected
digests differ and the received one does not. So the slice 6 feature reached the
step after its second push — the pull — and failed there. I looked for a second
cause on the push path and found none; the push path is now covered by the
tests below, and the two features will say for certain on the next host run.

| # | Review item | Commit | Test that pins it |
|---|---|---|---|
| 1 | Every document this slice publishes is unpullable | `fix(service): a pull verifies a pinned index against its own bytes, then the child it holds` | `image/mod.rs`: `a_pull_pinned_to_an_index_verifies_the_pin_against_the_index_and_the_child_it_holds`, `a_child_the_index_did_not_name_is_refused_even_though_the_index_matches_the_pin`, `a_pin_that_is_neither_the_manifest_nor_an_index_this_host_is_in_stays_a_mismatch`; `image_index.rs`: `an_index_a_registry_served_is_read_back_with_the_digest_a_document_pins_it_by`, `a_plain_manifest_is_no_index_and_holds_no_entry` |
| 2 | A failed read is read as "this architecture published nothing" | `fix(cli): a registry read that fails stops the push instead of erasing an entry` (+ `fix(cli): a read that cannot authenticate keeps the login recipe`) | `build/push_auth.rs`: `a_tag_no_push_has_written_yet_reads_as_holding_nothing`, `a_read_that_failed_is_never_read_as_an_architecture_that_published_nothing` |
| 3 | The slice 5 @microvm feature was broken and not updated | `test(e2e): the slice 5 feature asserts the image line this round prints` | `microvm_containerfile_push.feature` itself, with two new steps |
| 7 | An entry for a platform outside `ARCHITECTURES` is dropped | `fix(cli): an index entry this lns cannot build for survives the next push` | `artifact/distribute.rs`: `an_entry_for_a_platform_this_lns_does_not_build_survives_the_next_push` |
| 4 | New decisions in a coverage-exempt composition root | `refactor(service): the disclosure of what an index holds is tested where it decides` | `artifact/inspect.rs`: four `built_architectures_of` tests over a `Registry` double, plus the double's own refusals |
| 5 | A lint suppression that describes another function | `refactor(service): a suppression that describes another function is deleted` | the per-crate complexity gate, which passes without it |
| 6 | The architecture line fires on every ingest | `fix(service): the architecture line belongs to a run, not to every ingest` | `run/mod.rs`: `the_run_reports_the_image_it_booted_and_the_architecture_it_booted_it_on`, `an_imageless_run_reports_no_image_it_booted` |
| 8 | The body advertises a refusal no host can print | this description — the transcript above is corrected | — |

### 1, in detail

`pull_manifest_and_config` resolves an index to the child for this host and
answers with the child's digest, so comparing that to the reference's pin
refused every index a document names. The pin is now verified against **what the
reference names**:

- a plain manifest is its own bytes, exactly as before;
- an index is verified over the **index bytes** — `lns_artifact::image_index::held`
  hashes what arrived, so no `Docker-Content-Digest` header decides a pin — and
  the entry it holds for this host must be the child the registry answered with.

A registry that answers a child the index did not name is refused, naming both
digests. A pin that matches neither is the mismatch it always was. The extra
read happens only when the plain comparison fails, so no pull of a plain
manifest makes a second request.

The `Registry` port answers with the index as it arrived (`HeldIndex`: the
digest of its own bytes and the entries it holds) instead of a bare entry list.
`peek_base_with` verifies the same way, so a `FROM` pinned to a public index —
`FROM debian:12@sha256:…`, which is an index on every registry that publishes
multi-arch — now resolves too. That was broken before this stack.

Red first: reverting the one call in `pull_inner` to the old `verify_digest_pin`
makes the new test fail with the host's own message,
`manifest digest mismatch for ghcr.io/team/hermes@sha256:… — expected …, received …`.

### 2, in detail

A read now answers "nothing here" only where the registry says so: a 404, or an
envelope naming `MANIFEST_UNKNOWN` or `NAME_UNKNOWN`. `NAME_UNKNOWN` is in that
list beyond the review's two: a first push to a repository the registry has
never heard of answers with it, and that repository does hold no image. Every
other failure — a 429, a 5xx, an unreachable registry, a refused login —
propagates, named with the tag it was reading. A refused handshake keeps the
`lns login` recipe `push_error` already builds.

### 7, in detail

`held_entries` now seeds from the entries the index tag holds today and then
adds or replaces the entry each per-architecture tag owns. The tags stay the
record for the platforms this lns builds; anything else the index holds — an
architecture a newer lns published — is carried forward. The push makes no extra
request for it: the index read that decides whether to repair a stale tag is the
one that seeds the entries, and it moved ahead of them.

## The slice 6 row of the Plan table, mapped to code and tests

|---|---|---|
| A push from a second architecture adds an entry to the image index instead of replacing it | new `lns_artifact::image_index` (`with_entry`, `in_index_order`, `assemble`, `parse`); `distribute::publish_the_image` reads the per-architecture tags, adds its own entry and publishes the index; `PushableImage` carries the platform its config declares, filled by `containerfile::image::pushable` | Layer 3 `image_index.rs` 11 tests (add an entry, replace this architecture's entry and keep the others, one digest for one set whatever order it arrives in, an empty index refused); `distribute.rs` `a_push_from_a_second_architecture_adds_its_entry_to_the_index_the_first_published`, `an_index_the_registry_refuses_stops_the_push_naming_the_repository`, `a_repository_this_push_cannot_read_stops_it_before_it_publishes_anything`; `containerfile/image.rs` `a_pushable_image_carries_the_platform_the_index_will_publish_it_under`, `an_image_whose_config_declares_no_platform_is_not_one_an_index_can_hold`; Layer 2 `image_push.feature` "a push from a second architecture adds an entry to the index the first published" |
| The published document's `spec.image` names the index by digest | `publish_the_image` answers with `{repository}@{index digest}`, which `rewrite_to_built` writes | Layer 3 `the_published_document_names_the_index_rather_than_this_architecture_s_manifest`; Layer 2 "the published document names the index by digest and keeps the path the author wrote" |
| `pull` picks the entry for the host's architecture | **no new selection code** — the pull client already resolves an index by `linux_platform_resolver` (`image::want_arch`), which is the digest-pinned selection this row asks for; `image_index::select` is what the refusal and inspect decide by | Layer 3 `image_index.rs` `a_pull_takes_the_entry_for_this_host_by_digest`; Layer 1 `microvm_containerfile_arch.feature` |
| A run on a host nobody built for says so, names the architectures the index holds, and names the command that would add this one | `image_index::refuse_missing_architecture` + `image::explain_architecture`, which reads the index back through the new `Registry::pull_index` when a pull failed | Layer 3 `image_index.rs` `a_host_nobody_built_for_is_told_what_is_held_and_what_would_add_its_own`; `image/mod.rs` `a_pull_of_an_index_nobody_built_this_host_s_architecture_for_says_what_it_holds`, `a_pull_that_failed_for_another_reason_keeps_the_registry_s_own_error`, `an_index_that_does_hold_this_host_s_architecture_explains_nothing_away`, `a_reference_whose_index_cannot_be_read_keeps_the_failure_that_was_already_there` |
| `inspect` prints one built digest per architecture | `artifact/inspect.rs::built_architectures` + `BuiltImageDisclosure`, filled by `artifact/real.rs::built_architectures_of`; `SandboxView.image_architectures`; the CLI's `image_line` | Layer 3 `inspect.rs` `a_sandbox_built_from_a_containerfile_discloses_it_beside_the_digest_it_runs`; Layer 2 `image_pulled_disclosure.feature`, 3 scenarios (two architectures, an index this machine cannot read, an image nobody built) |
| The run summary prints the architecture it booted | `ingest::run` reports `Image  <ref>, linux/<arch>` after the architecture check that already ran there | Layer 3 `ingest.rs` `the_run_reports_the_image_it_booted_and_the_architecture_it_booted_it_on`; Layer 1 "the run booted this host's architecture" |
| `push --dry-run` says which architectures the index holds and which this push would add | `distribute::preview_the_image` reads the same tags and adds two lines | Layer 3 `a_dry_run_says_what_the_index_holds_and_what_this_push_would_add`, `a_dry_run_of_an_architecture_the_index_already_holds_says_it_would_add_nothing`, `a_dry_run_of_a_repository_nobody_has_pushed_to_says_the_index_holds_nothing_yet`; Layer 2 "push --dry-run says what the index holds and what this push would add" |
| A push whose image for this architecture is already in the index and whose key is unchanged publishes nothing and says so | the `held == entries` arm of `publish_the_image` | Layer 3 `a_push_whose_architecture_the_index_already_holds_publishes_nothing_and_says_so`; Layer 2 the scenario of the same name |
| The index is updated atomically from the pusher's view | the per-architecture tags are the record and the index is derived from them; a push that adds no entry still republishes an index that is not the one its entries assemble to | Layer 3 `a_push_that_adds_no_entry_still_repairs_an_index_a_race_left_behind`; see [registry semantics](#the-registry-semantics-this-relies-on) |
| The spec states the index, the selection rule and the refusal | `docs/sandbox-spec.md` §6.2 (its own commit), `docs/cli-spec.md` push / pull / inspect rows | — |

## What a push writes, and where slice 6 reads it

Three names, all derived from the `<REF>` the author typed:

| What | Tag | Written by |
|---|---|---|
| One architecture's image manifest | `<TAG>-image-<OS>-<ARCH>` | Only a push running on that architecture. |
| The image index over every architecture | `<TAG>-image` | Every push. |
| The document | `<TAG>` | Every push. |

The per-architecture tags are the record; the index is derived from them. That is
the change against the shape slice 5 proposed in its answers section: it named a
`sha256-<digest>` tag per manifest, which no other push can find, because a
digest differs per architecture. A tag derived from the reference can be found
without listing the repository — which the registry facade cannot do
(lens-artifact-hub#134) — so a push reads `ARCHITECTURES` tags and nothing else.
`lns_artifact::image::published_tag` is deleted with the shape it served.

Assembly is deterministic: entries are sorted by platform, so one set of entries
is always the same bytes and the same digest, and a push that adds nothing names
the index the last push named.

## The registry semantics this relies on

**What the registry gives.** A manifest `PUT` is atomic and a tag names one
manifest at a time, so no architecture can corrupt another's tag: each push
writes only `<TAG>-image-<OS>-<ARCH>` for the architecture it built on. A blob or
manifest `PUT` of bytes that are already there is a no-op, so re-publishing an
identical index costs nothing and cannot damage it. Pulling by digest is exact,
so the document names an index that cannot change under it.

**What it does not give.** There is no compare-and-swap on a tag in the OCI
distribution spec — no conditional `PUT`, no `If-Match` a registry has to honour.
So two pushes from different architectures that overlap can each read the tags
before the other has written, and each publishes an index without the other's
entry; the document that lands last then names an index missing one architecture.
**Nothing is lost when that happens**: each architecture's manifest is still under
the tag it owns, so the next push from either host assembles the full index again.
This round makes that repair automatic rather than conditional — a push that adds
no entry of its own still republishes the index when the index tag does not name
what its entries assemble to. What it cannot do is make the window zero; closing
that needs either a registry lock or a conditional write nothing in the spec
provides, and it is not invented here.

A consequence worth stating: the index a document names is the one the *pusher*
saw. A consumer pulling that digest gets exactly those entries, even if a later
push added another. Re-pushing the document is what publishes the wider index.

cloud-hypervisor/virtiofsd. The feature is written to be right by reasoning and
**is unverified by me**. One command on an arm64 Mac:

```
LNS_E2E_FEATURE="a second architecture adds an entry to the image index a document names" make e2e-microvm
```

`crates/e2e-tests/features/microvm_containerfile_arch.feature` pushes a path-form
document to the e2e registry and asserts the index holds this host's architecture
and nothing else. It then writes the other architecture's image straight into the
registry — `LocalRegistry::publish_architecture_image`, a documented harness seam,
under exactly the tag a push from such a host would own, because this harness has
one architecture to build on. Everything after that is real: the second push says
it publishes nothing and the index it names holds both entries; a home that has
never built anything pulls the document; `inspect` prints both digests and boots
nothing; and the run boots this host's entry and reads the file the `RUN` wrote.

The slice 3 and 4 scenarios are unchanged. `microvm_containerfile_push.feature`
now asserts the `image:` line this round prints and the built digest on it (review
item 3), and asserts that the push names the index it published in the artifact's
own repository — the assertion that used to read the repository off inspect's
image line, which no longer carries one.

## Deviations

- **The index module is in `lns-artifact`, not in `lns-service`.** The brief asks
  for the Layer 3 tests of index assembly in `lns-service`. Both sides decide by
  the same rules — the CLI assembles the index at push, the service selects from
  it at pull and reads it for `inspect` — and this repository's rule is one
  parser, one policy, read by every caller (slice 5's review item 5). So the
  entry type, the tags, the add-or-replace, the assembly, the digest-pinned
  selection and the refusal live in `lns_artifact::image_index` with their tests,
  and `lns-service` holds the Layer 3 tests for what it owns: the platform a
  built image carries, the refusal as a pull produces it, and the disclosure
  `inspect` projects.
- **`inspect`'s `image:` line replaces slice 5's `imageSource:` line.** The Plan's
  CLI Surface line is `image: built from ./image/Dockerfile (4 lines, context 2
  files), arm64 sha256:…, amd64 sha256:…`, and slice 5 printed the same facts as
  two lines with the index reference on the first. This round moves to the line
  the issue names. An artifact whose index this machine cannot read keeps the
  reference in that position, so the line always says what the guest starts from.
- **The push reads the registry, and so does the dry run.** Assembling an index
  over what the repository already holds means reading it: one manifest and its
  config per architecture, plus the digest the index tag names. A dry run makes
  the same reads and uploads nothing — it already went online in slice 5 to peek
  the base image.
- **An entry's platform is read from the manifest's own config**, not from the
  tag it was found under. A tag says who wrote it; only the config says what the
  image is, and a consumer selecting by platform must not be sent an image its
  host cannot run.
- **`ARCHITECTURES` is `amd64` and `arm64`.** They are the two lns builds a
  kernel and guest tools for (`crates/lns-service/src/kernel.rs`). A third
  architecture is one line here and a kernel pin there.
- **A run on a host nobody built for is refused by explaining the failure the
  registry already produced**, not by reading the index up front. The happy path
  makes no extra request; only a pull that failed asks what the index holds, and
  a failure that is not about architecture is passed on as the registry spelled
  it.
- **`pull_inner` was split.** The index refusal pushed it over the complexity
  gate, so the manifest checks a pull makes before fetching anything and the line
  it prints about what the pull will cost are each their own named step. No
  behaviour changed; the existing tests cover both.
- **The run's architecture line is a service log verb**, not a line the CLI
  composes: `Image  <ref>, linux/<arch>`, reported by the run orchestrator, so
  only a run prints it. The Plan calls it a run-summary line; this is where the
  run summary is written.
- **A pull makes one extra registry read only when a pin does not match the
  manifest the registry answered.** That is what tells an index from a tampered
  answer. A pull of a plain digest-pinned manifest makes no extra request.

## What is left out

- **No garbage collection of a superseded entry.** A push that replaces this
  architecture's entry leaves the old manifest in the repository under no tag of
  its own. That is the same open decision slice 5 named — who removes a published
  image nothing names — and it is not answered here.
- **No `lns push --platform`.** The build always happens where the push runs
  (#393, "Where it builds"), so a host cannot publish an entry for an
  architecture it cannot build. Cross-building is not in this proposal.
- **The window in the two-pusher race is not closed**, for the reason above.
- **`--rebuild` on a reference-form `spec.image` is still a silent no-op**, and
  the build context is still walked under fileset rules on the push path — both
  inherited from slice 5, both unchanged here.

`lns-service` does not link in this container (no libgtk-3/libatk/libxdo on the
image), so the harness was made to compile and link against stub pkg-config files
and stub shared objects with `--unresolved-symbols=ignore-all`. **Nothing in the
repository changed for it.**

| Command | Result |
|---|---|
| `cargo test -p lns-artifact` | 466 passed |
| `cargo test -p lns-cli` | 1011 unit tests; cucumber 582 scenarios / 2934 steps, all green |
| `cargo test -p lns-service` | 2672 unit tests; cucumber 259 scenarios / 1032 steps, all green |
| `cargo build -p e2e-tests --tests` | builds (the @microvm scenarios do not run here) |
| `cargo clippy --all-targets -- -D warnings` on lns-artifact, lns-cli, lns-service, e2e-tests | clean |
| `cargo clippy --all-targets -- -D clippy::cognitive_complexity`, per crate, on lns-artifact, lns-cli, lns-service | clean |
| `make coverage` | **not run here** — the instrumented workspace does not fit in this container's disk, so CI is the oracle. |

Two coverage exemptions changed reason this round, and neither grew: the
`crates/lns-cli/src/build/push.rs` reason now names `names_nothing_yet` in
`build/push_auth.rs`, which is where the one rule those reads own is host-tested;
the `crates/lns-service/src/artifact/real.rs` reason names
`built_architectures_of`, which moved out of it into `artifact/inspect.rs`
(review item 4).

## Answers to the review's questions

**Does the registry GC you target keep an untagged index a document still names by digest?**
Not everywhere, and this is the one open risk I want a human decision on. The
index tag `<TAG>-image` is reused by every push, so the previous index becomes
untagged the moment a second architecture publishes. GHCR and Docker Hub keep
untagged manifests until somebody deletes them; the reference `registry:2`
garbage collector deletes every manifest no tag reaches, which would break an
already-published document that names that index by digest. The fix that stays
inside this shape is one more tag per index — `<TAG>-image-<INDEX-DIGEST>` — so
every index a document can name stays tag-reachable and nothing is ever
overwritten. It is not in this round: it is a new name in the repository, so
`docs/sandbox-spec.md` §6.2 has to state it first.

**Slice 7 needs `inspect` to mark an image built outside the gate. Where does that marker live?**
Both, with the wire form deciding: **an annotation on the index entry**, carried
as a field on `IndexEntry` and projected onto `BuiltArchitecture`.

- The wire form is `annotations` on the entry's descriptor *inside the index* —
  for example `run.lns.built-outside-the-gate: "true"` — not on the
  per-architecture manifest. Two reasons. The document pins the index by digest,
  so an annotation on an entry is covered by the digest a consumer already
  verifies; an annotation on the manifest is not, and a registry could serve a
  different one. And `inspect` reads the index and nothing else today: marking
  the manifest would cost one request per architecture, on a command that is
  meant to print without fetching.
- The model then carries `IndexEntry { …, built_outside_the_gate: bool }`, which
  `image_index::parse` reads and `assemble` writes, and
  `inspect::built_architectures` projects onto `BuiltArchitecture` so the CLI
  line can mark it. One parser, one policy, both sides reading the same field.
- `docs/sandbox-spec.md` §6.2 states the annotation and its exact spelling
  before any of it lands, per the transitional-mode rule.

**Is the Docker path expected to bypass `ensure_runnable_here`, and what keeps the cache key's arch in step with the entry's?**
No. `resolve_base` refuses a base this host cannot run, so the lns path can only
build for the architecture it runs on, which is what makes "the build happens
where the push runs" true. A Docker path that cross-builds would have to be held
to the same rule or publish under `--platform` explicitly; it is not proposed
here. The key is already per-architecture without carrying the architecture: it
stands on the **resolved** `FROM` digest, which is the platform child's manifest
digest, so an amd64 and an arm64 build of the same Containerfile have different
keys. A cross-building path would break that only if it stopped pinning the base
per target platform. The entry's own platform is never taken from the key or the
tag — it is read from the built image's config — so an index can never mislabel
an image even if a key went wrong.

**The Plan asks the run summary to print the architecture it booted; this lands as a service log verb. Deviation to list?**
Yes, and it is listed below. It also moved this round: it now fires in the run
orchestrator, so only a run prints it (item 6).

**Is a digest-form push target reachable?**
It is reachable at the CLI — nothing refuses `lns push ghcr.io/team/agent@sha256:…`
— and it was already meaningless before this round: the document manifest would
be `PUT` to a digest the artifact it is uploading does not have. `tag_of`
answering `latest` there is that shape and no other. The honest fix is to refuse
a digest-form push target with a message that says a push names a tag; that is a
CLI-grammar decision (`docs/cli-spec.md` §push), so I have left it for a human
rather than inventing it in a fix round.

**A Docker schema2 base puts a Docker manifest media type into an OCI index entry. Which registries has this been pushed to?**
Only the in-process registry the e2e harness runs. I cannot reach a real
registry from where I work, so treat it as unverified. It is the same mix
`docker buildx` publishes (an OCI index whose entries are Docker manifests), and
GHCR, Docker Hub and `registry:2` accept it, but I have not proved that here.
The entry's media type is read from the manifest the push built, so it is always
what the registry actually holds.

## CI

CI runs only for pull requests based on `main`. This PR was pointed at `main` for
this round and answered `CONFLICTING` with no run at all, because slices 0 to 5
are not merged — so the run was taken the way slice 5's and the first slice 6
round's were: on `ci-oracle/containerfile-arch`, which is `main` with the whole
stack merged onto it, through the throwaway
[#417](https://github.com/lensapp/lens-sandbox/pull/417). The base is back on
`feat/containerfile-push`.

**The fix-round run: https://github.com/lensapp/lens-sandbox/actions/runs/34432698949
— every job green**, including `lint`, `test` (which carries `make e2e`),
`coverage`, `coverage-as-root`, `shell-tests-macos` and both
`check-release-build` cross-builds. (The `labels` check on the throwaway PR fails
because a throwaway carries no type label; it is not part of this branch.)

The first attempt at that run, https://github.com/lensapp/lens-sandbox/actions/runs/34432091433,
failed `coverage` on one line: the `ImageManifestNotFoundError` arm of
`names_nothing_yet`, which no test reached. `test(cli): every spelling of "the
registry holds nothing here" is exercised` fixes it.

The previous round's run, for the code before these fixes, was
https://github.com/lensapp/lens-sandbox/actions/runs/34427615088.

Auto-merge is not armed: a stacked PR's base carries no required checks, and
arming it merges the PR unreviewed.


chenhunghan (commented):
Reading review of slice 6 of lens-sandbox issue #393, against `feat/containerfile-push`. Nothing was run in the review; the host verification of this branch on an arm64 Mac is running as this is posted and its result follows in the fix round's brief. Every claim was verified by reading this branch and the pinned `oci-client` 0.16.1 source. Verdict at the end.

# Findings, ranked

**1. Every document this slice publishes is unpullable: the pin check refuses an index digest.** `distribute.rs:528` rewrites `spec.image` to `{repository}@{index digest}`; every consumer path runs it through `pull_inner`, and `image/mod.rs:597` `verify_digest_pin` requires the digest in the reference to equal the digest the registry answered with. `oci-client`'s `pull_manifest_and_config` resolves an index to the platform child and answers with the child's digest, never the index's. Failure: `lns push ghcr.io/team/agent:1.4.0` publishes index `sha256:AAAA` holding `linux/arm64 sha256:BBBB`; `lns pull ghcr.io/team/agent:1.4.0` fails at `pull_dependency` with a pin mismatch. No test covers it: every new `pull_inner` test sets `manifest_failure` first, and the @microvm scenario was not run. Fix: read the raw manifest first, verify the pin against the index bytes when the reference names an index, then verify the selected child against the entry it came from.

**2. A registry read that fails is read as "this architecture published nothing", so a push can drop the other architecture's entry.** `build/push.rs:146` `image_at` collapses every failure into `Ok(None)`, as does the auth handshake at `:139-145`; `held_entries` at `distribute.rs:88` treats `None` as an empty tag. Failure: CI published `1.4.0-image-linux-amd64`; an arm64 push meets one 429 on that GET, assembles an index over arm64 alone and overwrites `1.4.0-image`. The Layer 3 test `a_repository_this_push_cannot_read_stops_it_before_it_publishes_anything` asserts a behaviour the real port does not have. Fix: `Ok(None)` only for 404 or `MANIFEST_UNKNOWN`; propagate everything else.

**3. The slice 5 @microvm scenario is broken by this change and was not updated.** `microvm_containerfile_push.feature:42` asserts `imageSource: built from ./image/Dockerfile`, a line this PR deletes from `render_build_source`; and `steps/microvm.rs:2089` looks for `{repository}@sha256:` on the `image:` line, which now prints the per-architecture digests and no repository. The body says the feature was updated; `git diff --stat` shows it unchanged. Fix: assert the line the code prints, and the per-architecture digests.

**4. New decisions landed in a coverage-exempt composition root.** `artifact/real.rs:447-463` `built_architectures_of` holds three untested rules (unparseable document, unparseable image, registry error each disclose nothing) under the exemption at `scripts/coverage-floor.sh:66`, whose reason was not touched. Fix: move them beside `inspect::built_architectures`, or name them in the reason.

**5. A lint suppression survived a split and now describes a different function.** `image/mod.rs:470` `announce_the_pull` carries `#[allow(clippy::cognitive_complexity)]` with a reason about manifest fetch, digest verify and layer pulls, none of which it does. Fix: delete it, or attach it to the function that needs it with its own reason.

**6. The architecture line fires on every ingest, not only on a run.** `ingest.rs:29` is shared by the run, the build guest's `FROM` and tool provisioning, so `lns push` and `lns pull` now print an `Image … linux/arm64` line that is not a run summary. Fix: report from the run orchestrator.

**7. An entry for a platform outside `ARCHITECTURES` is dropped on the next push.** `image_index.rs:11` fixes the list to amd64 and arm64 and `held_entries` reads only those tags; `index_at` returns a digest and no entries. A newer lns that adds a third architecture publishes it; an older lns's next push republishes an index without it. Fix: seed `held_entries` from the entries the current index holds, then add or replace this architecture's.

**8. The PR body advertises a refusal no host can print.** `want_arch()` at `image/mod.rs:153` reports amd64 for every non-aarch64 host, so the body's riscv64 transcript cannot happen. Fix: correct the transcript, or refuse a host with no mapping.

Confirmed by reading: the index is one platform manifest per architecture, sorted, so one entry set is one digest; `with_entry` replaces only the matching platform; a push that adds no entry still republishes a stale index tag; auto-merge was not armed.

# Questions for the author

- The index tag `<TAG>-image` is reused by every push, so an already-published document's index loses its only tag on the next push. Does the registry GC you target keep an untagged index a document still names by digest?
- Slice 7 needs `inspect` to mark an image built outside the gate. `IndexEntry` and `BuiltArchitecture` carry only architecture and digest. Where does that marker live?
- `resolve_base` enforces `ensure_runnable_here` against `want_arch()`, so the lns path cannot build for a foreign architecture. Is the Docker path expected to bypass that, and what keeps the cache key's arch in step with the entry's?
- The Plan asks the run summary to print the architecture it booted; this lands as a service log verb. Deviation to list?
- `tag_of` answers `latest` for a digest-form target while `repository_of` strips the digest. Is a digest-form push target reachable?
- A Docker schema2 base puts a Docker manifest media type into an OCI index entry. Which registries has this been pushed to?

# Verdict

The index model is sound and honest about its race. Before merge: 1 (nothing published is pullable), 2 (a transient read error can erase another architecture), 3 (a slice 5 feature regressed), 7 (unknown platforms are dropped). Can follow: 4, 5, 6, 8. The two @microvm features touching push and pull are run on the Mac after the round.

chenhunghan:

PR #418, opened 2026-09-10
[reading only, merges via #422] feat: a configuration switch builds a Containerfile on the host Docker daemon

chenhunghan:

Part of #393 — slice 7, the host-Docker switch. Stacked on
[#415](https://github.com/lensapp/lens-sandbox/pull/415) (slice 6, the
per-architecture image index) and opened against `feat/containerfile-arch`.

A machine that has a Docker daemon and prefers it may point the build at that
daemon instead of a build guest. The gate does not apply there, so lns says so
everywhere it names the image:

```
$ lns config set build.engine docker
Set build.engine to docker in /home/you/.lns/config.yaml

$ lns config list
build.engine = docker

$ lns run .
   Image  built from ./image/Containerfile, arm64, sha256:761743bb… — built outside the gate by the host Docker daemon

$ lns push ghcr.io/team/agent:1.4.0
key sha256:5f0b1a2c…
built ./image/Containerfile as ghcr.io/team/agent@sha256:761743bb… (2 layers) — built outside the gate by the host Docker daemon
index ghcr.io/team/agent@sha256:aa41f0c9… holds linux/arm64 sha256:761743bb…

image: built from ./image/Containerfile (2 lines, context 2 files), arm64 sha256:761743bb… (arm64 was built outside the gate by the host Docker daemon)

$ lns config set build.engine lns
Set build.engine to lns in /home/you/.lns/config.yaml

$ lns run .
   Image  built from ./image/Containerfile, arm64, sha256:761743bb… — built outside the gate by the host Docker daemon
```

The last two lines are the point of this round. The switch is off again, the key
is the same, and the image the key answers with is still the daemon's — so the
run still says so, and a push of it still publishes the annotation.

With the switch off, nothing changes. A machine whose switch says `docker` and
whose daemon does not answer refuses by name, and names the switch to turn off:

```
$ lns push ghcr.io/team/agent:1.4.0
error: building ./image/Containerfile: this machine's build.engine is docker, and
no Docker daemon answered at /var/run/docker.sock; start the daemon, point
build.dockerSocket at the socket it listens on, or run `lns config set
build.engine lns` to build in a guest instead: No such file or directory (os error 2)
```

The last clause is what the socket answered, so `Permission denied` reads
differently from a socket that is not there at all.

## The slice 7 row of the Plan table, mapped to code and tests

|---|---|---|
| A configuration switch points the build at a host Docker daemon | `build.engine` (`lns`, the default, or `docker`) and `build.dockerSocket` in the family `push.imageLimit` opened: `config::BuildSection`, `ConfigKey::{BuildEngine,BuildDockerSocket}`, `config::load_build_engine`. It reaches the service as `lns_ipc::BuildEngine` on `RunImageArgs`, `Request::BuildSandbox` and `Request::BuildImageForPush` | Layer 3 `config/mod.rs`: `a_machine_that_says_nothing_about_the_engine_builds_in_a_guest`, `a_machine_that_prefers_its_daemon_builds_through_the_socket_it_names`, `a_socket_named_beside_the_default_engine_changes_nothing_about_where_a_build_runs`, `an_engine_lns_does_not_have_is_refused_naming_the_two_it_does`, `an_engine_a_hand_edited_file_invented_is_refused_where_a_build_reads_it`, `a_docker_socket_that_is_no_path_is_refused_where_it_is_set`, `a_run_carries_the_engine_this_machine_builds_with`, and the `every_key_survives_a_set_get_list_unset_lifecycle` rows for both keys |
| through the Docker Engine API | new `lns-service/src/containerfile/docker.rs`: `GET /v1.43/_ping`, `POST /v1.43/build`, `GET /v1.43/images/{tag}/get`, over a `Daemon` port that is one HTTP/1.1 exchange on a Unix socket. The request encoder, the chunked/plain response decoder and every refusal live there; `RealDaemon` in `containerfile/real.rs` is the socket | Layer 3 `docker.rs`, 30 tests: the request bytes, query escaping, chunked and plain decoding, five truncated-answer refusals, the ping, the build stream's own error, an outright refusal, the export |
| the Containerfile and its context are sent to the daemon | `docker::context_tar` over the same `ContextFs` the executor's `COPY` reads, so §3.1.1's rule holds — a symlink is not part of a context and the daemon never sees one; `Located::in_context` names the file inside the tar; `executor::arg_defaults` states every `ARG` default as a build argument | `the_context_goes_up_as_a_tar_of_its_files_and_directories`, `a_context_file_the_host_cannot_read_names_it_rather_than_sending_half_a_context`, `an_entry_that_disappeared_between_the_listing_and_the_read_is_left_out`, `a_context_directory_that_cannot_be_listed_names_it`; `locate.rs` `the_file_is_named_inside_the_context_for_an_engine_that_reads_only_the_tar`; `executor.rs` `every_arg_with_a_default_is_a_build_argument_expanded_against_the_ones_before_it`; `a_build_sends_the_context_the_file_inside_it_and_the_arg_defaults` |
| the image is exported and imported into lns's layer store the way the executor's images are | `docker::read_saved_image` reads the `docker save` tar (`manifest.json` → config blob → layer blobs), `docker::as_oci_image` describes them as the layer store already holds a built image, and `DockerHost::adopt` lands them through the unchanged `containerfile::import::import`, which now takes every layer rather than one | `an_exported_image_reads_back_as_its_config_and_every_layer_in_order`, `an_export_missing_what_its_manifest_names_is_refused_rather_than_half_imported` (9 spellings), `an_imported_image_is_addressed_the_way_every_other_built_image_is`, `a_daemon_that_exported_gzipped_layers_declares_them_as_gzipped`, `an_export_with_no_layer_is_nothing_a_guest_could_boot` |
| so the rest of the pipeline (slice 4's key, slice 5's publish, slice 6's index) is unchanged | `docker::build_image` takes the key over the **same** four inputs `executor::open` does — the resolved `FROM` digest, the file's text, the context hash, the architecture — by peeking the base through the same port, and answers the same `executor::Built`. Nothing in the push, the publish or the index knows which engine ran | `a_daemon_build_lands_in_the_local_store_under_the_key_the_lns_path_would_use` (asserts the key equals `key::image_key(...)` outright), `a_key_this_machine_already_answers_runs_no_daemon_at_all`, `a_rebuild_ignores_the_key_this_machine_holds_and_builds_again` |
| the gate does not apply to a daemon build, and both `lns inspect` and the run summary say so in words | one string, `lns_artifact::image_index::BUILT_OUTSIDE_THE_GATE`, read by both sides: `step::built_line` for the run summary and the CLI's `image_line` for `inspect` | `step.rs` `an_image_the_host_daemon_built_says_so_in_the_line_the_run_summary_prints`; Layer 2 `image_docker_engine.feature` "inspect marks the architecture a daemon built and leaves the other alone" |
| the record is an annotation on the entry's descriptor inside the image index, carried as a field on `IndexEntry` and projected onto `BuiltArchitecture` | `image_index::BUILT_OUTSIDE_THE_GATE_ANNOTATION` (`run.lns.built-outside-the-gate: "true"`), written by `assemble` and read by `parse`; `IndexEntry.built_outside_the_gate`; `inspect::built_architectures` projects it onto `BuiltArchitecture`; `PushableImage.built_outside_the_gate` carries it from the service to the uploader and `distribute::entry_of` puts it on the entry | Layer 3 `image_index.rs` `an_image_the_host_daemon_built_is_annotated_on_the_entry_the_index_holds`, `an_image_a_build_guest_built_carries_no_annotation_at_all`, `what_the_index_records_about_the_gate_reads_back_per_architecture`, `an_entry_annotated_with_anything_but_true_is_not_one_built_outside_the_gate`; `containerfile/image.rs` `an_image_the_host_daemon_built_is_handed_to_the_uploader_saying_so`; Layer 2 the two push scenarios |
| a machine whose switch says `docker` but has no reachable daemon refuses by name and names the switch to turn off | `docker::refuse_an_unreachable_daemon`, asked as a ping **before** the context is sent | `a_daemon_that_does_not_answer_names_the_socket_and_the_switch_to_turn_off`, `a_daemon_that_answers_a_ping_with_anything_but_200_is_no_daemon_to_build_on`, `a_daemon_that_does_not_answer_stops_the_build_before_the_context_is_sent` |
| `lns config list` shows it and `docs/cli-reference.md` describes it | `ConfigKey::ALL` gains both keys; the reference gains both rows, a worked example of what the switch changes, and the refusal | the lifecycle test above asserts `build.engine = docker` and `build.dockerSocket = /var/run/docker.sock` in the listing |
| the spec states the switch and the disclosure | `docs/sandbox-spec.md` §3.1.1 (the switch, the gate, the refusal) and §6.2 (the annotation and its exact spelling), `docs/cli-spec.md` §3.5 and the `inspect` row — all in their own commit, before any code | — |

## The review of 2026-09-10, item by item

Every finding is answered by one commit.

| # | Finding | What changed | Commit |
|---|---|---|---|
| 1 | The gate record came from the switch that is set now, so the PR's own recipe published a daemon build as gated | The cache entry records the engine that filled the key; `Cached { reference, built_outside_the_gate }` is what `BuildHost::cached` and `DockerHost::cached` answer with; `executor::Built` and `executor::Planned` carry the record; the run summary, `lns push` and the index all read it off the entry the reuse answered with. `BuildEngine::is_outside_the_gate` is deleted — the switch decides where a build runs and nothing else | `241ba63a` (the spec decision) and `c1128848` |
| 2 | The daemon resolved `FROM` itself, so the key stood on a digest the image may not have been built on | The Containerfile inside the context tar is rewritten to the digest-pinned reference `peek_base` answered with, before the tar is sent. The key still measures the file as written | `39562672` |
| 3 | A consumer running a pulled artifact was never told the gate did not apply | The pull reads the index entry its manifest came from, `PulledImage` and the ingest carry the record, and the run's image line says it in words. An image this machine built is named under a host no registry resolves, so it asks no registry for an index | `a5f04fd1` |
| 4 | `lns push` published the annotation and said nothing about it | The built line carries the phrase, off `PushableImage.built_outside_the_gate` | `b9a4f1e3` |
| 8 | The port method had a default, so no Layer 2 test could pin that the switch reaches the service | `SandboxService::build_engine` is required, both fakes answer it, and the push carries the engine on `PushPorts` / `DryRunPorts` into `ImageRequest` rather than reading the configuration inside the builder. Two scenarios pin `lns sandbox build`, one pins `lns push` | `ecdc1d56` |
| 5 | `DOCKER_HOST` was read in the tray-resident service | `lns_ipc::docker_socket` resolves the socket where `build.engine` is read — in the CLI, in the user's own environment — and the path travels on the request. The service connects to what it is given and reads no environment | `fba8b52c` |
| 6 | Every daemon build left a tagged image in the daemon | `DELETE /v1.43/images/{tag}?force=1` as soon as the export is in the layer store. A daemon that keeps its copy is named in a warning and the build stands | `26d5a09c`, `bc5646c6` |
| 7 | The context was held three times in memory | The tar is moved into the request and written to the socket after the head, so it is allocated once. Streaming the tar itself is still not done — that is the same open item as the exported image | `5d3d2b61` |
| 9 | The reference guide's refusal was not the message that ships | The guide quotes the wrapping and the io error, and says what the last clause is for | `47892c65` |
| 10 | On a two-architecture index the inspect note read as if it applied to the whole line | The note names its own architecture: `amd64 sha256:bbbb (amd64 was built outside the gate by the host Docker daemon)` | `d36dd791` |

## The review's questions

**"The spec still says 'same key, same image'; two engines now answer one key.
Is that the intent, or the defect in finding 1?"** Both, and they are separable.
The sentence is the intent: the key is a function of four inputs, the engine is
not one of them, and a machine that has already built an image for a key runs
nothing. The defect was that lns then read the gate off the switch rather than
off the image. `docs/sandbox-spec.md` §3.1.1 now says so as a decision
(`241ba63a`): the engine is not one of the four inputs, so a key one engine
filled is a key the other answers with, and an implementation MUST record which
engine built the image beside the key and take every disclosure from the image
the key answers with. Refusing to let one engine's key answer the other's build
was the alternative and is rejected: it would make the switch a fifth input,
rebuild everything twice on a machine that flips it, and hide that the two
engines are meant to produce the same image from the same inputs.

**"Should the annotation also ride on the per-architecture manifest?"** Yes, and
it is not done here. Two things argue for it. The one the review names: if the
`<TAG>-image` index tag is absent, `held_entries` starts from the per-architecture
tags alone, and every other architecture's record is dropped while its digest
survives — the digest is read back off a manifest, and a manifest says who wrote
an image and never which engine did. The second: a pull now spends one extra
`GET` on the index to learn what the entry records (`a5f04fd1`); an annotation on
the manifest would be in the bytes the pull already has. It is a format decision
in `docs/sandbox-spec.md` §6.2 and a producer change, so it belongs in its own
change rather than folded into a review round. Until then the index entry is the
record, and a manifest read back off a tag claims no gate.

**"Should a daemon build refuse a secret-shaped context file before the tar is
sent, as `lns push` refuses it after?"** A human decision, and it is not taken
here. The two engines send the same context by construction, so a refusal that
fires only for the daemon would make them differ; a refusal for both is slice 4's
open context question (its limits and wording), which is still undecided.

**"If the daemon answers 400 before draining a large body, does the user get the
unreachable-daemon wording?"** Still not verified — it needs a daemon. Reading
the code: `RealDaemon::round_trip` now writes the head and then the body, and a
daemon that answers early makes the second `write_all` fail with `EPIPE`. That
error is wrapped with `refuse_an_unreachable_daemon`, so the user would be told
to start a daemon that is in fact running and refusing. Naming it is the honest
report; fixing it means reading the socket while writing, which is a transport
change this round does not make.

**"`Built.layers` counts new layers on the lns path and every exported layer on
the daemon path. Deliberate?"** It is what the two engines can each say, not a
decision that the two numbers mean the same thing. The lns path knows which
layers it committed; the daemon path only receives an export and counts what it
imported, which includes the base's layers. The line is `(N layers)` on both, so
the same image can read differently depending on which engine built it. Making
them agree means subtracting the base's layer count on the daemon path, which is
a behaviour change with its own test; it is not made here.

## The Docker Engine API, and what is asked of it

Every request is pinned to **API version 1.43** (Docker 24): old enough for
every daemon in use, new enough for `platform` on a build. Three endpoints, and
no others:

| Endpoint | Why |
|---|---|
| `GET /v1.43/_ping` | Asked first, so a machine with the switch on and no daemon is refused before a gigabyte of context is uploaded. |
| `POST /v1.43/build?dockerfile=&t=&platform=&buildargs=&rm=1&nocache=` | The context tar is the body, `Content-Type: application/x-tar`. `nocache=1` is `--rebuild`. The daemon answers `200` before it runs, so the failure is a line of the JSON stream, not a status — `refuse_a_failed_build` reads it. |
| `GET /v1.43/images/{tag}/get` | `docker save`: the built image as a tar. The tag is derived from the build key, so two builds never collide and a rebuild replaces its own. |
| `DELETE /v1.43/images/{tag}?force=1` | The tag lns wrote is lns's own leftover inside a daemon `lns sandbox prune` cannot see, so it goes as soon as the export is in the layer store. Intermediate images are the daemon's, and `rm=1` is what asks it to drop them. |

The transport is one HTTP/1.1 exchange on a Unix socket with
`Connection: close`: the head is written, then the body, then the whole answer is
read to EOF, so a gigabyte of context is never copied into a request buffer and
no framing decision sits in the socket adapter: the encoder, the chunked and `Content-Length` decoders
and every refusal are pure functions with their own tests.

## What only a machine with Docker can verify

**The Docker Engine API path was exercised only against the fake.** There is no
Docker daemon where I work, and the e2e harness has none either, so nothing in
this PR has spoken to a real `dockerd`. What is unverified is exactly the wire:
whether a real daemon accepts the query string as spelled, whether its build
stream carries an `error` line in the shape `refuse_a_failed_build` reads, and
whether its `docker save` tar is the `manifest.json` + blob layout
`read_saved_image` expects (both the modern `blobs/sha256/…` layout and the
older `<id>/layer.tar` one are read by name out of `manifest.json`, so both
should hold, but neither has been proved).

One command on a machine with Docker, from a document whose `spec.image` is a
path:

```bash
lns config set build.engine docker
lns sandbox build          # builds through the daemon and fills the cache
lns run .                  # prints the `built outside the gate` line and boots the result
lns config set build.engine lns
```

**No `@microvm` scenario is added, and none is needed:** a daemon build boots no
lns guest. The guest-side scenarios of slices 3 to 6 are untouched.

## Deviations

- **The engine reaches the service three ways, because three verbs build.**
  `lns run` carries it on `RunArgs.build_engine`, filled by
  `config::apply_run_defaults` exactly as `run.cpus` is; `lns sandbox build`
  reads it through the required `SandboxService::build_engine` port method;
  `lns push` reads it in `artifact/real.rs`, the wiring that already reads
  `push.imageLimit`, and carries it on the push's ports. The alternative — the
  service reading `~/.lns/config.yaml` itself — would put the config parser in a
  second crate. The socket travels with the switch, because the service must not
  read the environment that started it (review finding 5).
- **The key stands on a peeked base, not an ingested one.** A daemon build pulls
  its own base, so lns reads one manifest and one config off the registry to get
  the digest its key is taken over, and fetches no layer. That is slice 5's
  `peek_base`, reused.
- **The exported image is read whole into memory, and so is the context tar.**
  So is a committed layer on the lns path (slice 1's open item 2.7). Review
  finding 7 removed the two copies of the context beyond the tar itself; streaming
  the tar and the export is one change, and it is not made here.

- **A build cache entry written before this round is not read.** The entry now
  carries the engine that filled it, and an entry that does not is not parsed —
  so it answers nothing and the next build runs again. That is the format change
  the transitional rule asks for (no shim, no second spelling); the cost is one
  rebuild per key on a machine that has built before.
- **A tag read back off the registry never claims a gate.** `build/push.rs`
  builds an `IndexEntry` from a manifest and its config, which say who wrote an
  image and never which engine did. So the record is carried forward off the
  index the repository already holds, and only for the very manifest it was
  written about — `distribute::was_built_outside_the_gate`. A digest that changed
  is a different image and records nothing.
- **The context tar is walked under the executor's context rules**, so a build
  through the daemon sends exactly what a build in a guest would copy. That
  inherits slice 5's open item 8 — the context's own limits and wording are still
  undecided — and widens nothing.
- **`lns sandbox build` does not print the disclosure itself.** The service logs
  `Built <ref> … — built outside the gate by the host Docker daemon`; the verb's
  own `built <label> as <ref>` line is unchanged, because `Response::SandboxBuilt`
  carries no such field. The two places the issue names — `lns inspect` and the
  run summary — do say it.
- **Red first, honestly.** The annotation on `IndexEntry` was written test-first
  and watched fail (`cargo test -p lns-artifact image_index`, four tests, "struct
  `IndexEntry` has no field named `built_outside_the_gate`"). The `docker.rs`
  Layer 3 tests and the Layer 2 feature of the first round were written alongside
  their code, not before it.

  In this review round: finding 1's tests were written first and watched fail
  (13 compile errors, "no field `built_outside_the_gate` on type `Built`",
  "expected `Cached`, found `String`"); finding 4's scenario was written first and
  watched fail (`585 scenarios (584 passed, 1 failed)`). Findings 2, 3, 6, 8 and
  10 had their tests written before their code, and the red I saw was the compile
  failure of the missing symbol or the missing step — I did not run each of those
  suites in the red state before writing the code. Findings 5, 7 and 9 are a
  behaviour change with its test, a refactor covered by the tests already there,
  and a document.

## What is left out

- **No `--platform` and no cross-build.** The daemon is asked for
  `linux/<this host's arch>`, so the rule that makes "the build happens where the
  push runs" true (#415's answer to the review) holds for both engines.
- **No `.dockerignore`.** Neither engine reads one; slice 4's review already
  carries that as an open question, and answering it for the daemon alone would
  make the two engines send different contexts.
- **No credential or egress bridging.** The daemon's own network and its own
  `~/.docker/config.json` decide what a `RUN` reaches and what a `FROM` pulls.
  That is the whole point of the disclosure, and it is stated in
  `docs/sandbox-spec.md` §3.1.1 rather than papered over.
- **No `docker buildx` / BuildKit options.** The classic builder endpoint is
  what is called, which is the semantics #393 decided for the lns executor too.
- **The build stream is not relayed to the user.** A daemon build prints the
  `Built …` line and nothing between; the lns path prints one line per
  instruction. Relaying the stream is a display decision and is not made here.

`lns-service` does not link in this container (no libgtk-3/libatk/libxdo on the
image), so the harness was made to compile and link against stub pkg-config
files and stub shared objects with `--unresolved-symbols=ignore-all`. **Nothing
in the repository changed for it.**

| Command | Result |
|---|---|
| `cargo test -p lns-artifact` | 470 passed |
| `cargo test -p lns-ipc` | 182 passed |
| `cargo test -p lns-cli` | 1020 unit tests; cucumber 587 scenarios / 2980 steps; `no_dead_steps` 8 passed — all green |
| `cargo test -p lns-service --lib` | 2722 passed |
| `cargo test -p lns-service --test behaviours` | 259 scenarios / 1032 steps, all green |
| `cargo clippy --all-targets -- -D warnings` on lns-artifact, lns-ipc, lns-cli, lns-service | clean |
| `cargo clippy --all-targets -- -D clippy::cognitive_complexity`, per crate, on the same four | clean |
| `make coverage` | **not run here** — the instrumented workspace does not fit in this container, so CI is the oracle. |

One coverage exemption changed reason and none grew: the
`crates/lns-service/src/containerfile/real.rs` entry now also names `RealDaemon`
(one `UnixStream` write-then-read-to-EOF) and the `DockerHost` impl beside it,
and says where each decision they feed is host-tested.

## CI

**No CI run was taken for this round, and here is exactly why.**

`.github/workflows/ci.yml` runs on `pull_request` only for PRs whose base is
`main`, so a stacked PR never reports on its own.

1. **This PR cannot be pointed at `main` any more.** `gh pr edit 418 --base main`
   answers `GraphQL: Cannot change the base branch because the pull request is
   part of a stack. (updatePullRequest)`. In the previous round the same edit was
   accepted and answered `CONFLICTING` with no run; now it is refused outright.
   The base is, and stayed, `feat/containerfile-arch`.
2. **The throwaway oracle could not run either.** `ci-oracle/containerfile-docker`
   — `main` with the whole #393 stack merged plus this branch — was updated with
   this round's commits (`e1079421`) and pushed, and the throwaway PR
   [#419](https://github.com/lensapp/lens-sandbox/pull/419) was closed and
   reopened to force the event. GitHub started no workflow, because #419 is
   `CONFLICTING` against today's `main` and GitHub runs no `pull_request`
   workflow for a PR it cannot merge.
3. **The conflict is `main`, not this branch.** `main` moved from `e4ad3579` to
   `aae90c1a` while this stack was in review, and one of those commits
   (`feat(cli): add JSON artifact push output`) rewrites the same `push` path
   that slices 4 to 7 rewrite: `push_formatted`, `push_dry_run_formatted` and a
   `--format` argument through `push_local`, against this stack's `PushPorts`,
   `ImageRequest` and index publishing. Merging `main` into the oracle is the
   stack's rebase, several hundred conflicting lines in
   `artifact/distribute.rs`, `artifact/real.rs` and the Layer 2 steps. I started
   it, saw the size, and aborted rather than invent an integration nobody would
   review on a throwaway branch.

**What this costs:** the coverage gate is unverified for this round. Every other
gate step was run here and is in the table above. The last green CI run of this
branch is
[34438461413](https://github.com/lensapp/lens-sandbox/actions/runs/34438461413),
which covers the state before this review round.

**What unblocks it:** rebasing the #393 stack onto `aae90c1a`, one slice at a
time, so `feat/containerfile-arch` and this branch merge into `main` cleanly
again. That is a decision for whoever owns the stack, not a change I make inside
a review round.

Auto-merge is not armed: a stacked PR's base carries no required checks, and
arming it merges the PR unreviewed.


chenhunghan (commented):
Reading review of slice 7 of lens-sandbox issue #393, against `feat/containerfile-arch` at dc9503e2. Nothing was run: no build, no tests, no daemon; the verifying Mac has no Docker, so the daemon path is not exercised there either, and the switch-off path is re-run on it after the round. Auto-merge is not armed. Verdict at the end.

# Findings, ranked

**1. The gate record comes from the switch that is set now, not from the image the cache answers with, so a daemon-built image is published and reported as gated.** `containerfile/real.rs:163` takes `outside_the_gate` from `request.args.build_engine`; both engines key an image over the same four inputs (`docker.rs:489` asserts that on purpose) and the cache `Entry` at `cache.rs:34` records no engine. Failure: `lns config set build.engine docker`, `lns sandbox build`, then `lns config set build.engine lns`, which is the last line of this PR's own verification recipe; `lns run .` computes the same key, the cache answers with the daemon-built image, and the run reports it as gated and a push publishes it without the annotation. Fix: record the engine on the cache entry and derive the disclosure from the entry a reuse answers with, or refuse to let one engine's key answer the other's build.

**2. The daemon resolves `FROM` itself, so the key stands on a digest the image may not have been built on.** `docker.rs:506` sends the Containerfile as written with no `pull`, while `docker.rs:489` keys on the digest lns peeked off the registry. Failure: the host's Docker store holds a month-old `node:24-bookworm`; lns keys on today's digest; the daemon builds on the old one; the published document names a base it never stood on and every later build reuses it. Fix: rewrite `FROM` to the digest-pinned reference `peek_base` returned before sending.

**3. A consumer running a pulled artifact is never told the gate did not apply.** The pull selects the index entry at `image/mod.rs:484`, but `PulledImage` at `:239` has no field for the annotation and no run-path line reads it; only `inspect` projects it. Failure: `lns run ghcr.io/team/agent:1.4.0` of an ungated image prints a summary that says nothing about the gate, against the spec paragraph this PR adds. Fix: carry the flag through `PulledImage` into the run's image line.

**4. `lns push` publishes the annotation and says nothing about it.** `distribute.rs:563,578` print digests and platforms only; the service line with the words drains into a frame channel at `real.rs:290`. Fix: print the phrase on the push's built line from `PushableImage.built_outside_the_gate`.

**5. `DOCKER_HOST` is read in the service process, so the socket depends on which environment started the service.** `real.rs:214` reads it in the tray-resident service; `cli-reference.md:514` promises it. Failure: a Colima user's build works after a shell-spawned service and refuses after a reboot. Fix: resolve the socket in the CLI where `build.engine` is read and send the path, or drop the promise from both documents.

**6. Every daemon build leaves a tagged image in the host daemon that lns never removes and `prune` cannot see.** `docker.rs:201` tags `lns-build.local/docker:<key>`; no `DELETE /images` is issued. Fix: delete after a successful export, or document the daemon's copy and the command that clears it.

**7. The build context is held three times in memory before a byte leaves the service.** `docker.rs:218` clones the tar into the request and `:100` copies it into the wire buffer. A 1 GiB context allocates about 3 GiB. Fix: stream, or build once and write without the copies.

**8. The new port method has a default answering `lns`, so no Layer 2 test can pin that push or `sandbox build` send the switch.** `service/client.rs:33`; only `RealSandboxService` overrides it; both production reads live in coverage-exempt files whose reasons were not updated. Fix: make it a required method.

**9. The reference guide's refusal is not the message that ships.** The refusal is wrapped with `building {label}:` and ends in `: No such file or directory (os error 2)`. Fix: quote what the code prints, or stop wrapping over the io error.

**10. On a two-architecture index the inspect note reads as if it applied to the whole line.** `artifact/mod.rs:657` appends the note inside the comma-joined list, and the scenario at `image_docker_engine.feature:41` asserts that ambiguous spelling. Fix: bind the note to its architecture.

Confirmed by reading: the three endpoints and their API version are as the body states; the exported `docker save` tar is read into the layer store through the existing import; the switch off touches no default-path code; four conventional commits with the spec change on its own; auto-merge not armed.

# Questions for the author

- If the `<TAG>-image` index tag is ever absent, `held_entries` starts empty and every other architecture's annotation is dropped while its digest survives. Should the annotation also ride on the per-architecture manifest?
- The spec still says "same key, same image"; two engines now answer one key. Is that the intent, or the defect in finding 1?
- Should a daemon build refuse a secret-shaped context file before the tar is sent, as `lns push` refuses it after?
- If the daemon answers 400 before draining a large body, does the user get the unreachable-daemon wording? Not verified.
- `Built.layers` counts new layers on the lns path and every exported layer on the daemon path. Deliberate?

# Verdict

The switch is small and behind its default, as the plan asked. Before merge: 1 (the disclosure is the whole point of the switch, and the PR's own recipe defeats it), 2, 3 and 4 (the three places a person reads the gate's status), 8 (nothing pins that the switch reaches the service). Can follow: 5, 6, 7, 9, 10. The daemon path remains unverified until someone with Docker runs the four commands in the body; the switch-off path is re-run on the Mac after the round.

chenhunghan:

PR #424, opened 2026-09-11
feat(service): serve the macOS guest network from a per-run gvproxy instead of the vmnet bridge

chenhunghan:
Refs #380

## The problem

On macOS an lns guest hangs off the Apple `vmnet` bridge
(`VZNATNetworkDeviceAttachment`). The guest gets its address from the host
DHCP server on `bridge100`, behind a shared content-filter flow. When that
lease does not arrive, the guest boots with no address, no route and no DNS,
and every egress rule, connector and approval card downstream of it never
comes into play. The review rounds on #380 kept landing on the same host
surfaces — `bridge100` readiness, a root-readable
`com.apple.vmnet.plist`, remembered leases.

## The design

Serve the guest's link from a per-run userspace network process instead of a
host bridge. This is how Podman machine, vfkit and libkrun already run guests.

- `lns-service` starts one `gvproxy` per run, beside the run's vsock socket in
  `~/.lns/runs/<id>/`.
- `gvproxy` binds a unixgram socket (`--listen-vfkit unixgram://<run-dir>/net.sock`).
- lns binds a second unixgram socket (`net-vm.sock`) in the same directory,
  connects it to gvproxy, and sends the four-byte `VFKT` datagram. gvproxy
  learns from that datagram which address to answer on; it consumes it rather
  than treating it as a frame.
- That socket is handed to Vz as `VZFileHandleNetworkDeviceAttachment`.
- gvproxy answers DHCP, DNS and NAT itself on `192.168.127.0/24`, gateway and
  resolver at `192.168.127.1`.

No `vmnet`, no `bridge100`, no host DHCP server, no shared content-filter flow.

The guest is unchanged. `lns-session-broker` still runs `ip link set eth0 up`
and `udhcpc`, and still writes `/etc/resolv.conf` from the DHCP-supplied
resolver (`crates/lns-session-broker/src/network.rs`). Nothing in the guest
pins a subnet: the only `192.168.*` literals in that crate are test fixtures.
The in-guest supervisor proxy, its transparent redirect and its DNS stub read
the gateway out of DHCP, so they need no change either.

## What changed

**`crates/lns-service/src/vm/netdev.rs` — the netdev abstraction.**
Backend selection, the socket layout in the run dir, the gvproxy argv, binary
lookup, checksum verification, the datagram handshake, and the
spawn → wait-for-socket → attach → reap sequencing. It follows the shape of
`vm/cloud_hypervisor/{launch,orchestrate,process}.rs`, which is how virtiofsd
is already spawned and reaped.

**`crates/lns-service/src/vm/netdev/real.rs` — the macOS composition root.**
Resolves the binary, spawns through `RealSpawner`, and hands back a
`NetDevice` guard. The guard holds the child and the socket, and kills the
backend when it drops. The guard is moved into the task that owns
`vm::boot`, so the backend lives exactly as long as the guest.

**`crates/lns-service/src/vm/vz/ffi.rs`** — `network_attachment` builds
`VZNATNetworkDeviceAttachment` or `VZFileHandleNetworkDeviceAttachment` from
the spec. The random locally administered MAC is kept. The `NSFileHandle` is
built with `closeOnDealloc: false`, because the netdev guard owns the
descriptor and closes it after the VM stops.

**Binary provisioning, with no brew or apt preflight.** In order:

1. `LNS_GVPROXY_BIN` — a wrong value refuses the run rather than falling back.
2. `PATH`.
3. The lns-managed copy at `~/.lns/gvproxy/0.8.9/gvproxy`, downloaded on first
   use from the pinned `gvisor-tap-vsock` v0.8.9 `gvproxy-darwin` asset and
   verified against `c6f7b4bc…40067c` from that release's `sha256sums`. This is
   the `src/kernel.rs` path: the same `download::ensure_pinned`, the same
   atomic install, the same refusal on a sha mismatch. `LNS_GVPROXY_CDN`
   overrides the release base, like `LNS_KERNEL_CDN`.

A missing or unverifiable gvproxy refuses the run before the workload, so
`lns run` exits `125` per `docs/cli-spec.md` §5.

**MTU.** `--mtu 1500`. `VZFileHandleNetworkDeviceAttachment` refuses an MTU
below 1500, and its socket-buffer contract (`SO_RCVBUF` at least twice
`SO_SNDBUF`, four times it for throughput) is met with 1 MiB / 4 MiB — the
same values vfkit and gvproxy use.

**`--ssh-port -1`.** gvproxy defaults to publishing `127.0.0.1:2222` on the
host, forwarded to `192.168.127.2:22`. Two concurrent runs would collide on
that port, and an lns guest runs no sshd. The forward is disabled.

**Rollout switch.** gvproxy is the default on macOS. `LNS_NETDEV=vmnet`
restores the previous `VZNAT` attachment for one release. An unknown value
refuses the run and names both choices.

**Banner and audit.** The launch prints one line —
`Network  gvproxy (192.168.127.0/24)` — and the run's audit chain gets a
`network` event carrying `lns_network_backend` and `lns_network_detail`, so
`lns audit` shows which backend a run used.

**The provisioner guest** (`tools/provisioner/real.rs`) gets the same netdev.
It pulls tools over the network, so it needed a link too.

## Port publishing still works

`spec.ports` / `-p` never used the network device. `crate::forward::establish`
binds the host listener and splices to the guest's `FORWARD_PORT` over the
vsock `GuestTransport` (`forward/real.rs`, `crates/lns-session-broker/src/forward/`).
Changing the netdev does not touch that path, and `lns run -p` keeps working
unchanged. The existing fake-forwarder tests in `forward/mod.rs` still pin it.

## How to verify

Unit tests, all in `crates/lns-service/src/vm/netdev.rs`, all host-runnable on
Linux CI:

- the default backend is gvproxy; `LNS_NETDEV=vmnet` and `LNS_NETDEV=gvproxy`
  select explicitly; an unknown value refuses and names both choices
- the socket layout lands in the run dir beside `vsock.sock`
- the gvproxy argv, exactly
- `LNS_GVPROXY_BIN` wins over `PATH`; a wrong one refuses; `PATH` wins over a
  download; nothing on the host leaves the managed copy to answer
- bytes that are not the pin never land at the managed path (wiremock)
- the `VFKT` handshake reaches a stand-in gvproxy listener, and it arrives
  *from* the VM socket's own path — which is the whole of what gvproxy needs
- a run dir too deep for a 104-byte `sun_path` refuses before anything spawns
- a backend that never listens, and a VM end that cannot attach, are both
  killed and reaped

Manual smoke, on a virt-capable macOS host:
`crates/lns-cli/tests/smoke/gvproxy-netdev.sh`. It boots a real guest and
curls a host through the guest proxy. It skips itself when gvproxy or Vz
cannot run, in the style of the existing `interactive-shell.exp` smoke. **The
master runs it on a macOS 15 host. No CI job runs it**, because no runner in
this repository can start a guest.

The first revision of that script could not pass: it named no definition, so it
read `./lns.yaml` from the workspace root, which does not exist, and a
definition with no egress rule would have raised an approval card for
`example.com` instead of reaching it. `1c3bd407` fixes it. It now writes a
sandbox definition into a temp project, the way `interactive-shell.exp` does:

```yaml
apiVersion: lns.run/v1
kind: sandbox
name: gvproxy-smoke-<pid>
spec:
  image: docker.io/curlimages/curl:8.11.1
  egress:
    http:
      - match: example.com
        verdict: allow
```

and runs it:

```bash
lns run --rm --yes --quiet --name gvproxy-smoke-<pid> \
  --entrypoint /bin/sh -f <project>/lns.yaml -- \
  -c 'curl -sS -o /dev/null -w "%{http_code} %{ssl_verify_result}" https://example.com; echo; cat /etc/resolv.conf'
```

It asserts `200 0` (the status and the TLS verify result, so the guest proxy's
CA is proven too), then `nameserver 192.168.127.1`, then
`lns audit <name>` — scoped to the run, not the whole chain. The image carries
a real `curl`: busybox `wget` over HTTPS fails through the guest proxy on both
backends, so it proves nothing here. `LNS_SMOKE_IMAGE` overrides the image. The
skip conditions are unchanged.

`lns audit` already resolved a run name, as
[cli-spec §2](docs/cli-spec.md) says a `RUN` may be. Its argument help named
only an id and a prefix, so this PR adds that word and the Layer 2 scenario
that pins it.

### Acceptance on macOS 15

The master accepted the branch on a macOS 15 Apple Silicon host —
[the acceptance report](https://github.com/lensapp/lens-sandbox/pull/424#issuecomment-5632359084).
What passed there:

- `alpine:3.20` guest: `eth0` took `192.168.127.3/24` by DHCP, default route
  via `192.168.127.1`, `/etc/resolv.conf` = `nameserver 192.168.127.1`, the
  supervisor's DNS stub on `upstream=192.168.127.1:53`. Banner:
  `✓ network gvproxy (192.168.127.0/24)`. Boot to session ready in about two
  seconds.
- `curlimages/curl:8.11.1` guest: `https://example.com` → `200`,
  `ssl_verify_result 0`; `https://api.anthropic.com/v1/models` → Anthropic's
  own `401 authentication_error`. A fetch of `deb.debian.org`, which the policy
  does not allow, raised an approval card, and answering it let the fetch
  through — the ask flow is intact on the new backend.
- `lns audit` shows `network  guest network served by gvproxy (192.168.127.0/24)`
  per run.
- No `gvproxy` process is left after a `--rm` run.
- Control: the same definition on the installed 0.25.0 (`vmnet`) booted with no
  address at the same minute. That is #380 happening live.

### What I could not run

I could not run a guest, and neither can CI. CI has macOS runners, but a
GitHub-hosted macOS runner cannot start a Virtualization.framework VM, so no
job in this repository boots a guest on macOS. Everything above the Vz call is
covered by the Linux unit tests; the Vz call itself and the live DHCP exchange
need the acceptance run on a real macOS 15 host.

My own sandbox could not compile `lns-service` at all (its GTK/atk system
libraries are not installable there), so every compile and test result for
this branch comes from CI, not from me. As of `457caea9` all required checks
are green: `lint`, `test`, `coverage`, `coverage-as-root`, `shell-tests-macos`,
`audit`, and both `check-release-build` jobs — including the macOS one, which
is what actually compiles the `VZFileHandleNetworkDeviceAttachment` call.

## Relation to #395

#395 is untouched and stays open. I did not rebase onto it, did not reuse its
host-address reservation, and did not reuse its plist observation.

One piece of it stays useful with either backend and is **not** in this PR:
the `ClientFrame::ConfigureNetwork` vsock frame. It removes the DHCP exchange
from every boot. It is not needed here — gvproxy's own DHCP server answers on
a socket that exists before the VM starts, so there is no bridge to wait for —
but landing it later would cut one round trip off each boot.

## Follow-up: a gvproxy orphaned by a dead service

`NetDevice::drop` kills gvproxy, and the child is spawned with
`kill_on_drop(true)`, so every ordinary stop reaps it. If `lns-service` is
killed outright, the run's gvproxy survives with `net.sock` still bound. The
next `lns start` of that run calls `NetLayout::remove_stale`, which unlinks
both sockets, so the run boots — but the old process lingers.

This is the treatment virtiofsd already gets, not a gap this PR opens.
`vm/cloud_hypervisor/process.rs` spawns virtiofsd with the same
`kill_on_drop(true)`, `launch.rs` has the same `remove_stale`, and
`run_registry::rebuild_from_records` registers each record as stopped without
looking for a live child of the dead service. There is no parent-death watch
and no sweep at service start for any child today, so gvproxy joins the
existing treatment exactly.

A sweep at service start belongs to every spawned child at once — virtiofsd,
cloud-hypervisor and gvproxy — and needs a record of each child's pid in the
run dir to be safe against a recycled pid. That is its own change, so it is a
follow-up:

- **Reap the children a killed service leaves behind.** Write each spawned
  child's pid beside its socket in the run dir, and at service start kill the
  ones whose run is not live. Covers virtiofsd and cloud-hypervisor as well as
  gvproxy.

## Follow-up: Linux

Linux is unchanged by this PR. A Linux guest still gets no `--net` device at
all; `cloud_hypervisor_args` passes none. The follow-up is passt over
vhost-user: `passt --vhost-user --socket <path>` binds the socket and
cloud-hypervisor connects to it with
`--net vhost_user=true,socket=<path>,vhost_mode=client,mac=…`. passt needs no
capabilities and no `/dev/net/tun`. It slots behind the same `netdev`
abstraction this PR adds.

## Notes for the reviewer

- `CHANGELOG.md` is generated by release-please from the conventional commit,
  so this PR does not edit it.
- One IGNORES entry is added, for `vm/netdev/real.rs`: it only compiles on
  macOS, and no coverage job runs there. Its reason names what is host-tested
  instead.
- `docs/sandbox-spec.md` and `docs/cli-spec.md` are not edited. Nothing here
  changes a settled decision: the document format is untouched, and exit code
  `125` is what §5 already says a pre-workload refusal costs.
- Auto-merge is deliberately not armed. The master merges after acceptance.


chenhunghan (commented):
Review from the master, read against `ead1676`, plus two checks run on a macOS 15 host.

**Verified**
- gvproxy v0.8.9 `gvproxy-darwin` sha256 `c6f7b4bc…40067c` matches the release's `sha256sums`.
- The exact argv from `gvproxy_args` (`--mtu 1500 --ssh-port -1 --listen-vfkit unixgram://… --log-file …`, no `--listen`) starts gvproxy on this host: it logs `waiting for clients...`, binds `net.sock`, stays alive. No control listener is needed.
- The `VFKT` datagram is a real gvproxy contract: `pkg/transport/unixgram_unix.go` peeks the first datagram, records the peer address, and consumes it when it equals `VFKT` ("If it's the old vfkit handshake, consume it").
- The netdev guard is moved into the task that owns `vm::boot` in both the run orchestrator and the tool provisioner, so gvproxy lives exactly as long as the guest. `lns start` reuses the run dir and `remove_stale` clears both sockets first. The Linux launch path only gains the `net: Nat` field.
- Port publishing is unaffected: `forward::establish` splices over the vsock transport, as the PR body says.

**Defect: `crates/lns-cli/tests/smoke/gvproxy-netdev.sh` cannot pass as written.**
`"$LNS" run --rm --name "$RUN_NAME" -- sh -c '…'` names no definition, so it runs `./lns.yaml` from the workspace root, and the repository has none; `lns run` refuses before booting. Even with one, a definition with no egress rule raises an approval card for `example.com`, which times out after 60 s into a plain network error, so the `curl` fails for a reason unrelated to the backend. `interactive-shell.exp` (line 40) shows the pattern: write a definition into a temp project first. The fix is a temp dir with an `lns.yaml` whose `egress.http` allows `example.com`, `lns run --rm --name … -f <that file> -- sh -c …`, and a `curl` that also proves the resolver: `getent`/`nslookup` do not work in the guest (raw UDP DNS is gated), so keep the `cat /etc/resolv.conf` check and let `curl` be the DNS proof. Please also make `lns audit "$RUN_NAME"` scope to the run rather than reading the whole chain.

**Question, no change required unless cheap:** `NetDevice::drop` sends `start_kill` to gvproxy, so a normal stop reaps it. If `lns-service` itself dies, the per-run gvproxy children are orphaned with their `net.sock` still bound; the next `lns start` of that run removes the stale socket, but the old process lingers. Is there an existing pattern for virtiofsd orphans (a prctl/`kqueue` parent-death watch, or a sweep at service start) that gvproxy should join? If not, a follow-up issue is fine.

**Acceptance** on this host happens once the coverage jobs are green and the smoke test is fixed: build the branch, run the second `lns-service` on its own `LNS_HOME`/`LNS_SOCKET_PATH`, boot a guest with an egress rule, `curl` through the guest proxy, read `lns audit`, and run one mahler dispatch that talks to `api.anthropic.com`. Leave the PR open after your fixes; the master merges after acceptance.

chenhunghan:
**Acceptance on macOS 15 (Apple Silicon), branch at `ead1676` + test commits, built with `cargo build --release -p lns-cli -p lns-service` and signed with `crates/lns-cli/scripts/codesign-macos.sh`, run as a second `lns-service` on its own `LNS_HOME`/`LNS_SOCKET_PATH`, `LNS_GVPROXY_BIN` pointing at the verified v0.8.9 binary.**

Passed:
- `alpine:3.20` guest, definition allowing `example.com`: `eth0` got `192.168.127.3/24` by DHCP, default route via `192.168.127.1`, `/etc/resolv.conf` = `nameserver 192.168.127.1`; the supervisor's DNS stub reports `upstream=192.168.127.1:53`. Launch banner: `✓ network gvproxy (192.168.127.0/24)`. Boot to session ready in about two seconds.
- `curlimages/curl:8.11.1` guest: `curl https://example.com` → `200`, `ssl_verify_result 0` through the guest proxy's CA; `curl https://api.anthropic.com/v1/models` → Anthropic's own `401 {"type":"error","error":{"type":"authentication_error",…}}`, so TLS interception and egress to a model provider work; a fetch of `deb.debian.org`, absent from the policy, raised an approval card in the tray, and the answer let it through: the ask flow is intact on the new backend.
- `lns audit` shows `network  guest network served by gvproxy (192.168.127.0/24)` for each run.
- No `gvproxy` process left after each `--rm` run (`pgrep` finds none).
- Control: the same definition on the installed 0.25.0 (`vmnet`) at the same minute booted with **no address** (`inet` line absent, fallback resolvers `1.1.1.1`/`8.8.8.8`, `http://example.com` → `502` from the guest proxy). That is #380 happening live; the gvproxy guest beside it came up cleanly.

Noted, not this PR's fault: busybox `wget` over HTTPS through the guest proxy fails with `error getting response` on both backends; use `curl` in the smoke test.

Still needed before merge: the smoke test fix from the review above (it runs with no definition and no egress rule). After that this PR is accepted from my side.

chenhunghan:
Smoke test at `1c3bd40` run on the macOS 15 host against the branch build (second service, `LNS_GVPROXY_BIN` set): `[pass] the guest reached example.com over TLS through the guest proxy`, `[pass] the guest resolver is the gvproxy gateway`, `[pass] the chain names gvproxy for this run`, `[ok]`. No gvproxy process left afterwards. Accepted from my side; merging once this round's CI is green and the PR body carries the acceptance.

chenhunghan:
## Review: does this do what vfkit/krunkit and podman machine do?

**The wiring: yes, exactly.** I compared `crates/lns-service/src/vm/netdev.rs` and `vz/ffi.rs` against vfkit's `pkg/vf/virtionet.go`, krunkit's `virtio-net,unixSocketPath=…` device and podman's `pkg/machine/apple/apple.go`:

- gvproxy is started with `--listen-vfkit unixgram://<sock>` and the VM end is a bound unixgram socket connected to it, with the `VFKT` magic sent first so gvproxy learns the peer address — the same handshake vfkit does and krunkit calls `vfkitMagic`.
- `SO_SNDBUF` 1 MiB and `SO_RCVBUF` 4 MiB are the values vfkit sets; `--mtu 1500` matches gvproxy's default and the Vz minimum.
- The fd goes into `VZFileHandleNetworkDeviceAttachment` via an `NSFileHandle` that does not close on dealloc, with the netdev guard owning the lifetime; libkrun documents the same fd-ownership contract for its `unixgram_fd` device.
- podman waits for the socket with a bounded backoff and kills gvproxy through its pid file; here the child is held by the guard and killed on drop, which is equivalent for a per-run process (the orphan-on-service-kill follow-up stands).

Two things are **not** what podman does, and both need a change before gvproxy becomes the default.

### 1. Blocker: a host with podman installed cannot boot any guest

`located_gvproxy` prefers any `gvproxy` on `PATH` over the pinned copy. podman 4.9 ships `/opt/podman/bin/gvproxy` (build `012bc90`, Jan 2024), which rejects the `--ssh-port -1` this PR passes:

```
$ /opt/podman/bin/gvproxy --mtu 1500 --ssh-port -1 --listen-vfkit unixgram://…/net.sock --log-file …/net.log
level=error msg="ssh-port value must be between 1024 and 65535"
```

Every run then fails after 10 s with `gvproxy did not expose its vfkit socket`, and the run dir holds no `net.log` to say why (reproduced twice on this host; the pinned v0.8.9 accepts `-1`). Fix: use the lns-managed pinned binary unless `LNS_GVPROXY_BIN` is set explicitly — a `PATH` hit is podman's build far more often than a compatible one — and when gvproxy exits before the socket appears, include its exit status and the tail of `net.log` in the error. The installer's "Using the gvproxy already on this host" branch has the same problem.

### 2. The gvproxy topology podman wants is one a sandbox must not have

podman's VM is trusted, so gvproxy deliberately gives the guest two things (`cmd/gvproxy/config.go` and `main.go` in v0.8.9, not switchable by flag):

- an HTTP control API on the gateway, `192.168.127.1:80`, with `/services/forwarder/expose`, which binds a **host** TCP port and forwards it into the guest;
- a NAT entry `192.168.127.254 → 127.0.0.1`, so the guest reaches services bound to the host's loopback.

Neither existed on the vmnet bridge, and the guest cage does not stop them. Verified on this PR build (curl image, egress `example.com` and `api.anthropic.com` only):

```
$ curl http://192.168.127.1/services/forwarder/all          → 200 []
$ curl http://192.168.127.254:47123/                        → 200 HOST-LOOPBACK-SERVICE   (a python server bound to 127.0.0.1 on the host)
$ curl -X POST http://192.168.127.1/services/forwarder/expose -d '{"local":"127.0.0.1:47124","remote":"192.168.127.3:8000"}' → 200
host$ curl http://127.0.0.1:47124/                          → FROM-GUEST!   (lsof: gvproxy LISTEN 127.0.0.1:47124)
```

`lns audit` recorded each as `GET 192.168.127.1:80 — allowed always` plus an `allow-always … needs your decision` card, i.e. the request went through before the person answered. With a `match: "*"` policy, the common shape for agent sandboxes, there is no card at all. So with this PR as-is, any sandbox can open ports on the host and talk to anything listening on the host's loopback.

Suggested fix, in order of preference:
- Deny the guest's own subnet as an egress destination in the cage, the same way the proxy already floors loopback and link-local (`proxy.rs` "CONNECT to loopback/link-local/unspecified denied" in lens-sandbox-core), letting only DHCP and the resolver path reach `192.168.127.1`. That closes both the forwarder API and the `.254` NAT, whatever the egress rules say.
- Independently, start gvproxy from a `--config` file whose `nat` maps the virtual IP somewhere unroutable, so the loopback bridge is gone even if the cage is off (a non-root supervisor already means no cage).

Everything else in the earlier acceptance still holds: DHCP lease, resolver, TLS egress, cards for unlisted hosts, no orphan gvproxy after a clean stop.

chenhunghan:
Independent confirmation from the parity harness (PR #428, case `loopback-witness`): a guest on this branch's gvproxy backend reached a listener on the host's 127.0.0.1 through the `.254` NAT (one accept recorded), where the netstack of PR #426 and the vmnet bridge record zero. This PR stays a held fallback.

PR #425, opened 2026-09-11
feat(service): serve the Linux guest network from a per-run passt over vhost-user

chenhunghan:
## The problem

On Linux a guest has no network device at all. `crates/lns-service/src/vm/cloud_hypervisor/launch.rs` passed cloud-hypervisor `--api-socket --cpus --memory --kernel --initramfs --cmdline --console --serial --disk --fs --vsock` and no `--net`. The guest booted and `exec` worked, but the broker's `ip link set eth0 up` + `udhcpc` failed best-effort, so nothing in the sandbox could reach a model provider — and the egress policy, connector injection and approval cards never came into play on Linux at all.

## The design

One `passt` per run, started beside the run's vsock socket and stopped with it, exactly where #424 starts `gvproxy` on macOS:

```
passt --vhost-user --socket <run-dir>/net.sock --foreground \
      --pid <run-dir>/net.pid --log-file <run-dir>/net.log \
      --ipv4-only --mtu 1500 \
      --address 192.168.127.2 --netmask 24 --gateway 192.168.127.1 \
      --dns 192.168.127.1 --dns-forward 192.168.127.1 --no-map-gw \
      --tcp-ports none --udp-ports none
```

and then, in the cloud-hypervisor argv:

```
--net vhost_user=true,socket=<run-dir>/net.sock,mac=<random locally administered MAC>
```

`passt` binds the socket; cloud-hypervisor's default `vhost_mode=client` connects to it, so lns waits for the socket to appear before launching the VMM — the same spawn → wait → connect → reap sequencing the gvproxy backend already had, now shared. `--memory size=…M,shared=on` was already there for vhost-user-fs and is what vhost-user-net needs too; a test pins that the two travel together.

**Addressing.** `passt` hands the guest the host's own address and gateway by default, which leaves a guest unable to tell host traffic from its own and makes every host's sandbox read differently. lns asks for `192.168.127.2/24`, gateway and resolver on `192.168.127.1` — the subnet the macOS backend serves, so a guest and the docs read the same on both platforms. `--dns-forward` puts the resolver on an address `passt` itself answers (the stub the guest is handed over DHCP must be one that replies), and `--no-map-gw` stops the guest reaching the host's own loopback services by aiming at its gateway. The guest is unchanged: `eth0` up, `udhcpc`, the in-guest proxy, the nftables cage.

**Provisioning, no system dependency added lightly.** `passt` publishes no versioned, checksum-pinnable static binary the way `gvproxy-darwin` is pinned in #424 — I found nothing to pin (see "what I could not run"), so lns does not download one. It looks at `LNS_PASST_BIN`, then `PATH`, then `/usr/bin`, `/usr/local/bin`, `/usr/sbin`, `/usr/local/sbin`, and a host without one is refused **before the workload** with exit `125` and a message naming the package and the override. `scripts/lns-install/lns-install.sh` probes for it next to `cloud-hypervisor` and `virtiofsd`.

**Rollout.** `passt` is the default on Linux. `LNS_NETDEV=none` keeps today's no-network behaviour for one release; `vmnet` stays macOS-only. The `Backend` enum from #424 grew `Passt` and `Off`, and now takes the host whose backends a run may choose from, so a name the other platform serves is refused by name with this platform's choices. The launch banner and the run's audit chain name the backend, as in #424.

## Relation to #424 and #380

**Stacked on #424** (`feat/380-gvproxy-netdev`) and opened against that branch — it extends the netdev abstraction #424 introduced rather than duplicating it: one `Backend`, one `NetLayout`, one spawn/wait/reap path, one banner and audit line. Please do not merge this before #424; it should be retargeted to `main` once #424 lands and this has been accepted on a KVM host.

Refs #380

## Container requirements (issue #380's open question)

`passt` isolates itself in a user namespace as it starts. Where `unshare(2)` is blocked — Docker's default seccomp profile blocks it without `CAP_SYS_ADMIN` — it logs `Couldn't create user namespace: Operation not permitted`. **This is fatal in passt: its isolation step dies rather than continuing degraded, and there is no option to turn the isolation off.** So a container hosting `lns-service` needs `unshare(2)` allowed (`--security-opt seccomp=unconfined`, or `--cap-add SYS_ADMIN`); Podman's default profile already allows it. That is now documented in `docs/service.md`. Stated plainly: I read this from passt's documented isolation behaviour — **I could not run passt to confirm it** (see below), and the failure mode lns shows either way is a refused run, not a guest with a broken network.

## What I could and could not run

- **Ran:** the unit tests for the argv, the socket layout, the binary lookup, the refusal message, the MAC, the backend selection and the spawn/wait/reap sequencing. My sandbox cannot build `lns-service` itself (no GTK/pango dev packages and no network to install them), so I compiled `vm/netdev.rs` verbatim in a scratch crate with stubbed `Spawner`/`which` and ran its suite there: 49 tests, all green, including every new one. CI is the real signal for the workspace build, `make lint`, `make complexity` and the coverage gate — I'm watching it.
- **Could not run:** anything that boots a guest. No `/dev/kvm` in CI or in my sandbox, and GitHub-hosted Linux runners have no nested virtualisation, so `passt` + cloud-hypervisor with the lns kernel and initramfs is **untested end to end**.
- **Could not check:** `passt --help` itself. The environment has no outbound network and no `passt` package, so every flag here comes from passt's documented option set, not from a binary I ran. The two flags most worth a second pair of eyes on a real host are `--no-map-gw` (if `--dns-forward` turned out to depend on gateway mapping, guest DNS would fail and dropping `--no-map-gw` is the fix) and `--tcp-ports none` / `--udp-ports none`.
- **Could not verify:** that no maintained, checksum-pinnable static `passt` build exists. I could not reach passt.top or any distro index from here. If one does exist, pinning it the way `gvproxy-darwin` is pinned would remove the system dependency, and I'd rather do that than ship the refusal.

## How to accept this on a KVM host

```bash
sudo apt install passt          # or dnf install passt
make build && ./bin/lns service start
bash crates/lns-cli/tests/smoke/passt-netdev.sh
```

The smoke script (new, beside `gvproxy-netdev.sh`; skips itself with exit 0 where `/dev/kvm` or `passt` is missing, and is run by hand because no GitHub-hosted runner can boot the guest) asserts the four contracts: a 200 from `https://example.com` through the in-guest proxy, `inet 192.168.127.2` on `eth0`, `nameserver 192.168.127.1` in the guest's `/etc/resolv.conf`, and a `guest network served by passt` line in `lns audit`. Worth also running `LNS_NETDEV=none lns run -- echo hi` (boots with no device, as before) and `lns run` on a host with `passt` removed (refused with exit 125 before the workload).

## Docs

`docs/service.md` grew the Linux half of "The guest network" — where `passt` comes from, the addressing and why, the container requirement — and `docs/getting-started.md` names `passt` in the Linux prerequisites. No changes to `docs/sandbox-spec.md` or `docs/cli-spec.md`: nothing here changes a settled decision. The CHANGELOG is generated by release-please from the conventional commit, so there is no hand-written entry.


chenhunghan (commented):
Review from the master, with the checks the PR body asked for, run on a macOS host inside Debian trixie containers (`passt 0.0~git20250503.587980c-2+deb13u1`, the same package the deploy image would carry). No guest was booted; that needs a KVM host.

**Verified**
- The exact `passt_args` argv (`--vhost-user --socket … --foreground --pid … --log-file … --ipv4-only --mtu 1500 --address 192.168.127.2 --netmask 24 --gateway 192.168.127.1 --dns 192.168.127.1 --dns-forward 192.168.127.1 --no-map-gw --tcp-ports none --udp-ports none`) is accepted by passt: no option is rejected, `net.sock` and `net.pid` appear, and the log reads `Template interface: eth0 (IPv4)`, `DHCP: assign: 192.168.127.2`. So `--netmask 24` as a prefix length, `--tcp-ports none` and `--udp-ports none` are all fine.
- Under Docker's default seccomp profile passt exits after `Couldn't create user namespace: Operation not permitted`, with the socket already bound; with `--security-opt seccomp=unconfined` it runs. The container requirement in `docs/service.md` is correct and the failure is fatal, not degraded, as the body says.
- passt.top publishes builds only for x86_64 (`passt`, `passt.avx2`, `.deb`, `.rpm`) with no checksum file, so there is nothing to pin the way `gvproxy-darwin` is pinned. Refusing with the package names is the right call for now.
- cloud-hypervisor v53 `--net vhost_user=true,socket=…,mac=…` matches its `--help`; default `vhost_mode=client` connects to the socket passt binds, so the roles fit.

**To settle on a KVM host at acceptance (nothing to change now):** whether `--dns-forward 192.168.127.1` together with `--no-map-gw` lets the guest's DNS stub resolve; if not, drop `--no-map-gw`. And whether the supervisor's nftables cage plus passt's `--tcp-ports none` leave any path the guest needs (host-side forwards for `-p` ride vsock, so I expect none).

**Nit:** the crate attribute became `#![cfg_attr(any(target_os = "macos", target_os = "linux"), allow(dead_code))]`, which now silences dead-code warnings on the two platforms the module actually targets. The previous form allowed them only where the module is compiled but unused. Consider `#![cfg_attr(not(any(target_os = "macos", target_os = "linux")), allow(dead_code))]` plus per-item `cfg` where a helper is genuinely platform-specific, so a real dead path still warns.

This PR stays open on top of #424; once #424 merges I retarget it to `main`. Acceptance needs a Linux host with `/dev/kvm`, passt and cloud-hypervisor; `crates/lns-cli/tests/smoke/passt-netdev.sh` is the script to run there.

chenhunghan:
Verification update: the per-file coverage gate now passes locally for the touched crate — `COVERAGE_CRATES="lns-service" make coverage` reports `[crates/lns-service/] 120 at 100%, 32 ignored, 0 failures`, with `vm/netdev.rs` at 669/669 and `vm/cloud_hypervisor/launch.rs` at 274/274. `cargo clippy -p lns-service --all-targets -- -D warnings` is clean and `cargo test -p lns-service --all-targets` is green (2561 + 8 tests).

Note that CI reports no checks here: `.github/workflows/ci.yml` triggers only on pull requests targeting `main`, so a stacked PR based on `feat/380-gvproxy-netdev` gets no runs until it is retargeted after #424 lands. Still unrun anywhere: anything needing `/dev/kvm` or a real `passt` binary — see the acceptance steps in the description.

chenhunghan:
Status: on hold, not to be merged.

Direction change after the review on #424 (https://github.com/lensapp/lens-sandbox/pull/424#issuecomment-5633495035): the macOS backend is being rewritten as an in-process userspace netstack inside lns-service (no external binary; the host boundary is decided by our own policy). Linux will get the same stack behind a vhost-user-net front end for cloud-hypervisor, so passt is no longer the plan. This PR stays open as the fallback until that front end exists; if it is merged after all, it must pass `--no-map-gw` so the gateway does not map to the host.

PR #426, opened 2026-09-11
feat(service): serve the macOS guest network from an in-process userspace netstack instead of the vmnet bridge

chenhunghan:
## The problem

Issue #380: on macOS a guest gets its address from the host's `vmnet` DHCP
server over the Apple NAT bridge. That server is shared, it needs
`bridge100` to exist first, and its state lives in a root-readable plist.
A guest that does not get a lease has no network and the run is useless.

## Why in-process, not gvproxy

PR #424 served the same link from a per-run `gvproxy`. The review of that PR
(https://github.com/lensapp/lens-sandbox/pull/424#issuecomment-5633495035)
found two defects:

1. **An external binary decided whether a run could boot.** `gvproxy` was
   looked up on `PATH`, so a host with podman installed picked podman's
   build, which refuses the `--ssh-port -1` lns passes, and every run failed
   after a 10-second timeout.
2. **The guest could reach the host through the gateway.** gvproxy serves an
   HTTP control API on `192.168.127.1:80` with
   `/services/forwarder/expose`, which binds a **host** TCP port, and a NAT
   entry `192.168.127.254 -> 127.0.0.1`. Neither existed on the vmnet
   bridge. With `match: "*"` egress, the common shape for an agent sandbox,
   there was no approval card either.

Both are impossible here. There is no binary to find: the stack is part of
lns-service. There is no control API of any kind and no NAT entry to the
host's loopback: the guest's own subnet is refused whole, except the
resolver on 53 — UDP and TCP — and ICMP echo at the gateway.

## The design

**1. Transport.** `socketpair(AF_UNIX, SOCK_DGRAM, 0)`. One end goes to Vz as
`VZFileHandleNetworkDeviceAttachment` through a non-closing `NSFileHandle`
(unchanged from #424); the other is a `tokio::net::UnixDatagram` owned by the
run. `SO_SNDBUF` 1 MiB and `SO_RCVBUF` 4 MiB, the values vfkit uses. One
datagram is one ethernet frame. MTU 1500 both ways; read buffers are 1518
bytes, so a full frame with a VLAN tag is never truncated. Nothing touches
the filesystem.

**2. Layer 2** (`netdev/link.rs`). Ethernet II through `etherparse`. The
guest's MAC is learned from the first frame it sends. ARP for the gateway is
answered with `0e:6c:6e:73:00:01` — locally administered, `6c 6e 73` spells
"lns", `01` is the gateway. IPv6, VLAN-tagged frames and every other
ethertype are dropped and counted.

**3. DHCP** (`netdev/dhcp.rs`). The wire format is smoltcp's own DHCPv4
types: `DhcpPacket` and `DhcpRepr` read the guest's message and write the
reply. smoltcp ships only a client, so the server side is ours — but only the
one-lease policy, not the parser. One lease, no pool: `192.168.127.2`, mask
`255.255.255.0`, router and DNS `192.168.127.1`, 24 h. DISCOVER→OFFER,
REQUEST→ACK, and a bare REQUEST (renew) is acknowledged too. Options 1, 3, 6,
51, 53 and 54; the broadcast flag decides where the reply goes. A client
asking for an address this link does not lease gets a NAK rather than
silence.

**4. Layer 3** (`netdev/engine.rs`). `netstack-smoltcp` 0.2.4 turns the
guest's IPv4 packets into a `TcpListener` of streams and a `UdpSocket` of
datagrams.

**Admission before allocation.** The library creates a smoltcp socket, two
buffers and two ring buffers — about 256 KiB — for every segment that carries
SYN without ACK, and looks nothing up by four-tuple. So the engine keeps the
flows a guest holds and decides per SYN, at ingress, before the stack sees
the packet: a SYN for a flow it already has is dropped and counted (smoltcp
retransmits its own SYN-ACK from SYN-RECEIVED, so nothing is lost); a SYN for
a new flow when the table is full is answered with the same reset a refusal
gets, and counted; only an admitted SYN is forwarded, and its permit is taken
there rather than after the stream arrives. A flow leaves the table when its
relay ends, and a flow the guest never finished opening leaves after 30 s,
which ends its relay and drops its stream.

**What the engine reads is what the stack reads.** An ingress packet is
parsed with smoltcp's own `Ipv4Packet::new_checked` and `UdpPacket::new_checked`
— the parsers the library uses — and trimmed to the length its IPv4 header
states, because a frame padded to the 60-byte ethernet minimum carries
trailing bytes. A segment keeps etherparse as its one parser: its check and
smoltcp's `TcpPacket::new_checked` accept exactly the same bytes, and a test
pins that they agree. As a seatbelt, an item the library will not hand over
is one unusable item, not the end of the stream: the UDP and TCP accept loops
read on until the run is cancelled.

**5. Host boundary** (`netdev/policy.rs`). Every destination is decided
before a host socket is opened — in fact before the stack sees the packet.

| Destination | Result |
|---|---|
| `192.168.127.1:53` UDP | The gateway relays the query, fitted to what the guest can take. |
| `192.168.127.1:53` TCP | The gateway relays the query, length-prefixed per RFC 1035 §4.2.2. |
| `192.168.127.1`, ICMP echo | The gateway answers. |
| Anything else in `192.168.127.0/24` | Refused. |
| `127.0.0.0/8` | Refused. |
| `0.0.0.0/8` | Refused. |
| `169.254.0.0/16` | Refused. |
| `224.0.0.0/4` | Refused. |
| `255.255.255.255` | Refused. |
| Not IPv4 | Refused. |
| Everything else, including the host's LAN | Carried, as on vmnet. |

A refused TCP segment is answered with a reset (RFC 793: from the segment's
own ACK when it carries one, from zero when it does not), and a refused
datagram with an ICMP port-unreachable quoting the header the guest sent, so
the guest fails at once instead of waiting out a timeout. Each refused
destination is logged once at debug level, up to 1024 distinct destinations
per run; past that the run says once that it counts the rest without logging
them. Every refusal is counted, logged or not. The
policy is a value (`Boundary`) passed into the stack; tests use
`Boundary::Permissive`, and a dedicated test asserts the default refuses each
range above.

**6. DNS.** The gateway is a plain relay: it resolves nothing itself and
caches nothing, so the host's answers, TTLs and negative answers are what the
guest gets. The resolver list is the host's own — the nameservers of
`/etc/resolv.conf` plus the per-domain resolvers `scutil --dns` reports,
which is where a split-DNS VPN's internal resolvers live and where
`resolv.conf` never has them. A resolver is asked on the port its own block
names, 53 when it names none. Every resolver whose domain is the longest
match answers a name, lowest `order :` first and the rest as its fallback,
per resolver(5); a name no domain covers goes to the default resolvers.

The list is read again every 30 s, and again after any query nobody
answered, so a VPN coming up mid-run is picked up. A query is answered from
the list in hand and waits for no read: one refresh at a time runs on
`spawn_blocking`, with 2 s for `/etc/resolv.conf` and `scutil --dns`
together, and a read that outlasts that leaves the list in hand in place.

**The search fits under the guest's own deadline.** The guest's stub gives up
after 2 s (lens-sandbox-core `dns::UPSTREAM_TIMEOUT`, which matches glibc), so
the gateway hedges rather than waiting: it asks the first server, asks the
next after 700 ms while the first stays pending, and takes the first usable
answer. SERVFAIL, REFUSED and NOTIMP are a server saying "not me" and move to
the next server at once; NOERROR and NXDOMAIN are answers and come back as
they are. A truncated answer is asked again over TCP to the same server. The
whole search has 1800 ms, then the last upstream reply, else a SERVFAIL of
the gateway's own — never silence. A server that answered nothing is asked
last for 30 s, so the next query is not slowed by it again.
`hickory-proto` does the wire format; it was already in `Cargo.lock`.

**The answer is fitted to the guest.** Over UDP the gateway sends at most
what the guest said it can take — the payload size of the query's OPT record,
else 512 bytes — and never more than 1472, the MTU less an IPv4 and a UDP
header, because this link does not fragment and a bigger frame dies at the
guest's virtio-net MTU. An answer that does not fit comes back as its own
question with TC set and the answer, authority and additional sections
emptied, per RFC 2181 §9, keeping the OPT record the guest sent. The fitted
form carries the guest's own question, so it is never larger than the query
the guest sent in one datagram of this link.

**And the guest has somewhere to ask again.** The gateway serves DNS over TCP
on `192.168.127.1:53`: a SYN for the resolver is admitted the way a datagram
for it is, and the stream is served by a length-prefixed loop (RFC 1035
§4.2.2) — one or more queries per connection, the same relay and the same
search budget, closed after 10 s idle, counted under the DNS allowance rather
than carried to a host socket. Every other gateway port stays refused.

**7. Bounds and limits.** Per guest, per run:

| | |
|---|---|
| Concurrent TCP flows | 1024 |
| Concurrent UDP flows | 512 |
| DNS queries in flight | 256 |
| DNS connections over TCP to the gateway | 64, each carrying at most 64 queries, closed after 10 s idle or a 5 s unread answer |
| TCP connect timeout | 10 s |
| TCP buffer per direction per flow | 64 KiB (decided 2026-09-12, `916b7e8a`) |
| UDP flow idle timeout | 60 s |
| Half-open TCP flow timeout | 30 s |
| DNS connection idle timeout | 10 s |
| One DNS search, every server included | 1800 ms |
| A DNS server silent this long is hedged past | 700 ms |
| A DNS server that answered nothing is asked last for | 30 s |
| Reading the host's resolver list | 2 s |
| Largest DNS answer over UDP | 1472 bytes |
| Frames queued between the device and the stack | 512 |
| MTU | 1500, both ways, no reassembly |
| One frame on the wire | 1518 bytes: the MTU, the ethernet header, a VLAN tag |
| Refused destinations logged per run | 1024, then counted only |

Anything over a limit is dropped and counted at debug level rather than
queued. A guest that outruns the stack loses frames, as it would on a busy
wire. `engine::MAX_FRAME` is the one frame size of the link: the device
reads into it and the stack is built from the MTU it derives from.

**8. Lifecycle.** The netdev starts before the VM and refuses the run
(pre-workload, exit 125 through the existing error path) if it cannot come
up. One guard owns every task of the run: the listeners live in a `JoinSet`,
and every relay a listener starts — TCP, UDP and DNS — is spawned through a
`tokio_util::task::TaskTracker` and selects on a `CancellationToken`.
Dropping the guard cancels the token, closes the tracker and aborts the
`JoinSet`, so a run stopped mid-transfer stops copying at once. No processes
to reap. One stack per run, no shared state.

**The guard's life is the VMM's life.** `vm::boot_with_attachments` gives what
a run holds — the netdev guard and the run's volume leases — to the blocking
closure that runs the VMM, so they are dropped exactly when `backend.run`
returns. Aborting the task that awaits the boot, which the orchestrator does
when no connector arrives and the provisioner does when a stop did not work,
no longer cancels the stack under a guest that is still running. The host end
of the socket pair closes with the guard; the VM end was handed to Vz as an
NSFileHandle with `closeOnDealloc: false`, so nothing else closes it.

**9. Rollout.** Default on macOS. `LNS_NETDEV=vmnet` restores VZNAT for one
release; `LNS_NETDEV=netstack` names the default. `LNS_GUEST_SUBNET=<a /24>`
moves the link for a host that already uses `192.168.127.0/24`; the gateway
is always `.1`, the guest `.2`, and the boundary follows. A host address
inside the guest subnet is named at startup, because no guest can reach it.
The launch banner, the audit event and `lns audit` say `netstack` and the
subnet in force. The stack is reached through a "frames in, frames out"
channel pair (`engine::Frames`), so a vhost-user front end for
cloud-hypervisor can be added without touching the stack. The Linux path is
untouched.

Both variables are read by lns-service from its own environment, and neither
travels over the IPC a run is asked for on, so they are set where the service
starts — `lns service stop`, then `lns service start` with the variable in
its environment. `lns run` prints one warning line when it sees either in its
own environment. `docs/service.md` shows both routes.

**10. Port publishing** is unaffected. `spec.ports` / `-p` travels over the
run's vsock channel (`forward/real.rs`), not over this link; its tests are
unchanged and still pass.

## Removed

gvproxy download and checksum code, `LNS_GVPROXY_BIN`, `LNS_GVPROXY_CDN`, the
installer's gvproxy section, gvproxy in the docs, and the run-dir socket
paths (`net.sock`, `net-vm.sock`) with their path-length and stale-socket
handling. A socket pair needs none of it.

## Review amendments

The six amendments in the review comment are all in, at `66c1b843`:

| Amendment | Where |
|---|---|
| 1. DNS is a plain relay, one path | `netdev/dns.rs` (relay, scopes, longest-suffix with every tying resolver kept in `order`, 30 s refresh + refresh-on-failure off the workers, hedged search of 700 ms / 1800 ms, TC→TCP, SERVFAIL, no cache) and `netdev/resolver.rs` (`/etc/resolv.conf` + `scutil --dns`, the UDP and TCP transports). 256 in-flight per guest in `netdev/engine.rs`. `getaddrinfo` synthesis is gone. |
| 2. Bounds per guest | `engine::Limits` / `engine::Allowance`: 1024 TCP, 512 UDP, 256 DNS, 10 s connect, 60 s UDP idle, 64 KiB per direction, a 512-frame bounded queue that drops and counts. One guard per run owns the listeners (a `JoinSet`) and every relay (a `TaskTracker` and a `CancellationToken`). Tests: the 1025th TCP flow, the 513th UDP flow and the 257th query are refused, and dropping the guard aborts every task, an open relay included. |
| 3. Subnet override | `netdev::Subnet::from_env` (`LNS_GUEST_SUBNET`, a `/24` only, gateway `.1`, guest `.2`, boundary derived) and `netdev::overlap_warning`, logged at startup from `netdev/real.rs` over `getifaddrs`. Documented in `docs/service.md`. |
| 4. Packet boundary, explicit | ARP and IPv4 handled; IPv6, VLAN (`0x8100`) and every other ethertype dropped and counted; fragments dropped, never reassembled; ICMP echo at the gateway answered; refused TCP reset, refused UDP answered with ICMP port-unreachable. "IPv4 only" is stated in the docs. Malformed-input sweeps over truncated ethernet, ARP, IPv4, UDP, TCP, ICMP and DHCP assert nothing panics and nothing half-read is carried. |
| 5. Smoke test additions | `netstack-netdev.sh` now runs a host listener that logs every accept and fails if it sees one, checks `127.0.0.1:<port>` as well as `192.168.127.1` and `192.168.127.254:<port>`, and adds a dual-stack `https://example.com` fetch that reports the address actually used. |
| 6. PR body | The "Bounds and limits" table and the "DNS" paragraph are in the design section above. |

## Review at 6e29f21

Three items, one commit each:

| Item | Commit | What changed |
|---|---|---|
| 1. Relay tasks escape the guard | `6ab20368` | `Running` holds a `TaskTracker` and a `CancellationToken`. `carry_tcp`, `carry_udp` and `relay_query` are spawned through the tracker and select on the token. `Drop` cancels the token, closes the tracker and aborts the `JoinSet`. The guard test opens a TCP flow to a test listener first. |
| 2. The refusal log set is unbounded | `df78bc17` | `note_refusal` logs at most 1024 distinct destinations, then says once that it counts the rest without logging them. `Counters` counts every refusal either way. |
| 3. Two `MAX_FRAME` constants | `e0a2abac` | One `engine::MAX_FRAME` — the MTU, the ethernet header and a VLAN tag, 1518 — used by the device reader and derived from the MTU the stack is built with. |

`tokio-util` is pinned at `0.7.18` with `default-features = false` and the
`rt` feature. It was already in `Cargo.lock` as a transitive dependency.

## Review at c46ea5a9, DNS round

Five DNS findings of the second review, one commit each. This round is the
relay only: `netdev/dns.rs`, `netdev/resolver.rs` and their caller in
`netdev/engine.rs`.

| Finding | Commit | What it does now |
|---|---|---|
| 3. Resolver refresh blocked runtime workers under a mutex (P2) | `4eb73523` | `servers_for` answers from the list in hand and never waits for a read. A list that is due triggers one refresh, coalesced by a flag so a burst collapses into one, which runs `scopes()` on `spawn_blocking` with 2 s to finish. A read that outlasts that leaves the list in hand in place. The `Sources` trait is unchanged, so the tests keep their fake. |
| 6. DNS failover stopped at SERVFAIL (P2) | `26c7b65e` | SERVFAIL, REFUSED and NOTIMP move to the next server, as a timeout or a transport error already did. NOERROR and NXDOMAIN are answers and are returned as they are. When every server refused, the guest gets the last upstream's own reply; the gateway's own SERVFAIL only when there was none. |
| 7. A resolver's custom port became 53 (P2) | `f3804e49` | The parser reads scutil's `port :` line and gives it to every nameserver of that resolver, so a local resolver on `127.0.0.1:5353` is asked there. A loopback nameserver here is a host-side socket this service opens itself, which the guest boundary does not judge; the code says so. |
| 8. Equal-domain resolvers lost priority and fallback (P2) | `cfe5f855` | `order :` is parsed into the scope. A name is answered by every resolver whose domain ties for the longest match, lowest order first and file order breaking ties, with their servers one after the other. The default resolvers are ordered the same way. |
| 10. DNS fallback took longer than the guest waits (P2) | `3ba4613f` | Hedged fallback: the first server is asked, the next after 700 ms while the first stays pending, first usable answer wins. Whole search 1800 ms, then the last real reply or SERVFAIL. A silent server is asked last for the next 30 s. The three numbers are constants that name the guest stub's 2 s they sit under. `UPSTREAM_TIMEOUT` (5 s per server) is gone: the budget bounds the search. |

`1d0cb090` is a test-only commit: the wait helper's deadline is an
assertion inside the loop, not a `panic!` after it, so the per-file
coverage gate sees every line run.

Red first, in each case: five concurrent queries read the host's
configuration five times and each waited for its own read; a first server
answering SERVFAIL ended the search; a resolver with `port : 5353` was asked
on 53; the second of two `corp.example` resolvers replaced the first; and a
silent first server made the healthy second answer at 5 s.

### New tests

- **Refresh**: five queries against a source that blocks produce one read and
  none of them waits for it; a read that outlasts its 2 s leaves the previous
  scopes in place; the refreshed list lands once the read returns.
- **Failover**: a first server saying SERVFAIL and a second saying REFUSED
  send the query on to a healthy third; NXDOMAIN from the first is the answer
  and the second is never asked; when every server refuses, the guest gets
  the last upstream's own reply rather than one the gateway made up.
- **Port and order**: a scutil resolver with `port : 5353` is asked there on
  every nameserver it names, and one whose port is not a number keeps 53; two
  `corp.example` resolvers with orders 200000 and 100000 answer 100000 first,
  then 200000.
- **Hedging**: a silent first server and a healthy second answer in under 1 s;
  every server silent is a SERVFAIL inside the budget; the silent server is
  not asked at all on the next query and is asked first again after its
  demotion runs out.

## Review at c46ea5a9, everything outside the DNS relay

The other five findings of the second review, one commit each.

| Finding | Commit | What it does now |
|---|---|---|
| 1. Network ownership ended before the VMM did (P2) | `c5080cd5` | `vm::boot_with_attachments` hands the netdev guard and the run's volume leases to the blocking closure that runs the VMM. They drop when `backend.run` returns, on every path, so an aborted boot task no longer cancels the stack or releases a volume under a live guest. The orchestrator hands over both; the provisioner hands over its netdev guard. No cross-guest descriptor reuse is claimed: the VM end is Vz's, held by an NSFileHandle with `closeOnDealloc: false`, and the guard is the only thing that closes the host end. |
| 9. Empty UDP replies disappeared (P2) | `d02a06a7` | The gateway builds its own guest-bound UDP packets with etherparse and sends them on the frames path it answers ARP, DHCP and DNS on. An upstream's zero-length datagram now reaches the guest, and the reply's size is ours: one larger than a frame of this link, or one that is not IPv4, is dropped and counted rather than written oversized. `DNS_UDP_CEILING` became `UDP_PAYLOAD_CEILING`, because the same number now bounds every reply. |
| 4. The smoke witness passed when its listener never started (P2) | `4301981a` | The script runs a positive control from the host (`curl http://127.0.0.1:$WITNESS_PORT/` must return `REACHED!`) and prints its result, clears the witness log before the guest probes, and fails at the end if the witness subshell is gone. `shellcheck` is clean. |
| 2. The subnet and rollback docs set the variables in the wrong process (P2) | `ad86ba17` | `docs/service.md` now shows `lns service stop` followed by `lns service start` with the variable in its environment, plus the foreground alternative, and says that a service the login agent starts at the next login does not carry it. `lns run` prints one warning line naming whichever of `LNS_NETDEV` and `LNS_GUEST_SUBNET` it finds in its own environment. |
| 5. The reachability table overstated ICMP (P3) | `ba4dc1db` | The table holds a row for it: ICMP to anything but the gateway is dropped, only TCP and UDP are carried, and `ping 1.1.1.1` from the guest gets nothing where the bridge forwarded it. The same commit brings the DNS section up to what the DNS round shipped. External ICMP stays out of scope. |

Red first: aborting the boot task dropped the attachments while the fake
backend still ran; a host that answered with zero bytes left the guest
waiting until the test's deadline.

New tests: the attachments outlive an aborted boot task and drop when the
VMM returns; an upstream datagram with no payload reaches the guest through
the production engine; a reply with no payload is written as 20 bytes of
IPv4 and an 8-byte UDP header with nothing after it; a reply no frame can
carry and one that is not IPv4 are counted, and the next reply still goes;
and the three cases of the `lns run` warning.

## Acceptance on a real Mac, first pass

The stack passed; the smoke test read a correct build as a failure. Fixed
here, one commit each:

| Item | Commit | What changed |
|---|---|---|
| The refusal probes judged by curl's exit status | `b5c99ac9` | A refused destination answers in the guest as `HTTP/1.1 502 Bad Gateway` with an empty body, from the in-guest proxy whose upstream connect the netstack reset. curl exits 0 on a 502, so `curl … && echo GATEWAY-ANSWERED` fired on a correct build. Each probe now reads the body. |
| The guard test read the task count too early | `3350867a`, `c46ea5a9` | `JoinSet::abort_all` only asks; a worker thread drops the aborted future after it. The test read `Arc::strong_count` at once, so it failed under load in the env-parity pass. It now waits, bounded by `PATIENCE`, for the count to fall to one. |

How a probe reads now. The two witness targets are answered only when the
body carries the witness listener's `REACHED!`:

```sh
probe HOST-NAT http://192.168.127.254:47123/
probe LOOPBACK http://127.0.0.1:47123/
```

The gateway must serve nothing at all on TCP, so its probe fails on a
non-empty body. A passing run prints:

```
[pass]  http://192.168.127.1/ served nothing on TCP; the guest proxy answered 502
[pass]  http://192.168.127.254:47123/ is refused; the guest proxy answered 502
[pass]  http://127.0.0.1:47123/ is refused; the guest proxy answered 502
[pass]  the host listener saw no accept at all
```

The host listener that logs every accept stays the authoritative oracle. The
header comment of the script now states that a 502 from the guest proxy is
the expected shape of a refusal. `shellcheck` is clean on the script.

## Review at c46ea5a9

Three integration defects against the locked netstack-smoltcp 0.2.4, one
commit each:

| Finding | Commit | What changed |
|---|---|---|
| 1. TCP admission ran after the library had already allocated, and a retransmitted SYN opened a second connection (P1) | `3ab2437d` | The engine admits a flow at ingress, before the stack allocates: a repeated SYN is dropped and counted, a SYN over the 1024-flow bound is reset and counted, the permit is taken at admission, and a flow leaves the table when its relay ends or 30 s after a handshake the guest never finished. Watched red first: the same SYN sent three times opened three host connections. |
| 2. One malformed UDP datagram stopped UDP and DNS for the rest of the run (P2) | `22b89c67` | Ingress reads with the library's own parsers and trims the packet to its own length, so nothing the library rejects reaches it; and `accept_udp` / `accept_tcp` read on after an item the library will not hand over. Watched red first: a datagram that declared 100 bytes more than it carried ended DNS for the run. |
| 3. A DNS answer larger than the guest can receive over UDP had no delivery path (P2) | `e3103463` | DNS over TCP on the gateway, and the UDP answer fitted to the guest with RFC 2181 §9 truncation. |
| Coverage of the over-limit path | `a518f427` | The reset for a SYN over the bound is pinned on the path the run takes to the link, not only at classification. |

**Deviation.** The brief asked for smoltcp's `TcpPacket::new_checked` at
ingress as well. etherparse's `TcpHeader::from_slice` already accepts exactly
the same segments — both require 20 bytes, a data offset of at least 5 words
and a slice that holds the whole header — so a second check would be a line
no test can reach, which the per-file coverage gate rejects. The engine keeps
one parser and a test pins that the two agree over a sweep of mutated
segments, so a future bump that makes them disagree fails.

**No DNS-over-TCP probe in the smoke test.** The sandbox's own nftables rules
in the guest (`lens-sandbox-core`, `crates/lens-sandbox-core/src/network.rs`,
`meta mark != <mark> meta l4proto tcp redirect to :<transparent>`) redirect
every unmarked TCP connection to the in-guest transparent proxy, so no tool
in `curlimages/curl` can reach `192.168.127.1:53` over TCP. The script says so
in its header; the Rust tests drive the path through the smoltcp fake guest
instead. **Sibling finding:** the in-guest DNS stub
(`/ref/lens-sandbox-core/crates/lens-sandbox-core/src/dns.rs`,
`forward_upstream`) relays the gateway's answer verbatim and never retries a
TC=1 answer over TCP itself. The workload's own resolver then retries over
TCP and is taken by the transparent proxy. The marked resolver
(`crates/lens-sandbox-core/src/resolver.rs`) does reach the gateway on TCP, so
this half of the fix is used; the stub's retry is a change for that
repository, not this one.

### New tests

- **Admission**: a repeated SYN is dropped and one flow is held; a SYN over
  the bound is answered with a reset and nothing is allocated; a flow whose
  relay ended gives its place back; one admitted flow is one stream; a flow
  the guest never finished opening is expired and its relay cancelled; a
  stream whose flow has already ended is dropped and counted; a relay that
  ends releases its flow. End to end through the real stack: the same SYN
  sent three times opens **one** connection on the host.
- **Malformed input**: a datagram that declares more bytes than it carries,
  a packet that declares more bytes than the frame carries, and a segment
  whose data offset runs past its end are each dropped; a padded frame is
  carried at its own length; through the fake guest, all three are sent and
  the next DNS question is still answered; the reader treats an item the
  stack will not hand over as unusable, counts it, waits, and ends with the
  run.
- **DNS**: an answer that fits comes back unchanged; a 1800-byte answer comes
  back to a guest without EDNS as at most 512 bytes with TC set, an empty
  answer section and the question intact; to a guest advertising 4096 as at
  most 1472 bytes with TC set and its OPT record kept; the same question over
  TCP to `192.168.127.1:53` through the fake guest returns all 120 records
  with the length prefix and no TC; a connection serves more than one query
  and closes with the guest, with bytes that are not a question, over the
  query limit, and when left idle; a gateway SYN to any other port is still
  refused with a reset.

## Third review at ba4dc1db

Five findings about how the earlier fixes interact, one commit each:

| Finding | Commit | What changed |
|---|---|---|
| 1. DNS-over-TCP connections took TCP admissions without a bound of their own (P2) | `c9e6c1e9` | The resolver's connections have an allowance of their own: `MAX_DNS_TCP_CONNECTIONS = 64`, taken in `tcp_ingress` before the stack allocates, so the 65th gets a reset the way the flow bound gives one, and given back when the connection ends. One connection carries 64 queries, then the gateway closes it. An answer that does not reach the guest in 5 s closes it too. Each of the three is counted under its own name. |
| 2. The first resolver read ran synchronously on the runtime with no timeout (P2) | `f403830f` | `Resolvers::new` starts with an empty list and makes the first read the same bounded refresh every later one is. Nothing in a run's start path waits on `scutil`. A query put before the first read lands gets SERVFAIL inside the 1800 ms budget, and that failure makes the next query read again. |
| 3. A refresh timeout left the blocking read and its `scutil` child running (P2) | `09ca94dd` | `Sources::scopes` is async. `/etc/resolv.conf` is read with `tokio::fs` and `scutil --dns` runs through `tokio::process` with `kill_on_drop`, so the 2-second deadline drops the read and takes the child with it. The in-flight flag clears only when the read has ended, so a second refresh cannot start before then. Folded in: a read that names no server at all keeps the list in hand and says so once. |
| 4. Refusal and table-full resets ignored data on the SYN (P2) | `6b0e1355` | A segment that carries no acknowledgement is reset from sequence + 1 + payload length, per RFC 793, so a SYN carrying data gets a reset the guest accepts instead of waiting out its connect timeout. |
| 5. Dropping `Running` cancelled relays but never observed that they ended (P3) | `db8bebc6` | The drop spawns one detached wait on the tracker. Relays all gone: nothing is said. Relays still running after 5 s: the count is named at warn level in the service log. |

### New tests

- **DNS over TCP**: 64 connections to `192.168.127.1:53` are carried and the
  65th is answered with a reset and allocates nothing, while a SYN to a real
  destination is still admitted; a connection that ends gives its allowance
  back; a connection the flow table has no room for holds no DNS allowance;
  64 queries on one connection are answered and the 65th closes it, counted;
  an answer a guest never reads is given up on at the write timeout, counted.
- **The resolver list**: the first read holds up nothing the run starts, and
  a query put before it lands is answered from an empty list rather than
  waiting; a read that never ends is dropped at its deadline, the in-flight
  flag is true while it runs and false after, no second read starts before
  then, and the next query reads again; a read that names no server at all
  keeps the list in hand.
- **Resets**: a SYN carrying five bytes to a refused destination is
  acknowledged at sequence + 6.
- **Shutdown**: after a normal stop the tracker holds zero relays; a relay
  that outlives its run by the grace period is named in the service log; a
  stop that leaves nothing behind says nothing; a run dropped where there is
  no runtime waits for nothing.

A sixth commit, `bbd4fb72`, answers the coverage gate: a slow read that ends
inside its deadline had no test, and the multi-line warning for relays that
outlive their run is written on one line so its values are not an unreachable
argument line.

Docs: `docs/service.md` carries the new bounds next to the others, and says
what a read that is too slow or names nothing does.

**Not in this round.** The guest's own DNS stub cannot retry a truncated
answer over TCP, because the guest's nftables redirect TCP/53 into the
transparent listener, which drops it. That is a `lens-sandbox-core`
limitation that predates this PR — vmnet had the same — and is filed there.
The per-flow memory sizing went to the maintainers. They decided it on
2026-09-12: the buffers go to 64 KiB per direction and the 1024-flow bound
stays. `916b7e8a` carries that decision; round six below says what it costs.

## Review round six: buffer sizing

The maintainers decided the open question on 2026-09-12. One item, one commit:

| Item | Commit | What changed |
|---|---|---|
| Per-flow TCP buffers go from 256 KiB to 64 KiB per direction; the 1024-flow bound stays | `916b7e8a` | `engine::TCP_WINDOW_BYTES` is `64 * 1024`. It is still the one constant both `netstack-smoltcp` buffer options read, so the socket and its control block follow it together. `Limits`, the flow bound, the DNS-over-TCP allowance and `CLOSE_DELAY` are untouched. `docs/service.md` carries the new figure and what it costs per guest. |

**Why 64 KiB.** The smoltcp buffer covers the guest-to-host hop only. That
round trip is well under a millisecond, because the stack runs in lns-service
on the same machine as the guest. The wide-area latency is absorbed by the
host kernel socket on the other side of the relay, which autotunes its own
buffers. The measured ceiling — 15.7 MB/s, equal to gvproxy — is the poll
loop, not the window.

**What it saves.** smoltcp keeps a closed socket in TIME-WAIT for
`CLOSE_DELAY` (10 s) with four buffers: two in netstack-smoltcp's socket and
two in its control block. At 256 KiB that was about 1.3 MB a flow and about
1.3 GB at the 1024-flow bound. At 64 KiB it is about 256 KiB of buffers plus
overhead a flow, so about 256 to 330 MB per guest at the same bound, plus at
most ten seconds of closed flows.

**The test.** `the_window_the_stack_offers_the_guest_is_the_one_this_link_is_sized_for`
puts a real smoltcp guest on the link, opens one flow to a host listener, and
reads what the stack put on the wire: the window scale the SYN-ACK asks for,
applied to the window advertised once the connection is up. It read 262144
before the change and reads 65536 after. It was watched fail first.

**Not re-measured here.** This sandbox has no VM, so throughput and the
memory a flood of flows holds are for acceptance on a real Mac, with
`crates/lns-cli/tests/smoke/netstack-netdev.sh` and a 100 MB fetch.

## Review round seven: DHCP wire types

The maintainers decided this on 2026-09-14: one protocol parser fewer to own.
The lease policy is unchanged. One item, one commit:

| Item | Commit | What changed |
|---|---|---|
| The hand-written DHCPv4 parser and builder give way to smoltcp's own wire types | `5850d13e` | `parse` is `DhcpPacket::new_checked` plus `DhcpRepr::parse`. `build` fills a `DhcpRepr` and emits it into a buffer `DhcpRepr::buffer_len` sizes. The private `options`, `address` and `push_option` helpers and the byte offsets and option codes go with them. `answer`, `wants_another_address`, `reply_to`, `Lease`, `Reply` and `ReplyTo` are untouched. No new dependency: the crate already compiles smoltcp with `proto-dhcpv4`. |

**What the parser now refuses.** Two inputs the old parser accepted are
rejected:

- A message whose type option does not match the packet opcode. A packet with
  `op = 1` (BOOTREQUEST) carrying message type 2 (OFFER) is a client posing as
  a server. smoltcp checks the pair; the old parser read the type and ignored
  the opcode.
- A message whose hardware type is not ethernet, or whose hardware length is
  not 6. The old parser checked these too, so this is the same answer through
  a different check.

Neither input was ever answered: `answer` replies only to DISCOVER and
REQUEST. The change is that `parse` stops earlier.

**What the reply now looks like.** The same options with the same values. The
order is the library's — 53, 54, 3, 1, 51, 6 — where it was 53, 54, 1, 3, 6,
51. A DHCP client reads options by code, and the guest's `udhcpc` does. `secs`
is 0 and `giaddr` is unspecified, as before. A NAK still carries only 53 and
54, with `yiaddr` and `siaddr` unspecified.

**Tests.** Every existing test in `dhcp.rs` and `engine.rs` asserts what it
asserted before. `Request.flags: u16` became `Request.broadcast: bool` and
`Request.kind: u8` became `DhcpMessageType`, so three assertions name the
enum instead of a number.

Two tests are new:

- `a_type_that_does_not_match_the_packet_opcode_is_not_read_as_a_request`
  builds a BOOTREQUEST carrying message type OFFER and asserts `parse`
  returns `None`. It was watched fail on the old parser, which returned
  `Some(Request { kind: 2, .. })`.
- `an_offer_reads_back_through_the_library_that_wrote_it` runs the emitted
  OFFER back through `DhcpRepr::parse` and pins every field: message type,
  transaction id, secs, client hardware address, `ciaddr`, `yiaddr`,
  `siaddr`, `giaddr`, the broadcast flag, server identifier, router, subnet
  mask, DNS servers and lease duration.

**Not verified here.** This sandbox cannot build `lns-service` — the Linux
tray pulls GTK, and `pango` is not installed and not reachable. The DHCP
module and its tests were compiled and run in isolation against the same
smoltcp 0.12.0, red first and then green, with `clippy -D warnings` and
`fmt --check` clean. CI is the oracle for the workspace gate and the coverage
floor. The master re-runs
`crates/lns-cli/tests/smoke/netstack-netdev.sh` on a real Mac to confirm
`udhcpc` still takes the lease.

## What I verified

Everything below ran in this sandbox (Linux) or in CI. All netdev modules
build and test on Linux and macOS targets; only the wiring is
`cfg(target_os = "macos")`. 163 netdev tests, all green, and the coverage gate
passes with no IGNORES entry for any of the new code. `smoltcp` is now a
direct dependency at `0.12.0`, the version `netstack-smoltcp` already
resolves to, so `Cargo.lock` is unchanged and `audit` has nothing new to
judge.

- **Unit tests**: ethernet, ARP and the drop reasons; the DHCP exchange
  including a bare renew and a NAK; the DNS relay — a query forwarded
  unchanged, the TCP retry on a truncated answer, each server tried in turn,
  SERVFAIL when none answers, `resolv.conf` and `scutil --dns` parsing,
  longest-suffix selection, the 30 s refresh and the refresh-on-failure; the
  default boundary per range; packet classification, the reset, the ICMP
  port-unreachable and the ICMP echo reply; the per-guest limits at their
  real defaults, including that the 1025th TCP flow, the 513th UDP flow and
  the 257th query are refused; the subnet override and the host-overlap
  warning.
- **Malformed input sweeps**: every truncation of an IPv4/UDP packet shorter
  than its own headers, every truncation of a TCP packet, and every prefix of
  an arbitrary ethernet frame up to 64 bytes. Nothing panics and nothing
  half-read is forwarded.
- **An integration test with smoltcp as the fake guest.** A real smoltcp
  `medium-ethernet` device over the other end of the frame channels: the
  guest takes a lease with smoltcp's DHCPv4 client socket and gets
  `192.168.127.2/24`, router and DNS `192.168.127.1`; resolves a name through
  the gateway against a stand-in nameserver; gets SERVFAIL for a name that
  nameserver does not know; has an MX question relayed unchanged; gets the
  whole of an answer the nameserver truncates over UDP; exchanges two UDP
  datagrams with a host echo socket over one host socket; opens TCP to a
  listener the test starts, sends `ping!`, reads `pong`, sees its side close
  when the host closes, and closes its own. With the default boundary it is
  refused — socket `Closed`, not hung — for both `127.0.0.1:<port>` and
  `192.168.127.1:80`. With `tcp_flows: 1` and `udp_flows: 1` a host listener
  and a host echo socket each see exactly one connection.
- **A socket-pair test** that runs the whole thing over a real
  `socketpair()`: an IPv6 frame and an IPv4 fragment are dropped, an ARP
  request for the gateway is answered on the wire, and a ping at the gateway
  gets an echo reply.
- **A guard test**: a guest opens a TCP flow to a test listener and sends on
  it; the run's guard is then dropped. The host's side of the relayed flow
  closes, the relay task ends, every task adopted by the transport front end
  is aborted, and nothing reads the link any more.
- **A refusal-log test**: a guest that refuses 1040 distinct destinations
  leaves 1024 in the set the run logs from, says once that it counts the
  rest, and has every one of the 1041 refusals counted.
- `cargo clippy --all-targets -- -D warnings` and
  `cargo clippy -- -D clippy::cognitive_complexity` clean; `cargo fmt`
  applied. `audit` is green in CI, and none of the RustSec advisories for
  `hickory-proto` apply to the 0.26.1 already in `Cargo.lock`
  (`RUSTSEC-2025-0006` patched ≥0.25.0-alpha.5, `RUSTSEC-2026-0118`
  unaffected ≥0.26.0-beta.1, `RUSTSEC-2026-0119` patched ≥0.26.1). There are
  no advisories for `netstack-smoltcp`, `smoltcp` or `etherparse`.

## What I could not verify

- **No local workspace gate.** This sandbox has no GTK development
  packages, so `lns-service` does not link here out of the box (`atk-sys`
  fails in `pkg-config`). For the third and the sixth review rounds I stubbed
  the GTK libraries and ran the crate itself: `cargo test -p lns-service --lib`,
  2672 tests green, and `cargo clippy -p lns-service --all-targets -- -D
  warnings` clean. For the earlier rounds I ran the netdev modules —
  `netdev.rs`, `engine.rs`, `dhcp.rs`, `dns.rs`, `link.rs`, `policy.rs`,
  `resolver.rs` — in an isolated crate that compiles the same sources: 103
  tests green,
  `clippy --all-targets -- -D warnings` clean,
  `clippy -- -D clippy::cognitive_complexity` clean, `cargo fmt` applied.
  The two new tests were watched fail first against the old code. CI is the
  oracle for the workspace gate and the coverage floor.
- **No macOS host and no VM.** I could not boot a guest, so the Vz
  attachment, the real `udhcpc` exchange in `lns-session-broker`, the
  `scutil --dns` output of a real VPN, and the smoke test are unrun. I read
  the guest side (`crates/lns-session-broker/src/network.rs`): it runs
  `ip link set eth0 up` then `udhcpc -i eth0 -q -n -t 2 -T 3 -s <hook>` and
  consumes `ip`, `mask`, `router` and `dns` — all four are in every OFFER and
  ACK this server sends. No change to that crate was needed.
- **Throughput.** No VM, so no 100 MB CDN measurement. CI's macOS runners
  cannot boot a guest either. This is left to acceptance on a real macOS
  host: `crates/lns-cli/tests/smoke/netstack-netdev.sh` checks the lease, the
  resolver, the address, egress through the guest proxy, a dual-stack fetch,
  the audit line, that `curl http://192.168.127.1/`,
  `http://192.168.127.254:<port>/` and `http://127.0.0.1:<port>/` all fail
  from a guest whose egress is `match: "*"`, and that a host listener logging
  every accept sees none.

## Known limitation

`netstack-smoltcp` completes the TCP handshake before the host connect is
attempted, so a destination that is **allowed but refuses the connection**
leaves the guest's socket open until its own timeout instead of returning a
reset. Destinations the boundary refuses are not affected — those are reset
before the stack sees the SYN. A follow-up can synthesise the reset from the
connection's four-tuple.

## Follow-up

Linux keeps cloud-hypervisor and its own netdev. The same stack can serve it
through a vhost-user-net front end that feeds `engine::Frames`; nothing in
the stack is macOS-specific.

Refs #380


chenhunghan:
## Design amendments from review (apply before asking for acceptance)

These change the brief you are implementing; treat them as review threads to resolve in this PR.

1. **DNS is a plain relay, one path.** Drop the getaddrinfo synthesis for A/AAAA. Forward every query unchanged over UDP to the host's resolvers; on a TC-truncated answer retry over TCP to the same server. Resolvers: the nameservers of `/etc/resolv.conf` plus per-domain scoped resolvers parsed from `scutil --dns` on macOS (split-DNS VPNs are absent from resolv.conf); longest matching domain suffix wins; refresh the list every 30 s and on failure. 5 s per upstream, then the next, then SERVFAIL. Negative answers and TTLs pass through; no cache in lns-service. At most 256 in-flight queries per guest.
2. **Bounds per guest.** 1024 concurrent TCP flows, 512 UDP flows; TCP connect timeout 10 s; UDP idle 60 s; 256 KiB buffers per direction per TCP flow; a bounded channel between the socket reader and the stack, dropping frames when full and counting the drops at debug level. All relay tasks in one `JoinSet` per run, aborted when the guard drops. Add a unit test that the limits refuse the 1025th flow and that dropping the guard aborts every task.
3. **Subnet override.** `LNS_GUEST_SUBNET=<a /24>` replaces 192.168.127.0/24 for the service; gateway `.1`, guest `.2`, policy ranges derived from it. At netdev start, if a host interface address falls inside the guest subnet, log a warning naming the override. Document in docs/service.md.
4. **Packet boundary, explicit.** Ethernet II demux: ARP and IPv4 handled; IPv6, VLAN and anything else dropped and counted. MTU 1500 both ways, no IP reassembly. ICMP echo to the gateway answered. Refused TCP gets a reset; refused UDP an ICMP port unreachable if the stack can emit one, else a silent drop. State IPv4-only in the docs. Unit tests feed truncated and malformed Ethernet, ARP, IPv4 and DHCP inputs and assert nothing panics and nothing is forwarded.
5. **Smoke test additions.** A refused destination is verified by a listener on the host that logs accepts and sees none, not by the failed connect alone. Add a dual-stack host fetch (`https://example.com`) to show the A fallback works.
6. **PR body.** Add a "Bounds and limits" table and a "DNS" paragraph describing the above.

chenhunghan:
## Review at 6e29f21

Read every netdev module and the smoke test against the brief, the six amendments and my own earlier list. The amendments landed as the PR body's table says: DNS is a relay with scoped resolvers and TC→TCP, the per-guest limits exist and are tested, fragments are dropped, the reader drops and counts, the smoke test has a witness listener, `LNS_GUEST_SUBNET` works. Good.

Three things remain before acceptance:

1. **Relay tasks still escape the guard** (`engine.rs` `accept_tcp`, `open_flow`, `answer_dns`): `carry_tcp`, `carry_udp` and `relay_query` are started with `tokio::spawn`, so `Running::drop` aborts the listeners but not the relays. Their lifetime is now bounded by the 10 s connect, 60 s UDP idle and 5 s DNS timeouts, so it is no longer a leak, but the PR body's "dropping the guard aborts every task" is not true for them, and a run stopped mid-transfer keeps copying for as long as both sockets stay open. Give the relays the same owner: a `tokio_util::task::TaskTracker` plus `CancellationToken` held by `Running` (or an `Arc<Mutex<JoinSet>>` handle passed to the acceptors), cancel and close in `drop`, and make the existing "dropping the guard aborts every task" test open a relay first.

2. **The refusal log set is unbounded** (`note_refusal`, `Refused = HashSet<SocketAddr>`): every distinct refused destination is remembered for the life of the run; a guest scanning ports across loopback and the guest subnet grows it without limit. Cap it (1024 entries, then one "further refusals are counted, not logged" line) and add the test.

3. **Two `MAX_FRAME` constants**: `netdev.rs` 1518 and `engine.rs` 1514. One constant, the larger, in one place.

Everything else I looked for is in place. After these, I run the acceptance on a real Mac: lease, resolver, TLS egress, split-DNS VPN, refused destinations with the witness listener, concurrent boots, create/destroy cycles, throughput, interrupted download, and lns-service fd and RSS before and after.

chenhunghan:
Acceptance on a real Mac, first pass at 6e29f21: the stack works and the boundary holds, but the smoke test reports a false failure.

- Guest leased 192.168.127.2/24, resolver 192.168.127.1, `https://example.com` 200, `https://api.anthropic.com/v1/models` 401 (Anthropic's own answer). Boot to session about 5 s.
- From a `match: "*"` guest, `curl -i http://192.168.127.1/` and `curl -i http://192.168.127.254:47123/` both get `HTTP/1.1 502 Bad Gateway`, `Content-Length: 0`, from the in-guest transparent proxy whose upstream connect the netstack reset. A host listener on 127.0.0.1:47123 that logs every accept saw none.

The smoke test's refusal check is `curl -sS --max-time 5 http://192.168.127.1/ && echo GATEWAY-ANSWERED`. curl exits 0 on a 502, so the proxy's own error page counts as "the gateway answered" and the test fails on a correct build. Fix for this round: use `curl -sSf` (fail on HTTP ≥ 400) or, better, assert on the body, since the witness prints `REACHED!` and the gateway would print something non-empty; keep the witness log check as the real oracle.

chenhunghan:
## Acceptance on a real Mac at e0a2aba (the three fixes in)

Second lns service on its own home, signed release build, `curlimages/curl:8.11.1` guests, a host witness listener on 127.0.0.1:47123 logging every accept.

| Check | Result |
|---|---|
| Lease, resolver | 192.168.127.2/24, `nameserver 192.168.127.1`, boot to session about 5 s |
| 4 concurrent cold boots | all four leased in 2.0 s each |
| TLS egress | `https://example.com` 200; `https://api.anthropic.com/v1/models` 401 (Anthropic's own answer) |
| Gateway :80, `.254:47123`, `127.0.0.1:47123` from a `match: "*"` guest | guest proxy returns `502 Bad Gateway`, empty body; witness log: **0 accepts** across every probe of the session |
| DNS through the gateway | TXT for example.com answered; NXDOMAIN reported; direct UDP to 1.1.1.1:53 answered; SRV lookup returned nothing (see note) |
| IPv6 on the guest link | none, as documented |
| Throughput, 66 MB from dl.google.com, two runs each | netstack 15.7 and 11.1 MB/s; gvproxy build 15.6 and 10.9 MB/s, same minute. lns-service CPU peaked at 46 % for netstack; gvproxy's own process peaked at 62 % |
| Interrupted download (curl killed at 2 s) | 0 host connections to the server 6 s later |
| Run stopped mid-transfer (`lns stop`) | the host connection to the server was gone 2 s after stop, so the relay guard works |
| 30 create/destroy cycles | every run 200; lns-service RSS 82 → 285 → 293 → 301 MB, fds 47 → 86 → 116 → 146 |
| `LNS_NETDEV=vmnet` | boots, no lease on this host (that is #380, the control) |

**On the growth row:** the gvproxy build shows the same curve on the same cycles (82 → 259 → 280 MB, fds 47 → 83 → 113), and the extra fds are one `runs/<id>/console.log` and two unix sockets per finished run. It is pre-existing lns-service behaviour, not the netstack; I will open a separate issue for it.

**SRV note:** `nslookup -type=srv` from busybox printed no records for `_xmpp-server._tcp.google.com`; busybox's nslookup does not print SRV, so this is not evidence either way. A `dig`-capable image or the split-DNS check on a VPN host is still owed.

Still open before merge: the smoke test's refusal check counts the proxy's 502 as an answer (fix round dispatched). Otherwise this build is what I would ship as opt-in in release N per NEW_NETWORK.md §8.

chenhunghan:
Smoke test at the PR head, run on the real Mac against the second service: every check passes, including the three refusal probes (the guest proxy answered 502 / 000, the witness listener saw no accept) and the audit line. Nothing left open from my side; SRV and split-DNS on a VPN host are the remaining owed checks before the default flips, per NEW_NETWORK.md §8.

chenhunghan:
Second review, part two: five more findings at c46ea5a, each confirmed by reading the code. A fix round is queued for them (it follows the round now running for the SYN admission, malformed-UDP and DNS-over-TCP findings). Recorded here so the author sees them in one place.

1. **Network ownership ends before the VMM does (P2).** `orchestrator.rs:631` and `provisioner/real.rs:222` hold the netdev guard inside the `tokio::spawn` that awaits `vm::boot`, which runs the VMM on `spawn_blocking`. Aborting that task (orchestrator after a connector that never arrived, provisioner after a failed stop) drops the guard and cancels the stack while the VM keeps running until the VMM exits by itself. Fix: the guard's owner is the blocking closure, so it drops when `backend.run` returns, on every path. No cross-guest escape is claimed: the VM end is handed to Vz and not closed by the guard.
2. **The subnet and rollback docs set the variables in the wrong process (P2).** `docs/service.md` shows `LNS_NETDEV=vmnet lns run` and `LNS_GUEST_SUBNET=… lns run`, but lns-service reads both from its own environment and neither travels over IPC. Fix: document setting them on the service start, and have `lns run` warn when either is set in the CLI's environment.
3. **Resolver refresh blocks runtime workers under a mutex (P2).** `dns.rs:69` holds `std::sync::Mutex` while `resolver.rs` reads `/etc/resolv.conf` and runs `scutil --dns`; `stale()` after SERVFAIL makes a burst of failing queries run scutil once each, serialised on the lock. Fix: refresh on `spawn_blocking`, coalesced to one in flight, subprocess bounded to 2 s, cache swapped when done.
4. **The smoke witness passes when its listener never started (P2).** `netstack-netdev.sh:112`: `nc -l … || break` turns a bind failure into an empty log and a passing "zero accepts". Fix: positive control from the host before the guest probes, truncate the log, fail if the witness died.
5. **Reachability table overstates ICMP (P3).** Only echo to the gateway is answered; every other ICMP is dropped, while the last table row promises vmnet-equivalent forwarding. Fix the table; external ICMP is a separate scope decision.

chenhunghan:
Second review, part three: four more at c46ea5a, each confirmed by reading, queued for the same fix round as part two.

6. **DNS failover stops at SERVFAIL (P2).** `dns.rs:131` returns any non-truncated reply, so a first upstream answering SERVFAIL or REFUSED ends the search and a healthy second is never asked. Fix: SERVFAIL, REFUSED, NOTIMP and transport failures move to the next server; NOERROR and NXDOMAIN are final.
7. **A resolver's custom port becomes 53 (P2).** `dns.rs:204` parses the `nameserver[n]` address and pairs it with 53; scutil's separate `port :` line is never read. Fix: parse it and apply it to that resolver's servers.
8. **Equal-domain resolvers lose priority and fallback (P2).** `dns.rs:86` keeps one longest-suffix scope with `max_by_key` and ignores `order :`. Fix: keep every scope that ties for longest suffix, sorted by order ascending, servers concatenated.
9. **Empty UDP replies disappear (P2).** The library's `WriteHalf::start_send` returns Ok and emits nothing for a zero-length payload, so an upstream's empty datagram never reaches the guest. Fix: build guest-bound UDP replies ourselves and send them on the engine's own frames path, which also puts reply size under our control for the DNS truncation work.

chenhunghan:
Second review, part four, one more at c46ea5a, confirmed against both repositories and queued for the same round.

10. **DNS fallback takes longer than the guest waits (P2).** The guest stub in lens-sandbox-core (`crates/lens-sandbox-core/src/dns.rs`, `UPSTREAM_TIMEOUT` 2 s) abandons a query after 2 s; the gateway's `dns.rs:11` gives each upstream 5 s before asking the next. With a silent first nameserver the healthy second answers at 5 s to nobody, on every retry. Fix: hedged fallback in the gateway (next server asked after 700 ms while the first stays pending, first usable answer wins, 1800 ms total budget, silent servers demoted for 30 s), so a healthy fallback answers inside the guest's deadline.

Also noted from the same reviewer: the focused UDP stress checks passed (a second stack stayed responsive while the first was flooded with its output unread; the saturated stack recovered when drained; cancellation closed its frame output), and no additional policy bypass was found in the guest firewall and proxy path by reading. Live Vz, nftables and approval cards remain for the host acceptance.

chenhunghan:
Three integration defects against the locked `netstack-smoltcp` 0.2.4, one commit each. Every check on the branch is green at `70f7dad3`, coverage included, with no new IGNORES entry.

| Finding | Commit | What it does now |
|---|---|---|
| 1. TCP admission ran after the library had already allocated, and a retransmitted SYN opened a second connection (P1) | `3ab2437d` | The engine keeps the flows a guest holds and decides per SYN at ingress, before the stack allocates. A repeated SYN is dropped and counted; a SYN over the 1024-flow bound gets the reset a refusal gets, and is counted; the permit is taken at admission. A flow leaves the table when its relay ends, or 30 s after a handshake the guest never finished, which cancels its relay and drops its stream. |
| 2. One malformed UDP datagram stopped UDP and DNS for the rest of the run (P2) | `22b89c67` | Ingress now reads with the parsers the library uses (`Ipv4Packet::new_checked`, `UdpPacket::new_checked`) and trims the packet to the length its own header states, so nothing the library rejects reaches it. As a seatbelt, `accept_udp` and `accept_tcp` read an item the library will not hand over as one unusable item, not as the end of the stream. |
| 3. A DNS answer larger than the guest can receive over UDP had no delivery path (P2) | `e3103463` | The gateway serves DNS over TCP on `192.168.127.1:53` with the RFC 1035 §4.2.2 length prefix, closed after 10 s idle and counted under the DNS allowance. Over UDP the answer is fitted to what the guest advertised, never over 1472 bytes, and one that does not fit comes back as its own question with TC set per RFC 2181 §9. |

`a518f427` and `70f7dad3` are test commits that close the per-file coverage gate on the paths above.

Two things to know, both written up in the PR description:

- **The TCP check stays etherparse's.** `TcpHeader::from_slice` and smoltcp's `TcpPacket::new_checked` accept exactly the same segments, so a second check would be a line no test can reach. A test pins that the two agree over a sweep of mutated segments.
- **No DNS-over-TCP probe in the smoke test.** The sandbox's own nftables rules in the guest redirect every unmarked TCP connection to the in-guest transparent proxy, so no tool in `curlimages/curl` can reach the gateway on TCP. The script header says this; the Rust tests drive the path through the smoltcp fake guest instead.

Red first, in each case: the same SYN sent three times opened three host connections before `3ab2437d`; one datagram declaring 100 bytes more than it carried ended DNS for the run before `22b89c67`.

chenhunghan:
DNS round of the second review: items 3, 6, 7, 8 and 10, one commit each. Every check on the branch is green at `1d0cb090`, coverage included, with no new IGNORES entry. The PR description now describes the relay as it is.

| Item | Commit | What it does now |
|---|---|---|
| 3. Resolver refresh blocked runtime workers under a mutex | `4eb73523` | `servers_for` answers from the list in hand and waits for no read. A list that is due triggers one refresh, coalesced by a flag so a burst of failing queries collapses into one, which runs `scopes()` on `spawn_blocking` with 2 s to finish. A read that outlasts that leaves the list in hand in place. The `Sources` trait is unchanged, so the tests keep their fake. |
| 6. DNS failover stopped at SERVFAIL | `26c7b65e` | SERVFAIL, REFUSED and NOTIMP move to the next server, as a timeout or a transport error already did. NOERROR and NXDOMAIN are answers and come back as they are. When every server refused, the guest gets the last upstream's own reply; the gateway's own SERVFAIL only when there was none. |
| 7. A resolver's custom port became 53 | `f3804e49` | The parser reads scutil's `port :` line and gives it to every nameserver of that resolver, so a local resolver on `127.0.0.1:5353` is asked there. A resolver that names no port, or one that is not a number, keeps 53. A loopback nameserver here is a host-side socket the service opens itself, which the guest boundary does not judge; the code says so. |
| 8. Equal-domain resolvers lost priority and fallback | `cfe5f855` | `order :` is parsed into the scope. A name is answered by every resolver whose domain ties for the longest match, lowest order first and file order breaking ties, servers one after the other. The default resolvers are ordered the same way. |
| 10. DNS fallback took longer than the guest waits | `3ba4613f` | Hedged fallback: the first server is asked, the next after 700 ms while the first stays pending, and the first usable answer wins. The whole search has 1800 ms, then the last real reply or SERVFAIL. A silent server is asked last for the next 30 s. The three numbers are constants that name the guest stub's 2 s they sit under; `UPSTREAM_TIMEOUT` (5 s per server) is gone, because the budget bounds the search. |

`1d0cb090` is a test-only commit: the wait helper's deadline is an assertion inside the loop, not a `panic!` after it, so the coverage gate sees every line run.

Red first, in each case:

- five concurrent queries read the host's configuration five times, each waiting for its own read;
- a first server answering SERVFAIL ended the search and the healthy second was never asked;
- a resolver with `port : 5353` was asked on 53;
- the second of two `corp.example` resolvers replaced the first;
- a silent first server made the healthy second answer at 5 s, to a stub that stopped listening at 2 s.

What I could not verify: this sandbox has no GTK development packages, so `lns-service` does not link here. I ran the netdev modules in an isolated crate that compiles the same sources — 126 tests green, `clippy --all-targets -- -D warnings` and `clippy -- -D clippy::cognitive_complexity` clean, `cargo fmt` applied. CI is the oracle for the workspace gate and the coverage floor, and it is green. No macOS host, so the `scutil --dns` output of a real VPN is still unread; the port and order parsing are pinned by fixtures.

Items 1, 2, 4, 5 and 9 of the same review — the guard's lifetime, the documented variables, empty UDP replies, the smoke witness and the ICMP table row — are a separate round and are not in these commits.

chenhunghan:
Second review, everything outside the DNS relay: items 1, 2, 4, 5 and 9, one commit each. Every check on the branch is green at `ba4dc1db`, coverage included, with no new IGNORES entry. The PR description now describes the guard's lifetime, the resolver refresh and where the two variables are set.

| Item | Commit | What it does now |
|---|---|---|
| 1. Network ownership ended before the VMM did | `c5080cd5` | `vm::boot_with_attachments` gives what a run holds to the blocking closure that runs the VMM. The netdev guard and the run's volume leases drop when `backend.run` returns, on every path, so an aborted boot task no longer cancels the stack, closes the host end of the link, or releases a volume under a live guest. The orchestrator hands over both; the provisioner hands over its netdev guard. The volume leases had the same hazard and are fixed with it. No cross-guest descriptor reuse is claimed: Vz holds the VM end through an `NSFileHandle` with `closeOnDealloc: false`, so the guard is the only thing that closes the host end, and nothing closes the VM end but Vz. |
| 9. Empty UDP replies disappeared | `d02a06a7` | The gateway no longer uses the library's UDP writer. `write_udp` builds the IPv4 and UDP headers with etherparse and sends the packet on the frames path the engine answers ARP, DHCP and DNS on. An upstream's zero-length datagram reaches the guest, and reply size is now ours: a reply larger than one frame of this link, or one that is not IPv4, is dropped and counted rather than written oversized. `DNS_UDP_CEILING` became `UDP_PAYLOAD_CEILING`, because one number now bounds every reply. |
| 4. The smoke witness passed when its listener never started | `4301981a` | The script runs a positive control from the host before the guest probes — `curl -sS --max-time 3 http://127.0.0.1:$WITNESS_PORT/` must return `REACHED!` — and prints the result. The witness log is cleared after the control, so the guest's probes are judged on their own, and the run fails at the end if the witness subshell is gone. `shellcheck` is clean on the file. |
| 2. The subnet and rollback docs set the variables in the wrong process | `ad86ba17` | `docs/service.md` shows the route that works: `lns service stop`, then `lns service start` with the variable in the environment it passes to the service, plus the foreground alternative. It also says what does not carry the setting — a service the login agent starts at the next login, because the agent's plist holds no environment of yours. `lns run` prints one warning line when it finds `LNS_NETDEV` or `LNS_GUEST_SUBNET` in its own environment, naming the variables and the remedy. |
| 5. The reachability table overstated ICMP | `ba4dc1db` | The table has a row for it: ICMP to anything but the gateway is dropped, only TCP and UDP are carried, and `ping 1.1.1.1` from the guest gets nothing where the vmnet bridge forwarded it. The IPv4-only paragraph names the same difference. External ICMP stays out of scope for this round. The same commit brings the DNS section up to what the DNS round shipped — the tying resolvers in `order`, each on its own port, the hedged 700 ms / 1800 ms search, the demotion, and a refresh no query waits for. |

Red first, in each case:

- with a fake backend blocked on a channel, aborting the boot task dropped the attachments while the backend was still running;
- a host that answered a guest datagram with zero bytes left the guest waiting until the test's deadline.

New tests: the attachments outlive an aborted boot task and drop when the VMM returns; an upstream datagram with no payload reaches the guest through the production engine; a reply with no payload is written as 20 bytes of IPv4 and an 8-byte UDP header with nothing after it; a reply no frame can carry, and one that is not IPv4, are counted while the next reply still goes; and the three cases of the `lns run` warning.

What I could not verify here: this sandbox has no GTK development packages and no macOS host. I linked `lns-service` against stub libraries to run its tests — 2658 lib tests green, `clippy --all-targets -- -D warnings` and `cargo fmt` clean on both crates, `shellcheck` clean on the smoke script. CI is the oracle for the workspace gate and the coverage floor, and it is green. The smoke script's new control is BSD `nc` on macOS, so the host acceptance run is what proves it end to end.

chenhunghan:
Acceptance on this Mac at `ba4dc1db` (rounds three, 4a and 4b), second service on its own home, signed release build.

**Smoke test**: every check passes, including the new witness control (`REACHED!` from the host, log cleared, listener alive at the end) and the refusals judged by body.

**DNS through the gateway, with `dig` from `nicolaka/netshoot`**: A, SRV (`_xmpp-server._tcp.jabber.org`), TXT (`google.com`, 17 records, 1187 bytes over UDP with EDNS), DNS over TCP to `192.168.127.1:53` (with an `egress.tcp` rule for it), a hedged answer in 40 ms, NODATA for a nonexistent name exactly as the host resolver gives it. AAAA over UDP is empty because the guest's own stub answers AAAA with NODATA by design (lens-sandbox-core `dns.rs` header); over TCP to the gateway the AAAA records come back, which is worth knowing but harmless on an IPv4-only link. A storm of 600 concurrent queries finished and the next query answered.

**Bounds**: a guest opening idle TCP flows to the gateway's port 53 was admitted up to 512 sockets (the guest side stopped there), DNS over UDP kept answering during the flood, descriptors returned to baseline after it. Memory did not: lns-service went from 82 MB to 741 MB resident and stayed there. Three rounds of 600 flows in one guest plateau at 756 MB, so this is allocator retention, not a leak, but the peak is the finding: about 1.3 MB per admitted flow, which is the four 256 KiB buffers per connection (two in netstack-smoltcp's socket, two in its control block). At the 1024-flow bound one guest can pin about 1.3 GB in lns-service. Two remedies, a decision for the maintainers rather than a fix I would dispatch alone: smaller per-flow buffers (64 KiB would cap it near 330 MB; throughput at 256 KiB was 15.7 MB/s and should be re-measured at 64 KiB), or a lower flow bound (256) with the same buffers. Either way the number belongs in docs/service.md next to the bound.

**Not verifiable here**: a guest cannot reach any service on this host, on this backend or on vmnet, because this Mac's endpoint-security extensions black-hole TCP to the host's own addresses (handshake completes, listener never accepts; a plain shell shows it too). The parity harness (PR #428) therefore needs fixtures on a second machine before its transfer, policy and isolation cases can run here. VPN toggling is the user's, pending.

chenhunghan:
Third review, by a reader looking at how the earlier fixes interact, at `ba4dc1db`. Seven findings; my triage after each. A fix round for the first five is dispatched.

1. **DNS-over-TCP connections take TCP admissions without a bound of their own.** A guest opening idle connections to the resolver's port 53 fills its own flow table with them (I measured 512 admitted on a real guest) while taking no DNS permit. Real, though it harms only that guest: P2, fixed with a 64-connection allowance taken at admission, a per-connection query cap and a write timeout.
2. **The first resolver read runs synchronously on the runtime with no timeout.** `Resolvers::new` runs `scutil --dns` inline in the run's start path. Real: P2, the first read goes through the same bounded refresh.
3. **A refresh timeout leaves the blocking read and its scutil child running and lets the next refresh start another.** Real: P2, scutil moves to `tokio::process` with `kill_on_drop`, and the in-flight flag clears only when the read has ended.
4. **The guest stub cannot retry a truncated answer over TCP**, because the guest's nftables redirect TCP/53 into the transparent listener, which drops it. True, and not this PR's: the stub and its redirect are lens-sandbox-core's, vmnet had the same limitation, and this PR adds the server side of the path. Filed there.
5. **A refresh can replace a working cache with an empty list.** Real: P2, folded into item 3 (an empty result keeps the previous list).
6. **RSTs ignore data carried on a SYN.** Real: P2, ack = seq + 1 + payload.
7. **Dropping `Running` cancels relays without observing that they ended.** Cancelled relays release their sockets at their next poll; the gap is microseconds. P3: one detached wait with a warning if relays outlive the run by 5 s, so it is visible if ever wrong.

Confirmed by the same reader, by reading: repeated SYNs dropped, half-open entries bounded, claims released on relay end, the gateway admitted on port 53 alone, demotion and hedging preserve fallback order, NXDOMAIN never retried, UDP replies within `MAX_FRAME` with correct source and checksum.

chenhunghan:
Acceptance addendum at `db8bebc` (round five), same second service and signed build: the smoke test passes every check again.

The idle-flow flood against the resolver's port 53 now runs into the new bound: the service log at debug level counted 120 refusals of "a DNS connection over the guest's limit" while UDP DNS kept answering. Resident memory still peaked near 660 MB, and the reason is now clear rather than open: smoltcp holds a closed socket in TIME-WAIT for `CLOSE_DELAY` (10 s) with its four 256 KiB buffers, so memory follows the flows opened in the last ten seconds, not the flows admitted at once. The admission bound is honest about live flows; the buffer size decides the bill. That sharpens the sizing decision for the maintainers: 64 KiB buffers cut the per-flow cost to about 330 KB (256 MB at the bound, and churn adds at most ten seconds of it), where a lower flow bound alone would not touch the TIME-WAIT share. Throughput at 64 KiB should be re-measured against the 15.7 MB/s seen at 256 KiB before choosing.

From my side this PR is ready for the opt-in release once the maintainers pick the buffer size: every finding of the three reviews is closed or filed, the smoke test, 2679 unit tests, DNS through the gateway with `dig`, and the bounds probes all pass on a real Mac. What stays open is not this PR's: the guest stub's TCP retry (lens-sandbox-core #59), host-reachability on filtered Macs (design), the VPN toggle (a person), and the parity transfer cases (fixtures on a second machine).

chenhunghan:
VPN transition check on this Mac at `db8bebc`, netstack second service, one long-lived guest (`nicolaka/netshoot`, `dig` against `192.168.127.1`), the Cato client driven with `scutil --nc start/stop`.

| Moment | Host resolv.conf | Guest `dig example.com` | Guest `whoami.akamai.net` (which egress answered) |
|---|---|---|---|
| VPN on | 10.254.254.1 | answers, 12 ms | 74.125.74.147, the tunnel's egress |
| 8 s after disconnect | 1.1.1.1, 8.8.8.8 | empty (SERVFAIL) | empty |
| after the 30 s refresh | same | answers | 162.159.96.126, Cloudflare direct |
| right after reconnect | 10.254.254.1 | answers | 74.125.112.26, the tunnel again |
| after the refresh | same | answers | same |

So the gateway follows the host: with the VPN up the guest's queries leave through the tunnel's resolver, after a disconnect they move to the LAN resolvers, after a reconnect they move back, with no restart of the guest. The one gap is the window after a disconnect: the cached list still names the tunnel's resolver, the first queries fail with SERVFAIL, `stale()` triggers the re-read, and the list is right within the refresh interval (30 s at most, a few seconds in practice). A reconnect has no gap because the old resolvers stay reachable through the tunnel. No "host named no resolver" warning was logged across either transition.

Not checked: a private name only the tunnel's resolver can answer, because I do not have one to hand; the egress evidence above shows which upstream answered in each state, which is the same question from the other side.

chenhunghan:
Acceptance at `916b7e8a` (round six, 64 KiB buffers) on this Mac, second service on its own home, signed release build.

**Smoke test**: every check passes (lease, resolver, TLS egress, witness control, gateway and `.254` and loopback refused with zero accepts, audit chain names netstack).

**Throughput**, 71.6 MB (`go1.23.0.darwin-arm64.tar.gz` from dl.google.com) through a `curlimages/curl` guest, backends alternated within one minute:

| Run | netstack at 64 KiB | gvproxy build (PR #424) |
|---|---|---|
| 1 | 17.9 MB/s | 18.1 MB/s |
| 2 | 17.7 MB/s | 18.3 MB/s |
| 3 | 17.8 MB/s | |
| 4 | 17.3 MB/s | |

Within the 20 % gate of NEW_NETWORK §8; the 256 KiB build measured 15.7 MB/s against 15.6 the day before, so shrinking the window four times moved nothing, as expected for a window that only covers the guest-to-host hop.

**Memory, with a correction to my earlier figures.** The resident size of `lns-service` includes the guest's memory: it goes from 82 MB to about 527 MB the moment a 512 MiB `netshoot` guest boots, before any flow exists (checked with a guest that only sleeps and then writes 243 MB to tmpfs: RSS did not move). So the "1.3 MB per flow, 1.3 GB at the bound" I posted at `ba4dc1db` counted the guest, and the netstack's own share is the increment over the booted guest:

| Flood from one guest | Increment over booted guest | Per flow |
|---|---|---|
| 512 flows streaming `/dev/zero` at 256 KiB (round five, from its log) | about 133 MB | about 260 KB, one touched receive buffer |
| 512 flows streaming `/dev/zero` at 64 KiB | about 24 MB | about 48 KB |
| about 950 idle flows (live plus TIME-WAIT) at 64 KiB | about 64 MB | about 70 KB |

Zeroed buffers that a flow never writes into are not resident, so idle flows cost less than the 256 KiB the four buffers reserve. The 64 KiB decision still stands as the worst-case bound: a guest that fills every buffer at the 1024-flow bound pins about 256 MB instead of about 1 GB, which is what `docs/service.md` now says. DNS through the gateway kept answering before, during and after every flood.

**Not this PR's, found on the way.** A guest that opens 600 concurrent raw TCP flows to one destination has 256 of them established and 344 accepted-then-closed (the guest sees `CLOSE-WAIT`), and while they live and for about 30 s after they close, every new HTTPS request from that guest fails, to any destination, with `Recv failure: Connection reset by peer` or `Proxy CONNECT aborted`. The identical run on the gvproxy backend shows the identical picture (256 established, 344 `CLOSE-WAIT`, every destination unreachable during and 15 s after), so the limit is in the in-guest transparent proxy of lens-sandbox-core, not in the netstack. Filed as lens-sandbox-core issue lensapp/lens-sandbox-core#60.

From my side the opt-in release gate of NEW_NETWORK §8 is met at this head: smoke, throughput, memory bound, DNS, isolation probes, CI green, 2679 unit tests. Review requested.

chenhunghan:
Acceptance at `5850d13e` (round seven, DHCP through smoltcp's wire types) on this Mac, second service on its own home, signed release build.

**Smoke test**: every check passes: lease 192.168.127.2/24 from udhcpc against the library-emitted OFFER and ACK, resolver is the gateway, TLS egress to example.com over IPv4, witness control up, gateway :80 and `.254` and loopback refused with the host listener seeing no accept, chain names netstack.

**Tests**: `cargo test -p lns-service --lib` on macOS, 2683 passed, 0 failed, including the 16 dhcp tests (the sandbox could not build lns-service, so this is the first full run of the crate at this head outside CI).

Read the diff: one file, parsing and emission are `DhcpPacket`/`DhcpRepr`, the one-lease policy, the NAK for a foreign address, the broadcast rule and the echoed flag are unchanged; the raw-byte builder moved into the test module so malformed inputs can still be constructed. Two behaviours tightened, both covered: a message type option whose opcode does not match the packet opcode, and a non-Ethernet hardware type, are no longer read as requests.

Still requesting review from @jansav; the head moved from 916b7e8a to 5850d13e with this one refactor commit.

Issue #427, opened 2026-09-11
lns-service keeps console.log and two sockets open per finished run


chenhunghan:
Observed on macOS while accepting PR #426 and its gvproxy predecessor #424, so it is independent of the network backend.

After each finished `lns run --rm` the lns-service process keeps three descriptors open: the run's `runs/<id>/console.log` and two unix sockets. Thirty sequential runs took the process from 47 to 146 open files and its RSS from 82 MB to 301 MB (most of the memory in the first ten runs, so part of that is caches warming; the descriptors grow linearly).

Reproduce on a second service (`LNS_HOME`, `LNS_SOCKET_PATH`, `LNS_SERVICE_BIN`):

```
PID=$(lns service status | grep -o 'PID:\s*[0-9]*' | grep -o '[0-9]*')
lsof -p $PID | wc -l
for i in $(seq 1 10); do lns run --rm --yes -q -f lns.yaml -- true; done
lsof -p $PID | awk '$5=="REG"{print $9}' | grep -c console.log
```

Expected: descriptor count returns to its baseline after the runs finish. Suspects: `vm/diag_console` keeps the console.log handle after the VM exits; the two sockets look like the run's vsock and guest-transport ends.

PR #428, opened 2026-09-11
test: parity harness for the guest network backends, phase one (transfer integrity, lifecycle)

chenhunghan:
## Purpose

This is phase one of the verification programme for the in-process netstack
(issue #380, PR #426). It adds `tests/parity/`, a harness a person runs on a
real macOS host. The harness puts the same network cases to a lens-sandbox
guest on different network backends, and writes one machine-readable result per
run. Two results diff against each other, so a backend compares with another
backend — or with another build of itself.

Nothing here touches `lns-service`, `lns-cli` or the installer.

## The backend-as-configuration decision

A backend is **not a code path**. It is a name, an `lns` binary, an
`lns-service` binary and an environment:

```
$ cargo run -p net-parity -- run \
    --backend netstack --lns bin/lns --lns-service bin/lns-service \
    --env LNS_NETDEV=netstack --bind 192.168.1.50 --out netstack.json
```

The same shape repeats in a TOML file. The harness starts that service on a
private `LNS_HOME` and `LNS_SOCKET_PATH` with `LNS_HEADLESS=1` and the given
environment, runs every case through it, stops it, and checks that it is gone.
It never touches the service in your menu bar.

This is how `LNS_NETDEV=vmnet` and `LNS_NETDEV=netstack` are selected, and how
a PR #424 build with `LNS_GVPROXY_BIN` is driven — with no code in the harness
that knows what any of them is. The person supplies the binaries; the harness
records their SHA-256 and `lns --version` in the result. Nothing is downloaded.

Deliberate backend differences are named per backend in the TOML
(`expected_differences`). The diff prints them under their own heading and
stays green; anything else fails it.

## The cases

Images: `docker.io/curlimages/curl:8.11.1` where curl suffices,
`docker.io/library/alpine:3.20` for `dd`, `sha256sum` and busybox `nc`. Each
case verifies the commands it relies on inside the guest, and each is judged by
what the host fixture recorded — not by the guest's word.

| Case | Pass criteria |
|---|---|
| `lease-and-resolver` | `ip -4 addr show eth0` gives an address and prefix; `/etc/resolv.conf` names a nameserver. All three recorded. |
| `upload-100m` | The sink read 104857600 bytes and hashed them to the hash of 100 MiB of zeros. Throughput recorded. |
| `download-100m` | The guest's `sha256sum` equals the host's hash of 100 MiB of the pattern `byte i = (i * 7 + 3) mod 256`. |
| `bidirectional-100m` | Four counts (guest sent, fixture read, fixture sent, guest read) and both hashes agree. |
| `guest-half-close` | The fixture sees EOF after 10 MiB, replies 1 MiB, and the guest reads all of it. |
| `host-half-close` | The guest reads all 10 MiB sent before the fixture's half close; the fixture's hash of what the guest sent matches. |
| `reset-mid-transfer` | The fixture committed exactly 5 MiB and reset; the guest's command exits non-zero within 5 s. |
| `kill-mid-transfer` | After `lns kill`, the fixture's connection closes within 5 s, no byte arrives after it, and the service's descriptor count returns to its pre-case value within 10 s. |
| `udp-echo` | The 100-byte datagram comes back as 100 bytes and the empty one comes back empty. PR #426's second review found the empty one dropped: a fail here is a finding, not a harness bug. |
| `loopback-witness` | The witness on `127.0.0.1` recorded zero accepts for both probes (loopback and `<subnet>.254`). The TOML names each backend's expectation; the case records what happened either way. |
| `create-destroy-20` | Twenty run/kill/remove cycles leave the descriptor count within +3 of the baseline. The delta is always recorded; a larger one is `blocked-by-product` (issue #427), not a fail. |
| `service-stop-with-guests` | Both mid-transfer connections close within 10 s of `lns service stop` and the service PID is gone. |

While the cases run, the harness samples the service PID's RSS (`ps -o rss=`)
and open descriptor count (`lsof -p`) every 5 s into the result.

A case **skips with a reason** when a prerequisite is missing — a command the
image lacks, a busybox `nc` with neither `-N` nor `-q`, no guest tool that
sends a zero-length datagram. It never fails for the harness's own gaps.

`parity diff a.json b.json` prints the cases whose status or measured values
differ and exits 1 when an unexpected difference remains.

## What was unit-tested

96 unit tests, all in process, over loopback, no VM:

- **The fixtures** — the sink's counts and hash, the source's pattern, both
  directions at once, the half-close reply, the host half close, the reset with
  what it committed, the TCP echo, the UDP echo of an empty datagram, the
  witness's accept record, the report file the runner reads, and the refusal of
  a bind address no guest can reach.
- **The result schema** — round trip through the file the diff reads, refusal
  of another schema version, every status spelling, measures of each type, the
  skip and blocked reasons, the per-status counts.
- **The diff** — agreement, a changed status, a changed value, a measurement
  only one run took, a case only one run ran, an expected difference that keeps
  the diff green, and timings that are recorded but never diffed.
- **The harness plumbing** — the TOML grammar, `--env KEY=VALUE`, the case
  selection and its fixed order, the private service's own home, socket and
  environment, the service that reports no PID, the guest-command prelude, the
  `dd` byte count, throughput, the CLI grammar.

`cargo fmt`, `cargo clippy --all-targets -- -D warnings` and
`cargo clippy -- -D clippy::cognitive_complexity` are clean on the crate.
Dependencies are pinned exactly (`=1.0.102` and so on) and every one is already
in `Cargo.lock`, so the `audit` job covers them; there are no git dependencies.

## What needs a macOS host

**I could not boot a VM, so no case here has ever run against a guest.** This
sandbox has no macOS, no Vz and no KVM. The unit tests above cover the
fixtures, the schema, the diff and the plumbing; the cases themselves —
everything that talks to a real `lns` — are unrun. The first real run is the
master's, on a macOS host with two signed builds and a LAN IPv4 address.

CI compiles the crate and runs its unit tests with the rest of the workspace.
It never runs the harness. Adding the crate to the workspace `members` touches
the root `Cargo.toml`, so this PR's coverage job runs the full workspace gate;
the harness itself is outside `crates/`, so the per-file coverage gate does not
score it.

## Left out, by design

Policy cases, isolation under load and DNS fixture cases are later phases. The
`src/cases/` layout takes a new case as one function plus one registry line,
and the README names what those phases add.

## Fix round: the first real run (2026-09-11)

The master's run on a Mac found `lease-and-resolver` pass, `upload-100m` fail
after about nine minutes, and `download-100m` hang the same way. The cause was
this harness, not the backend. The generated definitions carried only an
`egress.http` catch-all. The sandbox puts its own nftables in front of every TCP
stream the guest opens and redirects it to the in-guest transparent proxy. The
proxy accepts the connect — which is why `nc -zv` reported the port open — and
then asks the policy about the stream. An `egress.http` rule does not decide a
raw stream, a destination no rule decides is asked about, and headless there is
nobody to ask. The stream was held and no byte reached the fixture.

### What changed

**1. Every generated definition decides the fixture destinations.**

```yaml
spec:
  image: docker.io/library/alpine:3.20
  egress:
    http:
      - match: "*"
        verdict: allow
    tcp:
      - match: "192.168.1.50:47200"
        verdict: allow
        description: parity host fixture
      - match: "192.168.1.50:47201"
        verdict: allow
        description: parity host fixture
      ...
```

A tcp `match` is a destination with a port (`docs/sandbox-spec.md` §4.2,
`crates/lns-policy/src/lib.rs`). An **IP literal with a port is accepted**, and
so is a CIDR with a port. There is **no port wildcard and no port range** — a
raw splice is granted unread, so "any port on this host" is not a grant LNS
writes. The harness therefore writes one entry per fixture port: the seven TCP
fixtures and the UDP echo port, taken from the ports the fixtures actually bound.
The witness is left out on purpose — it binds `127.0.0.1`, and
`loopback-witness` proves the guest cannot reach it. The http catch-all stays
for the curl cases.

**2. A preflight case, `fixture-reachable`, runs first.** The guest sends 16
bytes over raw TCP to the echo fixture and reads them back, with a 10 s
exchange. When it fails, every case that needs a raw stream is skipped with
`fixture unreachable from the guest: <the preflight error>` instead of stalling
one after another. A `--case` selection that needs a raw stream pulls the
preflight in whether you named it or not.

**3. Every case declares a budget.** The `lns run` command is killed at it, and
the case's guests are killed and removed before the next case starts. A case
past its budget is a fail: `budget exceeded after 121s, 0 bytes seen by the
fixture`.

| Case | Budget |
|---|---|
| The 100 MB transfers, the half closes, `reset-mid-transfer` | 120 s |
| `kill-mid-transfer`, `service-stop-with-guests` | 120 s — one cycle group plus one boot |
| `create-destroy-20` | 240 s — 60 s per group of five cycles |
| `fixture-reachable`, `lease-and-resolver`, `udp-echo`, `loopback-witness` | 120 s |

A budget covers the **whole** case, the guest boot included. Where this Mac is
slower than that, raise one without a code change:
`--budget download-100m=300`, or `budgets = { "download-100m" = 300 }` in the
config.

**4. The result is written after every case**, overwriting, not once at exit —
so a run that is stopped halfway still leaves a readable `result.json`. Each
case records what the fixtures saw while it ran — `fixture_seen_connections`,
`fixture_seen_bytes_in`, `fixture_seen_bytes_out` — whatever its verdict. A
budget failure with `fixture_seen_bytes_in: 0` says "nothing moved" from the
JSON alone.

**5. The docs say why.** The README prerequisites and the example TOML now name
the guest cage and the `egress.tcp` rule it needs.

### New tests

118 unit tests now (was 96). The new ones:

- the generated document's `egress.tcp` table, **parsed by the product's own
  parser** (`lns_policy::Egress`, the type `lns-artifact`'s `SandboxSpec` uses)
  and each rule put through `TcpEgressRule::validate()` — not compared as a
  string;
- the skip cascade: a failed preflight skips `upload-100m` with the preflight's
  own error and leaves `lease-and-resolver` and `loopback-witness` alone;
- the budget message, and that a case inside its budget or a skip past it keeps
  its own verdict;
- the command that is killed at its budget and still reports what it printed;
- the fixture destinations (the witness excluded) and the per-case activity
  counts;
- the `--budget NAME=SECONDS` grammar and its refusal of an unknown case.

`cargo fmt`, `cargo clippy --all-targets -- -D warnings` and `cargo audit` are
clean. The two new dev-dependencies — `lns-policy` (path) and `serde_yaml`,
both already in `Cargo.lock` — are test-only; `cargo audit` reports no
vulnerability and the same seven pre-existing warnings.

### How to re-run

```
$ cargo run -p net-parity -- run \
    --config parity.toml --backend netstack --out netstack.json
```

Nothing else changed in the invocation. Watch for this first:

```
[case]  fixture-reachable (budget 120s)
[pass]  fixture-reachable
```

A `fail` there means the raw stream still does not carry, and the run is over in
about two minutes instead of nine per case — `netstack.json` names the
destination and what the guest read. To put one case on its own,
`--case upload-100m` runs the preflight with it. If a boot on this Mac is slower
than a budget allows, raise that budget rather than reading it as a backend
fault.

### Deviations

- **Two commits, not three.** The egress table and the preflight are interlocked
  with the budgets: the preflight only fails fast because the guest command is
  killed at its budget, and the registry that carries the budgets is the registry
  that carries the preflight. Splitting them would have landed a commit that does
  not compile. The docs are their own commit.
- **The preflight's 10 s is the exchange, not the case.** The case has a 120 s
  budget, because it also boots a guest. The 10 s is the `nc` timeout the guest
  spends on the stream itself.
- **Still unrun against a guest.** This sandbox has no macOS, no Vz and no KVM.
  The first real run of the fix is the master's.

## Feature round: fixtures on another machine (2026-09-11)

The second real run showed the harness behaves and the **host** does not. On
that Mac, TCP to the host's own LAN address (192.168.1.49) or to the bridge
address (192.168.64.1) completes its handshake and is then black-holed: the
handshake succeeds and the accept never fires, for a plain shell and for a guest
alike, while the same listener on 127.0.0.1 accepts every time. Loopback is
refused by the netstack by design. So no fixture on that host is reachable from
a guest, and the fixtures have to live on another machine.

This round adds that mode. The in-process mode is unchanged.

### How to run the two halves

**On the fixture machine** (macOS or Linux on the same LAN), from this same
revision:

```
$ cargo build -p net-parity --release
$ ./target/release/parity fixtures --bind 192.168.1.77 --base-port 47200 --serve
bidirectional   192.168.1.77:47202
echo            192.168.1.77:47206
...
report  http://192.168.1.77:47220/report
reset   http://192.168.1.77:47220/reset
health  http://192.168.1.77:47220/health
a runner on another machine reads these with: parity run --fixtures-at 192.168.1.77:47200
writing fixtures.json every second; stop with ctrl-c
```

**On the machine with the guests**, name that address instead of `--bind`:

```
$ cargo run -p net-parity -- run \
    --config parity.toml --backend netstack \
    --fixtures-at 192.168.1.77:47200 --out netstack.json
[service]  pid 54120 home /tmp/parity-54098/home socket /tmp/parity-54098/service.sock
[case]  fixture-reachable (budget 120s)
[pass]  fixture-reachable
...
```

`--bind` is not needed then. The address decides which fixtures a run meets.

### What changed

**1. The fixtures serve their report.** `parity fixtures --bind IP --base-port N
--serve` runs the fixtures as before and also serves them over HTTP on
`N + 20`, bound to the same IP:

| Route | What it does |
|---|---|
| `GET /health` | The harness version and every fixture this process bound. |
| `GET /report` | The current fixtures JSON — the same document `fixtures.json` holds. |
| `POST /reset` | Clears the counters. |

The handler is hand-written over tokio: it reads the request line and answers
one of the three routes, 405 for another method on one of them, and 404 for
anything else. No web framework. tokio was already in `Cargo.lock`; the only
manifest change is the feature set this crate asks for (`rt`, `net`, `io-util`,
`time`). `--serve` asks the server for its own health before it prints the URLs.

**2. The runner drives the guest against those addresses.** It asks `/health`
first and refuses to start when nothing answers, when the answer is not this
harness's report server, when a fixture is missing, or when the version differs
— two halves of different builds measure different things. It then calls
`/reset` before **every** case and reads `/report` while the case runs, in place
of the in-process fixture handles. The generated `egress.tcp` rules and the
`fixture-reachable` preflight use the remote address.

**3. The loopback witness stays local.** It has to bind the service host's own
127.0.0.1 for `loopback-witness` to prove anything, so the runner binds it here
and folds its accepts into the report it reads. It is never a destination the
generated definition grants.

**4. The result says which fixtures it met**, so two results are only diffed
when they are comparable:

```json
"fixtures": { "host": "192.168.1.77:47220", "mode": "remote", "version": "0.25.0" }
```

`schema_version` is therefore **2**. `diff` compares two runs only when host,
mode and version all agree; when they do not it says so and exits 1:

```
$ cargo run -p net-parity -- diff netstack.json vmnet.json
netstack (lns 0.25.0) vs vmnet (lns 0.25.0)
these runs are not comparable: netstack used in-process fixtures on 192.168.1.49 (harness 0.25.0), vmnet used remote fixtures on 192.168.1.77:47220 (harness 0.25.0)
put both backends to the same fixtures before you read the differences below
no unexpected difference
```

**5. The README gets a "Fixtures on another machine" section** — how to build
and run both halves, the three routes, and a paragraph naming the host black
hole and its symptom, so a reader recognises it: the handshake succeeds and the
accept never fires. It also says how to tell it apart from a policy fault — a
`fixture-reachable` that fails against fixtures on a second machine is a real
backend finding.

### New tests

143 unit tests now (was 118), all in process over loopback. The new ones:

- the report server's three routes: health names the version and the fixtures it
  bound, report carries what a fixture read, reset clears the counters between
  two cases, an unknown route is 404 and a wrong method 405;
- the runner reading a remote report and resetting between cases, against a
  fixture process served in the same test;
- a version mismatch refused, a server that answers something else refused, a
  fixture set that is missing a fixture refused, and a reset that never answered
  reported rather than read as quiet;
- the local witness folded into the remote report, and left out of the
  destinations the definition grants;
- the result's `fixtures.host`, `fixtures.mode` and version through the file the
  diff reads;
- `diff` on mismatched fixture modes, on the same remote fixtures, and on
  another harness version;
- the `--fixtures-at` and `--serve` grammar, and the run that names neither
  `--bind` nor `--fixtures-at`.

`cargo fmt`, `cargo clippy --all-targets -- -D warnings`,
`cargo clippy -- -D clippy::cognitive_complexity` and `cargo audit` are clean.
`cargo audit` reports no vulnerability and the same seven pre-existing warnings.
Each of the three commits builds, tests and lints on its own.

### Deviations

- **`schema_version` went from 1 to 2 rather than gaining an optional field.**
  A result with no `fixtures` record cannot be judged comparable, so an older
  file is refused rather than diffed under a guess. No such file is kept
  anywhere.
- **The remote report is fetched per read, with no cache.** A poll costs one
  short HTTP request on the LAN. A fetch that fails keeps the last report, and
  the next `/reset` — one per case — names the failure and fails that case.
- **Still unrun against a guest.** This sandbox has no macOS, no Vz and no KVM,
  and the remote half needs two machines. The first real run of this round is
  the master's.

Refs #380


chenhunghan:
First real run, on this Mac, netstack backend at PR #426 head 70f7dad3, harness at a9376f8: `lease-and-resolver` passed, `upload-100m` failed after about nine minutes, `download-100m` hung the same way and I stopped the run.

Cause, confirmed from inside the stalled guest with `lns sandbox exec`: `nc -zv 192.168.1.49 47201` reports the port open, but `echo hi | nc 192.168.1.49 47201` never moves a byte. The connect is accepted by the guest's own transparent proxy, which the sandbox's nftables put in front of every TCP stream, and the generated definitions carry only an `egress.http` catch-all. With no `egress.tcp` entry deciding that destination the stream is held, and nothing reaches the fixture on the host. So this is a harness defect, not a backend one: the definitions need a `tcp` table for the fixture destinations, a preflight that proves one raw stream round-trips before the transfers, per-case budgets so a stall fails in seconds, and the result written after every case. A fix round is dispatched.

chenhunghan:
Second real run at 97b3694, netstack backend: the fix behaves. `fixture-reachable` failed in 21 s with the fixture counts recorded, the seven transfer cases skipped with its reason, `lease-and-resolver` and `loopback-witness` passed, and result.json was on disk after every case. The vmnet backend fails `fixture-reachable` and `udp-echo` the same way.

The remaining failure is the host, not the harness or a backend. On this Mac any TCP connection to the host's own LAN address (192.168.1.49) or to the bridge address (192.168.64.1) completes its handshake and is then black-holed: `nc` from a plain shell reports success and a Python listener bound to that address never gets the accept, three times out of three, while the same listener on 127.0.0.1 accepts every time. The host runs the SentinelOne network-monitoring and Cato Client network extensions; this is the TCP twin of the content-filter drop behind #380. Nothing the guest or the netstack does can reach a fixture on this host, and loopback is refused by design.

So the harness needs a mode the brief did not ask for: fixtures on another machine. `parity fixtures --bind` already runs them standalone and writes a report, but `run` only knows in-process fixtures. Needed: `run --fixtures-at HOST:BASE_PORT --fixtures-report PATH` (or an equivalent: the runner drives the guest and reads the remote report file, copied or served by the fixture process over the same TCP echo port) so a second Mac or a Linux box on the LAN hosts them. Until then the runnable cases here are the four that need no fixture.

chenhunghan:
One more for the next round, found while running the fixture-free cases on all three backends (netstack and vmnet pass `lease-and-resolver` and `loopback-witness`): the gvproxy backend (a PR #424 build with `LNS_GVPROXY_BIN`) fails to boot under the harness because the private home lives under the macOS temp directory and gvproxy's `net.sock` path comes out at 118 bytes, over the 103-byte unix socket limit. The harness needs a `--home-root DIR` (default the temp directory) so a backend that puts sockets in the run directory can be given a short one, such as `/tmp/p`. Record the chosen root in result.json.

chenhunghan:
gvproxy baseline, resolved and run. With `TMPDIR=/tmp/p` the private home is short enough for gvproxy's run-directory socket, so no `--home-root` is needed; a line in the README's prerequisites is enough ("a backend that puts sockets in the run directory needs a short `TMPDIR`"). Results of the two fixture-free cases on all three backends, harness at 15ad60b:

| Case | netstack (PR #426 db8bebc) | vmnet (same build) | gvproxy (PR #424 build, gvproxy 0.8.9) |
|---|---|---|---|
| lease-and-resolver | pass | pass | pass |
| loopback-witness | pass, zero accepts | pass, zero accepts | **fail: the witness on 127.0.0.1 accepted 1 connection from the guest** |

That failure is the harness doing its job: it reproduces the review finding that sent PR #424 back, the guest reaching the host's loopback through gvproxy's `.254` NAT. For the gvproxy backend this is a known, deliberate difference and belongs in the example TOML's `expected_differences` for that backend, with a comment saying what it is, so the baseline diff stays green while the record shows it.

